PR #6, opened 2017-04-18
ENH: add logic for binary ufuncs

jakevdp:
This adds the ``elemwise_binary`` method that implements an efficient version of the logic needed to perform binary ufuncs where both operators are ``COO`` matrices (as long as ``func(0, 0)=0``).

Currently I only enabled ``__mul__`` and ``__rmul__``, but it should be straightforward to add more.

First commits use a KD Tree approach; the final commit changes this to a binary search within the lex-sorted coordinate array, which should be much faster.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -403,12 +403,6 @@ class COO(object):
         unmatched_other = np.ones(len(other.data), dtype=bool)
         unmatched_other[matched_other] = False
 
-        print(self.coords.shape, self.data.shape)
-        print(matched_self, unmatched_self)
-        print('----------')
-        print(other.coords.shape, other.data.shape)
-        print(matched_other, unmatched_other)
-
         coords = np.hstack([self.coords[:, matched_self],
                             self.coords[:, unmatched_self],
                             other.coords[:, unmatched_other]])
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -4,6 +4,7 @@ import operator
 
 import numpy as np
 import scipy.sparse
+import scipy.spatial
 
 
 class COO(object):
@@ -357,7 +358,10 @@ class COO(object):
         return self + (-other)
 
     def __mul__(self, other):
-        return self.elemwise(operator.mul, other)
+        if isinstance(other, COO):
+            return self.elemwise_binary(operator.mul, other)
+        else:
+            return self.elemwise(operator.mul, other)
 
     __rmul__ = __mul__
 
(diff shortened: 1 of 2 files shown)


jakevdp:
It would probably also be more efficient to do this via a lexsort rather than a kdtree... but it would require a version of ``searchsorted`` that handles lexical sorting...

jakevdp:
OK, changed to lex sorting based on record arrays. That should be faster than a KD Tree approach in general.

jakevdp:
It would be worth benchmarking to see whether ``__add__`` should be implemented similarly, or implemented via ``sum_duplicates()`` as it is currently.

sparse/core.py
```diff
@@ -379,6 +382,52 @@ def elemwise(self, func, *args, **kwargs):
...
+        assert isinstance(other, COO)
+        if kwargs.pop('check', True) and func(0, 0, *args, **kwargs) != 0:
+            raise ValueError("Performing this operation would produce "
+                    "a dense result: %s" % str(func))
+        if self.shape != other.shape:
+            raise ValueError("Broadcasting is not supported")
+        self.sum_duplicates()  # TODO: document side-effect or make copy
+        other.sum_duplicates()  # TODO: document side-effect or make copy
+
+        # Sort self.coords in lexographical order using record arrays
```
mrocklin: Does `sum_duplicates` do this already?  If not then should it be made to?  The `sum_duplicates` method was changed a short while before you submitted this PR and now calls np.lexsort on coords directly.  It doesn't call it on a recarray though.  Honestly I didn't seriously look into how lexsort operates so am not very informed here.

sparse/tests/test_core.py
```diff
@@ -153,6 +154,17 @@ def test_elemwise(func):
     assert_eq(func(x), func(s))
 
 
+@pytest.mark.parametrize('func', [operator.mul])
```
mrocklin: Is there a reason not to add `operator.add` to this list?

sparse/tests/test_core.py
```diff
@@ -153,6 +154,17 @@ def test_elemwise(func):
...
+def test_elemwise_binary(func):
+    x = random_x((2, 3, 4))
+    y = random_x((2, 3, 4))
```
mrocklin: We might want to try a few shapes, in particular we should be sure to include a single-dimensional array.

sparse/tests/test_core.py
mrocklin: Ah, I see, the current solution just appends and waits for future operations for force sum_duplicates.

sparse/core.py
jakevdp: I did check this – ``lexsort`` and ``recarray`` sorting aren't identical: they use opposite ordering of the dimensions. I figured we could get it working, then optimize later.

jakevdp: I think using ``lexsort`` makes sense in ``sum_duplicates`` because it doesn't require copying of memory. The reason I had to use the record array here is to get the ``searchsorted`` functionality on lexically sorted arrays, but it does involve making a copy of all the coordinate data.

mrocklin: Agreed on "optimize later".  FWIW I haven't yet been avoiding memory copies.  Initial experiments  hasn't shown this to be an issue yet.

PR #8, opened 2017-04-25
Support multiplication with empty arrays

mrocklin:
@jakevdp you might be more familiar with this code.  Do you see a cleaner way to support this?  

Also, thoughts on removing non-zeros eagerly?  I can see going either way on this.


mrocklin:
Going to go ahead and merge this for now in favor of moving quickly.  We can revisit in the future.  Retroactive feedback still appreciated.

Issue #9, opened 2017-04-26
Fast conversion to scipy.sparse.csr/csc_matrix


mrocklin:
In computations involving tensordot we can be bound by routines to convert from sparse.COO to `scipy.sparse.coo_matrix` (this should be free) and to convert from `scipy.sparse.coo_matrix` to `scipy.sparse.csc_matrix`.  I do not believe that these are currently running at optimal speed.

I'm currently getting around some of this by caching the csr matrix on the sparse.COO matrix, but this is a bit kludgy and doesn't help if we're doing reshapings, transposes, etc..

Instead, given the importance of this operation, it might be worth defining a single step computation from a 2D sparse.COO matrix without duplicates to scipy.sparse.csr/csc matrices.

jakevdp:
Hmm... what that effectively involves is an efficient sorted group-by in the row numbers. Are you certain that scipy is suboptimal, or is that just inherently a slow operation?

jakevdp:
Actually, I take that back. Assuming no duplicates, you could do it as a O[N log N] lexsort on the column numbers, followed by an O[N] pass to construct the ``intptr`` array. It would be like 4-5 lines of numpy :smile: Would be worth benchmarking.

mrocklin:
scipy.sparse also spends about half of its time checking consistency.  If we're able to guarantee things externally it would be nice to skip this.

mrocklin:
What if we are sorted already?  In my current application (which I think is representative of a few applications), I want to perform tocsr-then-matmul computations repeatedly on the same data throughout the computation.  The data has already been normalized with `sum_duplicates` and so may already be in some predictable order.  When I was thinking about this briefly before I was curious if it was doable in two linear passes over the data.

jakevdp:
Here's some code for fast conversion to csr; csc would be similar:

```python
def to_csr(self):
    assert self.ndim == 2
    self.sum_duplicates()
    i = np.lexsort(self.coords[::-1])
    row, col = self.coords[:, i]
    data = self.data[i]
    indptr = np.zeros(self.shape[0] + 1, dtype=row.dtype)
    np.cumsum(np.bincount(row, minlength=self.shape[0]), out=indptr[1:])
    
    return csr_matrix((data, col, indptr), shape=self.shape)
```

It basically takes three passes through the data: summing the duplicates, sorting the indices, and cumulating the counts within each row. If you know that any of those conditions is already met, you could skip that step and speed things up.

jakevdp:
[...]

PR #11, opened 2017-05-01
Optimize CSR and sorted conversions

mrocklin:
This started by adding the optimzied version of tocsr from #9 (cc @jakevdp) 

Then, to avoid calling sorting routines excessively we started tracking `sorted=` metadata.

Then, to accelerate sorting and to consolidate lexsort and reshape code we based all sorting code onto a single function called `linear_loc`, which provides the linear location of any index in a C-ordered array.  This replaces `np.lexsort` with `np.argsort` and performs a quick `issorted` check beforehand.  This can dramatically speed up some workflows, but does limit the size of our arrays to `2**64` elements (zero or nonzero) so for example this library can no longer represent an array of shape `(1e6, 1e6, 1e6, 1e6)`.  This was already the case in reshape.  I'm personally ok with this for the time being.  I think that we can probably remove this limitation in the future (with a cost) if necessary.

There is still work to be done to optimize matvecs and matmuls, but this is a decent start.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -154,6 +154,7 @@ class COO(object):
         assert not self.shape or len(data) == self.coords.shape[1]
         self.has_duplicates = has_duplicates
         self.sorted = sorted
+        self._cache = defaultdict(lambda: deque(maxlen=3))
 
     @classmethod
     def from_numpy(cls, x):
@@ -209,6 +210,9 @@ class COO(object):
             index = (index,)
         index = tuple(ind + self.shape[i] if isinstance(ind, numbers.Integral) and ind < 0 else ind
                       for i, ind in enumerate(index))
+        if (all(ind == slice(None) or ind == slice(0, d)
+                for ind, d in zip(index, self.shape))):
+            return self
         mask = np.ones(self.nnz, dtype=bool)
         for i, ind in enumerate([i for i in index if i is not None]):
             if ind == slice(None, None):
@@ -325,19 +329,15 @@ class COO(object):
         if axes == tuple(range(self.ndim)):
             return self
 
+        for ax, value in self._cache['transpose']:
+            if ax == axes:
+                return value
+
         shape = tuple(self.shape[ax] for ax in axes)
         result = COO(self.coords[axes, :], self.data, shape,
                      has_duplicates=self.has_duplicates)
 
-        if axes == (1, 0):
-            try:
-                result._csc = self._csr.T
-            except AttributeError:
-                pass
-            try:
-                result._csr = self._csc.T
-            except AttributeError:
-                pass
+        self._cache['transpose'].append((axes, result))
         return result
 
     @property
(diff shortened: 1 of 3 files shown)


mrocklin:
I have extended this to cache the last few results of reshape and transpose.  This makes things run quite fast for me in practice.  Copying commit message here for convenience:

This stores the last few objects returned from reshape and transpose
calls.  This allows efficiencies from in-place operations like
`sum_duplicates` and `sort_indices` to persist in interative workflows.

Modern NumPy programmers are accustomed to operations like .transpose()
being cheap and aren't accustomed to having to pay sorting costs after
many computations.  These assumptions are no longer true in sparse by
default.  However, by caching recent transpose and reshape objects we
can reuse their inplace modifications.  This greatly accelerates common
machine learning workloads.

mrocklin:
Merging shortly if no objections

jakevdp:
Looks good. A couple things:

- the CSC convrtsiom could be made more efficient by sorting the indices as CSC directly, rather than doing CSR first

- for the cacheing, does this mean that sparse arrays are effectively read-only? If you modify them in place, cached results will be incorrect yes?

- in that subject, we should be careful about CSC/CSR conversions that might use views of data matrices, as you light inadvertently change the sparse matrix when changing the CSR generated from it.

mrocklin:
All good points.

1.  I'm tempted to leave csc as is until I (or someone else) actively needs a fast csc.  I do expect this to happen given my current application set, but it hasn't happened yet.
2.  For caching this gets tricky.  Currently I think that sparse arrays are practically read-only.  We haven't added any top-level in-place modification API.  However it is reasonable to expect that this will be added in the future.  We can easily invalidate whenever the parent is changed, but it would require more bookkeeping to invalidate whenever a child is changed.
3.  Regarding views how does numpy handle this?  If I change a view in NumPy then I expect the underlying array to change as well.  I guess that the situation here is more complex because it's not immediately clear to the user what is and is not a view.

For the caching issue I'm tempted to implement the book-keeping necessary for invalidation.

jakevdp:
Makes sense.

Regarding the book-keeping for cacheing: we could detect changes in the arrays by making them properties, but that still wouldn't protect against users modifying the contents of these arrays either directly, or via views in derived CSR matrices (for example). One way we might address that is by setting the ``writeable`` flag of the arrays to False.

mrocklin:
Alternatively we could have the user explicitly opt-in to caching.  This may be simpler all around.

jakevdp:
> Alternatively we could have the user explicitly opt-in to caching. This may be simpler all around.

+1, good solution for right now

mrocklin:
Planning to merge this soon if there are no further comments

jakevdp:
Looks good – one thing came to mind though: I wonder if it would be possible to use ``functools.lru_cache`` instead of maintaining the cache manually?

mrocklin:
I would love to use a standard LRU data structure.  However the `functools.lru_cache` solution is strictly a decorator, it doesn't let you easily inspect the contents, which I've found to be inconvenient in the past.  I'm inclined to use the simple approach with a deque here until it proves cumbersome.  If that fails then I have my own lru data structures that I can bring in if necessary.

jakevdp:
Makes sense!

PR #12, opened 2017-05-01
Fix windows errors around int/long

mrocklin:


mrocklin:
cc @jcrist these *may* have had an effect.  Not sure.

PR #14, opened 2017-05-02
Add @ operator

DavidMertz:


sparse/tests/test_core.py
```diff
@@ -136,6 +137,10 @@ def test_dot():
     assert_eq(a.dot(b), sa.dot(sb))
     assert_eq(np.dot(a, b), sparse.dot(sa, sb))
 
+    if sys.version_info >= (3, 5):
+        assert_eq(eval("a @ b"), eval("sa @ sb"))
+        assert_eq(eval("sa @ sb"), sparse.dot(sa, sb))
```
mrocklin: Cool.  If you wanted to test `__rmatmul__` then you might also add something like the following line:

    assert_eq(eval("a @ sb"), sparse.dot(a, sb))

DavidMertz: Done!

DavidMertz: Actually... no.  Your suggestion will simply exercise the `__matmul__` method of NumPy arrays, not the `__rmatmul__` method of sparse.  Let me fix that and push the change.

sparse/core.py
```diff
@@ -328,6 +328,11 @@ def T(self):
     def dot(self, other):
         return dot(self, other)
 
+    __matmul__ = dot
+
+    def __rmatmul__(self, other):
+        return dot(np.array(other), self)
```
mrocklin: Hrm, this coercion is a bit concerning.  What if, for example, `other` is a scipy.sparse matrix?  Coercing it to a numpy array may be unwelcome.

DavidMertz:
I did think of that. We need something with an `.ndim` attribute here. Would a COO sparse array itself be a better coersion? Or `scipy.sparse`?

DavidMertz:
My thought was that if it's something without it's own `__matmul__` method then it's probably something dense like a Python list(of lists). If it's scipy.sparse, or xarray, or similar, then it probably won't need the fallback to right hand side.

DavidMertz:
My thought was that if it's something without it's own `__matmul__` method then it's probably something dense like a Python list(of lists). If it's scipy.sparse, or xarray, or similar, then it probably won't need the fallback to right hand side.

jakevdp:
I think that ``ndarray.__matmul__(sparse.COO)`` should return a TypeError, at which point Python will try ``sparse.COO.__rmatmul__(ndarray)``, so you don't have to worry about any sort of coercion.

mrocklin:
Unless it doesn't recognize `sparse` arrays, in which case we should probably expect it to not understand what to do.  I think that there are two topics here:

1.  Logic like "how to handle other types" should be in the proper functions like tensordot, not in the API methods like `__rmatmul__`
2.  We should avoid coercing things to numpy arrays if we don't know what's going on.  Given that we are playing in sparse-land, this is dangerous

DavidMertz:
@jakevdp. Yes, you are right that `ndarray.__matmul__(sparse.COO)` raises. It's a `NotImplemented` rather than `TypeError`, but either way it doesn't do the operation).  So I added back in @mrocklin's proposed check.

However, I'm concerned about operations that might involve types that don't have an `.ndim` attribute.  NumPy handles them.  So does `scipy.sparse` (with a caveat that confuses me).  E.g.:

```python
>>> import numpy as np
>>> from sparse import COO
>>> import scipy.sparse
>>>  a = np.array([5,6,7,8,9])
>>>  sa = COO(a)
>>>  l = [1, 2, 3, 4, 5]
>>> spa = scipy.sparse.coo.coo_matrix([5,6,7,8,9])
>>> a @ l, l @ a
>>> (115, 115)
>>> spa @ l, l @ spa.T  # <-- why the transpose needed in this case?!
(array([115], dtype=int64), array([115], dtype=int64))
```

So far so good (except the difference in the transpose that confuses me).  What should *sparse* do in the equivalent operations? Right now, on my fork:

```python
>>> l @ sa
array(115, dtype=int64)
>>> sa @ l
[...]
AttributeError: 'list' object has no attribute 'ndim'
```
I think I want the result from `__rmatmul__` for `__matmul__` also.  It would be easy enough to do... if we could figure out how best to coerce the thing lacking an `.ndim` attribute.

jakevdp:
> Yes, you are right that ndarray.__matmul__(sparse.COO) raises. It's a NotImplemented rather than TypeError, but either way it doesn't do the operation. So I added back in @mrocklin's proposed check.

Exactly... so unless I'm mistaken, ``x @ M`` for an array ``x`` and COO matrix ``M`` will recognize this error, and then automatically call ``M.__matmul__(x)``. That is, you do not need to worry about this kind of type coercion here... it will be taken care of for you.

If you want ``dot`` to handle lists correctly, that should be fixed in ``dot``/``tensordot``, not in ``__matmul__``. It's best to think of ``__matmul__`` as just a thin wrapper of ``dot`` that should do as little as possible to the inputs (similar to the scipy.sparse code I linked to above).

sparse/tests/test_core.py
jakevdp: I don't think this should be merged unless ``x @ M`` for an array ``x`` and a matrix ``M`` works correctly. There's no reason why it shouldn't.

jakevdp:
Forgot one detail: to make matmul behave correctly with numpy arrays, we need to define a [__numpy_ufunc__](https://docs.scipy.org/doc/numpy-dev/reference/arrays.classes.html#numpy.class.__numpy_ufunc__) method.

DavidMertz:
I seem to have the same problem in http://stackoverflow.com/questions/41494823/use-of-numpy-ufunc.  The NumPy machinery seems simply to ignore `__numpy_ufunc__`. Do you have even a toy test that actually enters that magic method?

I've added this test, which passes:

```python
# Test that SOO's and np.array's combine correctly
assert_eq(eval("a @ sb"), eval("sa @ b"))
```

jakevdp:
Yeah, that's strange. I tried some things, and found that numpy never seems to actually use ``__numpy_ufunc__``, but if the method exists, then the ``rmatmul`` works correctly.

Again, I think we need to remove all array casting from ``__mul__``, ``__matmul__``, and ``COO.dot``, and basically make them straightforward references to the ``dot()`` function. It would be quite confusing if the various ways of calling ``dot`` all behaved slightly differently.

DavidMertz:
I've refactored to add more tests and remove all logic from `__matmul__` and `__rmatmul__`.  I've made the casting logic default-off (and tests for this expected failure of `sa @ lb`).  It will only do coercion that if explicitly asked for with `sparse.dot(..., make_array=True)` or by creating a sparse matrix with `COO(..., toarray_other=True)`.

I believe that it's fairly reasonable to believe that anything lacking `.ndim` is some sort of dense collection such as a list-of-lists.  But in the current revision, the code will not make this assumption unless specifically requested to do so.

I *do not* need any `__numpy_ufunc__` method to get the call to `__rmatmul__` as far as I can tell (I *think* the tests cover this).

mrocklin:
Personally I think that tracking `to_array_other=` and `make_array=` keywords is not necessarily something that we want to commit to handling.  Or, if we intend to handle these then we should probably check in with the numpy/scipy projects and see if they want to support these first.  I suspect that this topic is more complex than it appears at first.

Instead, I recommend tabling the problem of "how to handle calling `.dot` on nested lists" for another day and focus just on adding the `__matmul__` and `__rmatmul__` operators for this PR.

jakevdp:
Looks better.

I still maintain that this PR is not complete until ``sa @ lb`` works correctly. It works for ``scipy.sparse``; there's no reason it shouldn't work here. For reasons I don't fully understand (because the code within the method never seems to be executed), defining ``__numpy_ufunc__`` seems to be the key to getting this to work. See [scipy.sparse.base](https://github.com/scipy/scipy/blob/master/scipy/sparse/base.py) for an example.

I also agree with @mrocklin that the new keywords are not an ideal solution. I would be fine with ``dot`` (or better ``tensordot``) automatically casting list inputs to arrays, but it will have to be handled a bit more carefully. For example ``scipy.sparse`` does it [this way](https://github.com/scipy/scipy/blob/master/scipy/sparse/base.py#L405-L429).

I think it's fine to ignore that kind of casting logic for this PR if you don't want to tackle it, and for the time being just assume, like ``dot`` and ``tensordot`` currently do, that all inputs are arrays or matrices. We can correctly handle list and scalar inputs at a later date.

jakevdp:
Here's what I mean about ``__numpy_ufunc__``:

behavior without it:

```python
class Sparse(object):
    def __matmul__(self, other):
        return print('matmul', type(self), type(other))
    
    def __rmatmul__(self, other):
        return print('rmatmul', type(self), type(other))
    
M = Sparse()
x = np.random.rand(10)

M @ x
x @ M
```
```
matmul <class '__main__.Sparse'> <class 'numpy.ndarray'>
TypeError                                 Traceback (most recent call last)
<ipython-input-7-de5807ef8d32> in <module>()
     10 
     11 M @ x
---> 12 x @ M

TypeError: Object arrays are not currently supported
```

Behavior with it:
```python
class Sparse(object):
    def __matmul__(self, other):
        return print('matmul', type(self), type(other))
    
    def __rmatmul__(self, other):
        return print('rmatmul', type(self), type(other))
    
    def __numpy_ufunc__(self, *args, **kwargs):
        return NotImplemented
```
```
matmul <class '__main__.Sparse'> <class 'numpy.ndarray'>
rmatmul <class '__main__.Sparse'> <class 'numpy.ndarray'>
```

DavidMertz:
@jakevdp  wrote:
> still would maintain that this PR is not complete until sa @ lb works correctly. It works for scipy.sparse; there's no reason it shouldn't work here. 

This has been added as a test in the current revision, and it succeeds. For *some reason* I don't need `__numpy_ufunc__` like Jake's toy example does (nor can I get one to be executed if I add it; but in particular, it's presence/absence does not affect the SOO code the  way it does the toy one... which is very strange).

The test in the code is:

```python
assert_eq(eval("a @ lb"), eval("sa @ lb"))
```

In the shell (on my system with current code):

```python
In [1]: import numpy as np
In [2]: from sparse import COO
In [3]: l = [1, 2, 3, 4, 5]
In [4]: a = np.array(l)
In [5]: sa = COO(a, toarray_other=True)
In [6]: sa @ l
Out[6]: array(55, dtype=int64)
```

**However**, this depends on the setting/mode:

```python
In [8]: sa._toarray_other = False
In [9]: sa @ l
[...]
AttributeError: 'list' object has no attribute 'ndim'
```

I'm really **not** in love with the attribute, argument magic for enabling/disabling this "coerce list to NumPy" behavior.  But having the *never support lists* behavior violates an expectation of NumPy and scipy.sparse users.  Having the *always coerce* behavior has the potential to encounter something that is actually sparse (although lacking an `.ndim` attribute, which would be strange), and doing far too much work.  Allowing a switch felt like the easiest way of addressing this.

I considered trying to implement behavior for "convert lists with no more than N elements" (with N configurable in a way similar to what I actually did with the boolean switch).  But `len(x)` won't give me the right answer by itself.  I *could* write a small utility function to look at len(), then recursively at `x[0]`, `x[0][0]`, etc. as far as needed, and multiply for the total size.  Then potentially we could use `._toarray_smaller_than_N` as the magic switch (or some spelling, which really comes down to some keyword argument in `sparse.dot()` itself).

DavidMertz:
@jakevdp wrote:
> I would be fine with dot (or better tensordot) automatically casting list inputs to arrays, but it will have to be handled a bit more carefully. For example scipy.sparse does it this way.

What scipy.sparse does is almost exactly what I do.  I guess they use `np.asanyarray` rather than simply `np.array`.  But this only matters for things that are already arrays (and have `.ndim`), no? The real difference is just that scipy.sparse does the coercion *always*, where I but in a switch not to do it if not requested.

jakevdp:
I think scipy sparse is a bit more comprehensive than what you do: it first checks for known types (array, sparse matrix, scalar) and if all those fail, it tries ``asarray`` and then double-checks the shape and dtype to make sure it's not just creating a zero-dimensional object array. Those pieces are pretty important to avoid strange edge cases, I think.

jakevdp:
Sorry, I think I misspoke earlier: I think this PR should avoid the question of up-casting lists, but that the test you currently have marked as an expected failure should pass. I believe you can do that without adding the extra keywords, which in my opinion **should not** be added to the library.

mrocklin:
> it first checks for known types (array, sparse matrix, scalar) and if all those fail, it tries asarray 

You know, given the presence of libraries like `sparse` this may also be undesired.  Scipy.sparse may be incorrectly assuming that it knows the entire world of arrays out there.  I don't think that we should make this assumption.  I would be in favor of adding an explicit type check for lists in tensordot.  I would be somewhat against calling `asarray` on an input of unknown type.

DavidMertz:
If we do it based on type checking and eschew the keywords, we will at
least need to allow lists and tuples (and all nestings of one inside the
other). This is what np.array is happy to handle.

Is that agreeable to you also, Jake? I.e. automatically coerce
lists/tuples, regardless of size, but fail on anything else lacking an
.ndim.

Fwiw, I believe my check for .ndim is more accurate for checking "something
duck typed the right way" than what scipy does.

On May 3, 2017 9:29 AM, "Matthew Rocklin" <notifications@github.com> wrote:

it first checks for known types (array, sparse matrix, scalar) and if all
those fail, it tries asarray

You know, given the presence of libraries like sparse this may also be
undesired. Scipy.sparse may be incorrectly assuming that it knows the
entire world of arrays out there. I don't think that we should make this
assumption. I would be in favor of adding an explicit type check for lists
in tensordot. I would be somewhat against calling asarray on an input of
unknown type.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mrocklin/sparse/pull/14#issuecomment-298963711>, or mute
the thread

jakevdp:
What I'm saying is we should punt on the question of lists and tuples and focus on the meat of this PR: getting the matmul operator working for the inputs that dot currently supports.

jakevdp:
There's enough complication and subtlety in that without the distraction of how to coerce other iterable types.

jakevdp:
Once that's done and working, a new PR to discuss these more complicated questions would be useful. But the current discussion is hard to follow because it conflates two entirely separate issues.

DavidMertz:
Reopening with narrower concern in two separate PRs

PR #16, opened 2017-05-03
Add @ operator (simplify)

DavidMertz:
This supersedes the closed https://github.com/mrocklin/sparse/pull/14.

This request simply adds the `__matmul__` and `__rmatmul__` with no behavior other than that delegated to `sparse.dot()`.  Checks for "correct" non-combination of lists with COOs remains.  A more specific `return NotImplemented` is returned for things lacking an `.ndim`, which gives us a better error of:

```python
TypeError: unsupported operand type(s) for @: 'list' and 'COO'
```

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -802,6 +802,10 @@ def tensordot(a, b, axes=2):
 
 
 def dot(a, b):
+    if not hasattr(a, 'ndim') or not hasattr(b, 'ndim'):
+        raise NotImplementedError(
+                "Cannot perform dot product on types %s, %s" %
+                (type(a), type(b)))
     return tensordot(a, b, axes=((a.ndim - 1,), (b.ndim - 2,)))
 
 
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -395,21 +395,6 @@ class COO(object):
             strides *= d
         return out
 
-    def __matmul__(self, other):
-        try:
-            return dot(self, other)
-        except NotImplementedError:
-            return NotImplemented
-
-    def __rmatmul__(self, other):
-        try:
-            return dot(other, self)
-        except NotImplementedError:
-            return NotImplemented
-
-    def __numpy_ufunc__(self, ufunc, method, i, inputs, **kwargs):
-        return NotImplemented
-
     def reshape(self, shape):
         if self.shape == shape:
             return self
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -395,6 +395,21 @@ class COO(object):
             strides *= d
         return out
 
+    def __matmul__(self, other):
+        try:
+            return dot(self, other)
+        except NotImplementedError:
+            return NotImplemented
+
+    def __rmatmul__(self, other):
+        try:
+            return dot(other, self)
+        except NotImplementedError:
+            return NotImplemented
+
+    def __numpy_ufunc__(self, ufunc, method, i, inputs, **kwargs):
+        return NotImplemented
+
     def reshape(self, shape):
         if self.shape == shape:
             return self
--- a/sparse/tests/test_core.py
+++ b/sparse/tests/test_core.py
@@ -127,14 +127,68 @@ def test_tensordot(a_shape, b_shape, axes):
 
 
 def test_dot():
+    import operator
     a = random_x((3, 4, 5))
     b = random_x((5, 6))
+    lol = [[1, 2, 3, 4, 5]]
 
     sa = COO.from_numpy(a)
     sb = COO.from_numpy(b)
 
     assert_eq(a.dot(b), sa.dot(sb))
     assert_eq(np.dot(a, b), sparse.dot(sa, sb))
+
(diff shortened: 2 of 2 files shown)


sparse/tests/test_core.py
```diff
@@ -130,12 +131,43 @@ def test_dot():
     a = random_x((3, 4, 5))
     b = random_x((5, 6))
 
+    la = a.tolist()
+    lb = b.tolist()
+    la, lb          # silencing flake8
+
```
jakevdp: These are not used anymore, I think

sparse/tests/test_core.py
```diff
@@ -130,12 +131,43 @@ def test_dot():
...
     sa = COO.from_numpy(a)
     sb = COO.from_numpy(b)
 
     assert_eq(a.dot(b), sa.dot(sb))
     assert_eq(np.dot(a, b), sparse.dot(sa, sb))
 
+    if sys.version_info >= (3, 5):
```
jakevdp: Rather than an explicit version check, perhaps we could check ``hasattr(operator, 'matmul')``

sparse/tests/test_core.py
```diff
@@ -130,12 +131,43 @@ def test_dot():
...
+        # Basic equivalences
+        assert_eq(eval("a @ b"), eval("sa @ sb"))
+        assert_eq(eval("sa @ sb"), sparse.dot(sa, sb))
+
+        # Test that SOO's and np.array's combine correctly
+        assert_eq(eval("a @ sb"), eval("sa @ b"))
```
jakevdp: I think we could also check ``np.dot(sa, sb)``. If the classes are set up correctly, this should know how to delegate to ``sparse.dot`` (though I think it requires ``__numpy_ufunc__``)

sparse/core.py
```diff
@@ -681,6 +686,8 @@ def tensordot(a, b, axes=2):
 
 
 def dot(a, b):
+    if not hasattr(a, 'ndim') or not hasattr(b, 'ndim'):
+        return NotImplemented
```
jakevdp: I think this should be ``raise NotImplementedError``, because ``dot`` is often called on its own. You could catch the ``NotImplementedError`` and ``return NotImplemented`` within ``__matmul__`` if you want to use it.

sparse/tests/test_core.py
DavidMertz: Fixed.

sparse/tests/test_core.py
DavidMertz: Why? But OK, changed.

sparse/tests/test_core.py
DavidMertz: This gets a failure that doesn't seem straightforward to fix.  But moreover, I wouldn't really *expect* it to work as a user.

```python
sparse/core.py:448: in __mul__
    return self.elemwise_binary(operator.mul, other)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <COO: shape=(3, 4, 5), dtype=float64, nnz=6>
func = <built-in function mul>
other = <COO: shape=(5, 6), dtype=float64, nnz=4>, args = (), kwargs = {}

    def elemwise_binary(self, func, other, *args, **kwargs):
        assert isinstance(other, COO)
        if kwargs.pop('check', True) and func(0, 0, *args, **kwargs) != 0:
            raise ValueError("Performing this operation would produce "
                    "a dense result: %s" % str(func))
        if self.shape != other.shape:
>           raise NotImplementedError("Broadcasting is not supported")
E           NotImplementedError: Broadcasting is not supported

sparse/core.py:478: NotImplementedError
```

sparse/core.py
DavidMertz: Done.

DavidMertz:
I've pushed out the changes in response to review.

sparse/tests/test_core.py
jakevdp: Weird.. that error comes from the fact that ``np.dot(sa, sb)`` ends up calling ``sa.__mul__(sb)``, which... I have no idea why it would do that.

jakevdp:
I think this is good to merge, but first it needs to be rebased against master, and I also think the commit history should be cleaned up (though I'm not sure of @mrocklin's philosophy on that...)

mrocklin:
Clean commit history or just squashing would be fine.  Honestly though I
haven't taken great care with this project.

DavidMertz:
I'm lacking some git-fu to cleanup merge conflicts in rebasing and also prune commit history.  Would it be OK/reasonable if I simply open a new PR using a different branch of my fork (where hopefully I can clean the commit history)?

mrocklin:
Whatever works for you.  I apologize for being absent in this PR.  Normal
work duties are taking precedence.  I'll try to take a look at this tonight.

jakevdp:
Merged – thanks!

PR #18, opened 2017-05-03
Matmul/generic2

DavidMertz:
This is an alternative to the approach of https://github.com/mrocklin/sparse/pull/17/files.  No more than one of them should be accepted.

The gist of this is that—in keeping with NumPy and scipy.sparse behavior—subclasses of `list` or `tuple` will be automatically converted to NumPy (dense) arrays if they are passed as an argument to `sparse.dot()`.  Any other type of object lacking an `.ndim` will simply cause an exception to be raised.

In contrast to PR 17, this allows no opt-out or opt-in from the coercion behavior.  However, since the initial thing coerced is always a (dense) list/tuple, it cannot be too big to represent in memory (and the `np.array` version will only be smaller, albeit an additional object in memory).

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -328,6 +328,11 @@ class COO(object):
     def dot(self, other):
         return dot(self, other)
 
+    __matmul__ = dot
+
+    def __rmatmul__(self, other):
+        return dot(other, self)
+
     def reshape(self, shape):
         if self.shape == shape:
             return self
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -326,13 +326,12 @@ class COO(object):
         return self.transpose(list(range(self.ndim))[::-1])
 
     def dot(self, other):
-        return dot(self, other)
-
-    def __matmul__(self, other):
         try:
             return dot(self, other)
-        except AttributeError:
+        except AttributeError:  # missing .ndim
             return dot(self, np.array(other))
+
+    __matmul__ = dot
 
     def __rmatmul__(self, other):
         return dot(np.array(other), self)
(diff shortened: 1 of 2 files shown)


jakevdp:
I think this can be solved much more generally by using ``np.asanyarray`` after checking for either explicitly-supported types or checking for ducktypes. You'd then have to check that the result of ``np.asanyarray`` has an appropriate dtype and shape, and if not, raise an appropriate error. That would allow this to be used with *any* Python object that exposes the buffer interface. That's the solution used by ``scipy.sparse``.

The downside is that this would not allow people to define other kinds of sparse objects that would work with COO matrices... we could solve that by specifying an ``__array_priority__`` of, say, "10.2", so it would be just above that of ``scipy.sparse``. Then someone wanting to extend this behavior could set an even higher priority if they want to control ``dot`` interactions with their own objects.

jakevdp:
But ``__array_priority__`` may be superseded by ``__numpy_ufunc__``, though I'm not sure...

mrocklin:
Note also that __numpy_ufunc__ may be changing to __array_ufunc__
https://github.com/numpy/numpy/pull/8247

DavidMertz:
I've pushed the small change in the review of #16 that is relevant to this change.  If you are fine with this approach rather than the opt-in approach, just merge this (but close without merging #16 and #17)

mrocklin:
OK to close this?

DavidMertz:
I still think it would be good to coerce lists/tuples as NumPy and scipy.sparse do. Closing the keyword opt-in in #17 seems reasonable since neither of you liked it. It's up to you, of course.

If you want this, let me rebase this against the accepted PR and remove the unnecessary change history. The diff should just be a few lines that way.

PR #21, opened 2017-05-09
Add @ operator, no coersion

DavidMertz:
This replaces #16.  I did not have the git-fu to prune the commit history under my fork's master.  This has the same changes, but in a new branch that is rebased with cleaner commits.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -802,6 +802,10 @@ def tensordot(a, b, axes=2):
 
 
 def dot(a, b):
+    if not hasattr(a, 'ndim') or not hasattr(b, 'ndim'):
+        raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


DavidMertz:
Oh well... looks like @jakevdp merged the branch with the messier commit history while I was learning the maddening mysteries of `git rebase -i` (which is great except when it fails to merge a bunch of the commits).

I think if you apply this it will still clean the commit history (but should otherwise be identical code changes).  But I am way less than confident in my understanding of git, so use your judgement.

jakevdp:
Oh, sorry about that. I did a squash and merge on the other one so it's just a single commit. FWIW, for doing this kind of cleaning of commit history, I tend to use ``git reset --soft`` and then re-commit the changes manually. Thanks!

PR #23, opened 2017-08-05
Improve the source release for PyPI

ghisvail:
The current tarball distributed on PyPI is missing some files required for producing quality binary packages for Linux distributions.

In addition to the `README.rst` and package content, which `sdist` handles by default, this commit adds the licensing terms (also added by default if the file were named `LICENSE`) and the test suite, which is helpful for continuous integration purposes.


mrocklin:
Thanks @ghisvail !

PR #26, opened 2017-08-16
implemented __setitem__ for single items plus a test for it

### LINKED ISSUE #15: Item Assignment and Integer Slicing
Hi,

I found your library very useful, but I get a few questions for you. I wonder what is best way to add/delete data from the array? Also, is it possible to index a value using integer indexing?

```python
x = sparse.COO(coords, data, shape=((3,3,3))

#Is there a way to do achieve this?
x[2,2,2] = 2 #Cannot do this
```

Thanks.

bigerl:
this partly resolves #15

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -292,7 +292,6 @@ class COO(object):
         value
             the value to write
         """
-        # TODO: what if coordinates are too big for dtype of coords?
         # TODO: check input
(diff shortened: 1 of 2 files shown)


bigerl:
Should be mature now. Unfortunately I don't have time right now to support assigning slices as well .

PR #28, opened 2017-08-22
Elementwise and and or

### LINKED ISSUE #25: elementwise_binary doesn't return correct results
I'm using the core.elementwise_binary to set values in a tensor. But it doesn't return the expected result. Unfortunately I do not really understand the code of the function otherwise I would just have made a pull request. Here is a minimal example to reproduce the bug:

```
import sparse as sp
import operator
import numpy as np

size = 3
dimension = 3

coords1 = np.ndarray(shape = 0, dtype=bool)
data1 = np.ndarray(shape = 0, dtype=bool)
tensor1 = sp.COO(coords=coords1, data=data1, shape=((size,) * dimension))
print('emptyTensor')
print(tensor1)
print(tensor1.todense())

coords2 = np.asarray(a=[[0], [1], [2]])
data2 = np.asarray(a=[True], dtype=bool)
tensor2 = sp.COO(coords=coords2, data=data2, shape=((size,) * dimension))
print('\noneElementTensor')
print(tensor2)
print(tensor2.todense())

resultTensor = tensor1.elemwise_binary(operator.or_, tensor2)
print('\nresultTensor = emptyTensor | oneElementTensor')
print(resultTensor)
print(resultTensor.todense())
```

The `resultTensor` should be equal to `tensor2`(`True` at `tensor2[1,2,3]`, all other fields of tensor2 `False`) but it is all fields `False`. 
Additionally I would expect the `dtype` of `resultTensor` to be `bool` because both input `COO`s `tensor1`and `tensor2` are `dtype=bool` and `operator.or_` is generally expected to be a `bool`×`bool`→`bool` operation.

Output:
```
emptyTensor
<COO: shape=(3, 3, 3), dtype=bool, nnz=0, sorted=False, duplicates=True>
[[[False False False]
  [False False False]
  [False False False]]

 [[False False False]
  [False False False]
  [False False False]]

 [[False False False]
  [False False False]
  [False False False]]]

oneElementTensor
<COO: shape=(3, 3, 3), dtype=bool, nnz=1, sorted=False, duplicates=True>
[[[False False False]
  [False False  True]
  [False False False]]

 [[False False False]
  [False False False]
  [False False False]]

 [[False False False]
  [False False False]
  [False False False]]]

resultTensor = emptyTensor | oneElementTensor
<COO: shape=(3, 3, 3), dtype=int64, nnz=0, sorted=False, duplicates=False>
[[[0 0 0]
  [0 0 0]
  [0 0 0]]

 [[0 0 0]
  [0 0 0]
  [0 0 0]]

 [[0 0 0]
  [0 0 0]
  [0 0 0]]]
```

bigerl:
This resolves #25. 
Additionally it introduce the element-wise binary operators AND `&`, OR `|` and XOR `^` on any COO holding Integral values (that includes especially long, int and bool).

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -8,7 +8,6 @@ import operator
 import numpy as np
 import scipy.sparse
 
-
 try:  # Windows compatibility
     int = long
 except NameError:
@@ -589,6 +588,15 @@ class COO(object):
 
     def __pow__(self, other):
         return self.elemwise(operator.pow, other)
+
+    def __and__(self, other):
+        return self.elemwise_binary(operator.and_, other)
+
+    def __or__(self, other):
+        return self.elemwise_binary(operator.or_, other)
+
+    def __xor__(self, other):
+        return self.elemwise_binary(operator.xor, other)
 
     def elemwise(self, func, *args, **kwargs):
         if kwargs.pop('check', True) and func(0, *args, **kwargs) != 0:
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -611,26 +611,38 @@ class COO(object):
 
         # Sort self.coords in lexographical order using record arrays
         self_coords = np.rec.fromarrays(self.coords)
-        i = np.argsort(self_coords)
-        self_coords = self_coords[i]
-        self_data = self.data[i]
+        self_sort_order = np.argsort(self_coords)
+        self_coords = self_coords[self_sort_order]
+        self_data = self.data[self_sort_order]
 
-        # Convert other.coords to a record array
+        # Sort other.coords in lexographical order using record arrays
         other_coords = np.rec.fromarrays(other.coords)
-        other_data = other.data
+        other_sort_order = np.argsort(other_coords)
+        other_coords = other_coords[other_sort_order]
+        other_data = other.data[other_sort_order]
 
         # Find matches between self.coords and other.coords
-        j = np.searchsorted(self_coords, other_coords)
-        if len(self_coords):
-            matched_other = (other_coords == self_coords[j % len(self_coords)])
-        else:
-            matched_other = np.zeros(shape=(0,), dtype=bool)
-        matched_self = j[matched_other]
+        # TODO: implement a modified exponential search to find next fitting coordinate faster
+        matched_self = np.zeros(shape=(len(self_coords),), dtype=bool)
+        matched_other = np.zeros(shape=(len(other_coords),), dtype=bool)
+        i = 0
+        j = 0
+        while (i < len(self_coords)) & (j < len(other_coords)):
+            self_coord = tuple(self_coords[i])
+            other_coord = tuple(other_coords[j])
+            if self_coord == other_coord:
+                matched_self[i] = True
+                matched_other[j] = True
+                i += 1
+                j += 1
+            elif self_coord < other_coord:
+                i += 1
+            else:
+                j += 1
 
         # Locate coordinates without a match
+        unmatched_self = ~matched_self
         unmatched_other = ~matched_other
-        unmatched_self = np.ones(len(self_coords), dtype=bool)
-        unmatched_self[matched_self] = 0
 
         # Concatenate matches and mismatches
         data = np.concatenate([func(self_data[matched_self],
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -648,9 +648,9 @@ class COO(object):
         data = np.concatenate([func(self_data[matched_self],
                                     other_data[matched_other],
                                     *args, **kwargs),
-                               func(self_data[unmatched_self], 0,
+                               func(self_data[unmatched_self], False,
                                     *args, **kwargs),
-                               func(0, other_data[unmatched_other],
+                               func(False, other_data[unmatched_other],
                                     *args, **kwargs)])
         coords = np.concatenate([self_coords[matched_self],
                                  self_coords[unmatched_self],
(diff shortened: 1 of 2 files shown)


sparse/core.py
```diff
@@ -611,34 +619,46 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
+            other_coord = tuple(other_coords[j])
+            if self_coord == other_coord:
+                matched_self[i] = True
+                matched_other[j] = True
+                i += 1
+                j += 1
+            elif self_coord < other_coord:
+                i += 1
+            else:
+                j += 1
```
mrocklin: What is the performance impact of this?  I would expect iterating through a numpy array and creating many tuples to be somewhat expensive.

sparse/tests/test_core.py
```diff
@@ -201,6 +208,23 @@ def test_elemwise_binary_empty():
...
+def test_elemwise_binary_OR():
+    size = 3
+    dimension = 3
+    shape = ((size,) * dimension)
+
+    x = np.zeros(shape=shape,dtype=bool)
+    xx = COO.from_numpy(x)
+
+    y = np.zeros(shape=shape,dtype=bool)
+    y[0,1,2] = True
```
mrocklin: It might be wise to test numeric values as well.  The `|` and `&` operators in NumPy perform bitwise operations.

sparse/core.py
```diff
@@ -611,34 +619,46 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
 
         # Concatenate matches and mismatches
         data = np.concatenate([func(self_data[matched_self],
                                     other_data[matched_other],
                                     *args, **kwargs),
-                               func(self_data[unmatched_self], 0,
+                               func(self_data[unmatched_self], False,
                                     *args, **kwargs),
-                               func(0, other_data[unmatched_other],
+                               func(False, other_data[unmatched_other],
```
mrocklin: Why these changes?

bigerl: The result of a `operator.*` function has the data type of the input with the largest domain. So if you put `0` instead of `False` then `operator.and_(0, False)`will return 0 of data type `int` resulting in an COO with `dtype=int64`instead of the expected `dtype=bool`

sparse/core.py
bigerl: Surly it is not optimal in terms of performance. But the performance should be about the same as running `np.rec.fromarrays(self.coords)` plus `np.rec.fromarrays(other.coords)`. So in the worst case it may affects the performance by about a constant factor 2. 
Generally: his was more a fast approach to fix #25. This part of the code is for my own project quite performance critical. So I will rewrite this section about next week with focused on asymptotic runtime and memory consumption.

sparse/tests/test_core.py
bigerl: So does `^`. You're right, I'll just patch that.

sparse/core.py
mrocklin: > But the performance should be about the same as running np.rec.fromarrays(self.coords) plus np.rec.fromarrays(other.coords). So in the worst case it may affects the performance by about a constant factor 2.

How did you come to this conclusion?  Have you compared?  Generally I would expect Python code that generated many tuples to be much slower than NumPy's C code.

sparse/tests/test_core.py
```diff
@@ -392,6 +416,66 @@ def test_scalar_exponentiation():
...
+
+
+def test_int_xor():
+    x = random_x((2, 3, 4), dtype=int)
+    xx = COO.from_numpy(x)
+
+    y = random_x((2, 3, 4), dtype=int)
+    yy = COO.from_numpy(y)
+
+    assert_eq(x ^ y, xx ^ yy)
```
mrocklin: It looks like these tests can be merged using `pytest.mark.parametrize('op', [operator.xor, operator.or_, operator.and_, ...])`

sparse/core.py
mrocklin: This sounds like an asymptotic argument.  Have you actually benchmarked and tested the speed difference here?

What is this change supposed to do?  Is it necessary to handle the elemwise and and or operators?

bigerl: Of course it is an asymptotic argument and I didn't benchmark it. Sorry, I didn't have time for that then.

Anyways the change from line 634 to 649 is not to handle elemwise AND and OR operations but to handle any elemwise operations correctly.

If you take a look at the example from #25.

It takes two tensors - one empty and one with a True at (0, 1, 2) - and calculates the elemwise OR. The result is an empty tensor. But it should be a tensor with an True or 1 at (0, 1, 2).

You get the same result if you take equivalent int or float valued input tensors and ADD as operation. Result is again an empty (all 0) tensor. So there is definitely a bug in the coords matching between the two input tensors.

I know this is not the most beautiful solution but at least it returns correct results for now.

sparse/tests/test_core.py
bigerl: I'll have a look at it next week.

sparse/core.py
hameerabbasi: Technically, I think we should create two zero-elements as follows:

```zero_self = np.zeros((), shape=(), dtype=self.dtype)[()]
zero_other = np.zeros((), shape=(), dtype=other.dtype)[()]```

And use these. I would take into account any casts done by the function correctly. Using `False` or `0` may still be fragile.

hameerabbasi:
@bigerl Is it okay to close this since it was addressed in #38?

PR #29, opened 2017-08-22
implemented __setitem__

### LINKED ISSUE #15: Item Assignment and Integer Slicing

bigerl:
this partly resolves #15. It supports setting values in COOs like:
```
x = COO.from_numpy(np.zeros(shape=(5,5)))
x[1,2] = 5
```
It does not support slices so far. So values aren't set in form of structures like arrays but as simply as scalars.

The index format is compatible with `np.ndarray`but does not match `sparse.COO`'s format for get: 
```
value = x[[1],[2]]
```
That leads to the following unequitation:
```
x = COO.from_numpy(np.zeros(shape=(5,5)))
value = 1
x[1,2] = value 
get_value = x[[1],[2]]
assert value != get_value
``` 
here value is an int and get_value is an COO of shape (1,1).

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -292,7 +292,6 @@ class COO(object):
         value
             the value to write
         """
-        # TODO: what if coordinates are too big for dtype of coords?
         # TODO: check input
         # search for index in coords and get it's position if present
         # TODO: speedup by not checking every possible position in every dimension
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -287,22 +287,49 @@ class COO(object):
         Sets the single field at index to the given value.
         Parameters
         ----------
-        index : tuple
-            The coordinates of the field to write the new value to. Must be a tuple.
+        index : tuple of int
+            The coordinates of the field to write the new value to. Slices are not jet supported.
         value
             the value to write
         """
-        # TODO: check input
-        # search for index in coords and get it's position if present
+        # check input
+        # convert single valued inputs to tuples
+        if not isinstance(index, tuple):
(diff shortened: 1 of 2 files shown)


mrocklin:
This is, I suspect, very slow.  If we want to support setitem workloads efficiently then we should probably consider other data structures, like dictionary-of-keys and then provide methods to convert between different kinds of sparse arrays.  This is the approach taken by the `scipy.sparse` modules with COO/DOK/CSR formats

bigerl:
Yes, I suspect you're right about that. Copying all entries as well as all coordinates for every change between zero and non-zero values is very slow. 
Again here: I needed the interface so wrote a quick and dirty solution against what was there. 
The dictionary of keys sounds like a good idea especially because you can efficiently converted to COO. 

But unfortunately I don't have time to read-in on the topic and to implement it right now. But maybe I could spend some time on it around Oktober. Then I'll start my master thesis which will be using sparse.

mrocklin:
OK, I'm going to close this for now.  If you have time to implement DOK at some point that would be *very welcome*.   We probably shouldn't support this operation with the current data structures.

Thanks for starting this discussion, I hope that it leads to improvements in the future!

PR #30, opened 2017-08-22
reshape without coords scaling

bigerl:
Sometimes it is preferable to keep the coordinates of entries when reshaping a sparse multidimensional array. 
That for I propose adding the flag `scale={True,False}` to "COO.reshape". 

When set to `True` or by default `COO.reshape` shows normal `np.reshape`behaviour:
```
coords = [[1],[1]]
data = [1]
x = COO(data=data, coords=coords, shape=(3, 3))
x = x.reshape(shape=(4, 4)) #  equal to: x.reshape(shape=(4, 4), scale = True)
assert (x[[1], [1]].todense()[0,0] == 0)
assert ((x[[1], [0]].todense()[0,0] == 1))
```
But when set to `True`the coordinates of data points are preserved:
```
coords = [[1],[1]]
data = [1]
x = COO(data=data, coords=coords, shape=(3, 3)) #diff
x = x.reshape(shape=(4, 4), scale = True)
assert (x[[1], [1]].todense()[0,0] == 1) #diff
assert ((x[[1], [0]].todense()[0,0] == 0)) #diff
```

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -282,6 +282,54 @@ class COO(object):
                    has_duplicates=self.has_duplicates,
                    sorted=self.sorted)
 
+    def __setitem__(self, index, value):
+        """
+        Sets the single field at index to the given value.
+        Parameters
+        ----------
+        index : tuple
+            The coordinates of the field to write the new value to. Must be a tuple.
+        value
+            the value to write
+        """
+        # TODO: what if coordinates are too big for dtype of coords?
+        # TODO: check input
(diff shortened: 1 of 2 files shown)


mrocklin:
Does this API exist elsewhere, such as in NumPy or scipy.sparse?  I'm hesitant to break with the standard ndarray API here.

bigerl:
No, it doesn't. And realized that it is actually more about broadcasting when adding values at coordinates outside the shape. So I'll close that one.

PR #35, opened 2017-12-16
Added elementwise_binary broadcasting and broadcast_to

hameerabbasi:
I've tested this on Python 3.6.3, with unit tests and flake8.

Some formatting changes occurred because I used PyCharm's reformat file.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -862,6 +862,7 @@ class COO(object):
     # (c) senderle
     # Taken from https://stackoverflow.com/a/11146645/774273
     # License: https://creativecommons.org/licenses/by-sa/3.0/
+    @staticmethod
     def _cartesian_product(*arrays):
         broadcastable = np.ix_(*arrays)
         broadcasted = np.broadcast_arrays(*broadcastable)
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -727,8 +727,11 @@ class COO(object):
 
         nonzero = data != 0
         data = data[nonzero]
-        if len(data) != 0:
-            coords, idx = np.unique(coords[:, nonzero], axis=1, return_index=True)
+        coords = coords[:, nonzero]
+        if len(data) != 0 and len(self_reduced_coords) == 0:
+            linear_coords = self._linear_loc(coords, result_shape)
+            linear_coords, idx = np.unique(linear_coords, return_index=True)
+            coords = coords[:, idx]
             data = data[idx]
 
         return COO(coords, data, shape=result_shape, has_duplicates=False)
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -636,7 +636,7 @@ class COO(object):
         self_coords = self_coords[:, i]
         self_data = self_data[i]
 
-        # Convert other.coords to a record array
+        # Store coords
         other_coords = other.coords
         other_data = other.data
 
@@ -815,7 +815,7 @@ class COO(object):
         Expand coordinates/data to broadcast_shape. Does most of the heavy lifting for broadcast_to.
         Parameters
         ----------
-        coords : np.recarray
+        coords : np.ndarray
         expanded_data : np.ndarray
         params : list
         broadcast_shape : tuple[int]
@@ -875,7 +875,7 @@ class COO(object):
 
         Parameters
         ----------
-        coords1, coords2 : np.recarray
+        coords1, coords2 : np.ndarray
         shape1, shape2 : tuple[int]
 
         Returns
@@ -914,7 +914,7 @@ class COO(object):
 
         Parameters
         ----------
-        coords : np.recarray
+        coords : np.ndarray
         matched_coords : np.ndarray
         data : np.ndarray
         params : list
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -664,7 +664,6 @@ class COO(object):
         coords_list = []
 
         # Add the matched part.
-        # if func(_TEST_NONZERO, _TEST_NONZERO_2):
         matched_coords = self._get_matching_coords(self_coords[:, matched_self],
                                                    other_coords[:, matched_other],
                                                    self_shape, other_shape)
@@ -688,6 +687,7 @@ class COO(object):
             coords_list.append(self_unmatched_coords)
 
             if self_shape != result_shape:
+                matched_self = np.logical_not(unmatched_self)
                 self_broadcast_coords, self_broadcast_func = \
                     self._get_broadcast_coords_data(self_coords[:, matched_self],
                                                     matched_coords,
@@ -711,6 +711,7 @@ class COO(object):
             data_list.append(other_unmatched_func)
 
             if other_shape != result_shape:
+                matched_other = np.logical_not(unmatched_other)
                 other_broadcast_coords, other_broadcast_func = \
                     self._get_broadcast_coords_data(other_coords[:, matched_other],
                                                     matched_coords,
@@ -827,26 +828,51 @@ class COO(object):
         expanded_data : np.ndarray
             The data corresponding to expanded_coords.
         """
-        times_repeated = 1
-        expanded_coords = []
-        expanded_data = data
-        dim = 0
-        data_len = len(expanded_data)
-        dt = tuple(np.min_scalar_type(l) for l in broadcast_shape)
-        dt = np.result_type(*dt)
-        for param, l in zip(params, broadcast_shape):
-            if param:
-                expanded_coords.append(np.repeat(coords[dim], times_repeated))
-            else:
-                expanded_data = np.repeat(expanded_data, l)
-                expanded_coords = [np.repeat(coord, l) for coord in expanded_coords]
-                expanded_coords.append(np.tile(np.arange(l, dtype=dt), times_repeated * data_len))
-                times_repeated *= l
+        first_dim = -1
+        expand_shapes = []
+        for d, p, l in zip(range(len(broadcast_shape)), params, broadcast_shape):
+            if p and first_dim == -1:
+                expand_shapes.append(coords.shape[1])
+                first_dim = d
 
-            if param is not None:
+            if not p:
+                expand_shapes.append(l)
+
+        all_idx = COO._cartesian_product(*tuple(np.arange(d, dtype=np.min_scalar_type(d - 1)) for d in expand_shapes))
+        dt = np.result_type(*tuple(np.min_scalar_type(l - 1) for l in broadcast_shape))
+
+        false_dim = 0
+        dim = 0
+
+        expanded_coords = np.empty((len(broadcast_shape), all_idx.shape[1]), dtype=dt)
+        expanded_data = data[all_idx[first_dim]]
+
+        for d, p, l in zip(range(len(broadcast_shape)), params, broadcast_shape):
+            if p:
+                expanded_coords[d] = coords[dim, all_idx[first_dim]]
+            else:
+                expanded_coords[d] = all_idx[false_dim + (d > first_dim)]
+                false_dim += 1
+
+            if p is not None:
                 dim += 1
 
         return np.asarray(expanded_coords), np.asarray(expanded_data)
+
+    # (c) senderle
+    # Taken from https://stackoverflow.com/a/11146645/774273
+    # License: https://creativecommons.org/licenses/by-sa/3.0/
+    def _cartesian_product(*arrays):
+        broadcastable = np.ix_(*arrays)
+        broadcasted = np.broadcast_arrays(*broadcastable)
+        rows, cols = np.prod(broadcasted[0].shape), len(broadcasted)
+        dtype = np.result_type(*arrays)
+        out = np.empty(rows * cols, dtype=dtype)
+        start, end = 0, rows
+        for a in broadcasted:
+            out[start:end] = a.reshape(-1)
+            start, end = end, end + rows
+        return out.reshape(cols, rows)
 
     def broadcast_to(self, shape):
         """
@@ -927,16 +953,8 @@ class COO(object):
             The data corresponding to those coordinates.
         """
         full_coords, full_data = COO._get_expanded_coords_data(coords, data, params, broadcast_shape)
-
-        # TODO: Remove this sort. Only needed in one test case where no dimensions of
-        # the operands match. Will need to find out what's causing it.
-        # Not a high priority as the _match_coords call is O(n log n) anyway.
         linear_full_coords = COO._linear_loc(full_coords, broadcast_shape)
         linear_matched_coords = COO._linear_loc(matched_coords, broadcast_shape)
-        i = np.argsort(linear_full_coords)
-        linear_full_coords = linear_full_coords[i]
-        full_coords = full_coords[:, i]
-        full_data = full_data[i]
 
         overlapping_coords, _ = _match_coords(linear_full_coords, linear_matched_coords)
         mask = np.ones(full_coords.shape[1], dtype=np.bool)
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -638,6 +638,7 @@ class COO(object):
         other_coords = np.rec.fromarrays(other.coords)
         other_reduced_coords = np.rec.fromarrays(self._get_reduced_coords(other.coords, other_reduce_params))
         i = np.argsort(other_reduced_coords)
+        other_coords = other_coords[i]
         other_reduced_coords = other_reduced_coords[i]
         other_data = other.data[i]
 
@@ -657,8 +658,8 @@ class COO(object):
         # Add the matched part.
         # if func(_TEST_NONZERO, _TEST_NONZERO_2):
         matched_coords = self._get_matching_coords(self_coords[matched_self],
-                                                        other_coords[matched_other],
-                                                        self_shape, other_shape)
+                                                   other_coords[matched_other],
+                                                   self_shape, other_shape)
 
         data_list.append(func(self_data[matched_self],
                               other_data[matched_other],
@@ -714,7 +715,7 @@ class COO(object):
         data = data[nonzero]
         coords = coords[:, nonzero]
 
-        return COO(coords, data, shape=self.shape, has_duplicates=False)
+        return COO(coords, data, shape=result_shape, has_duplicates=False)
 
     @staticmethod
     def _get_broadcast_shape(shape1, shape2):
@@ -756,12 +757,14 @@ class COO(object):
         for param, l in zip(params, broadcast_shape):
             if param:
                 coords_list.append(np.repeat(coords[names[dim]], times_repeated))
-                dim += 1
             else:
                 data = np.repeat(data, l)
                 coords_list = [np.repeat(coord, l) for coord in coords_list]
                 coords_list.append(np.tile(np.arange(l), times_repeated * data_len))
                 times_repeated *= l
+
+            if param is not None:
+                dim += 1
 
         return coords_list, data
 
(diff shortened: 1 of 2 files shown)


sparse/core.py
```diff
@@ -603,55 +613,238 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
+
+            if p2 is not None:
+                dim2 += 1
+
+        return coords_list
+
+    @staticmethod
+    def _get_broadcast_coords_data(coords, matched_coords, data, params, broadcast_shape):
+        full_coords, full_data = COO._get_expanded_coords_data(coords, data, params, broadcast_shape)
+        full_coords = np.rec.fromarrays(full_coords)
```
mrocklin: Why `np.rec.from_arrays` here?  Is this a record array?  If so, why?

sparse/core.py
```diff
@@ -603,55 +613,238 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
+
+    @staticmethod
+    def _get_broadcast_shape(shape1, shape2, is_result=False):
+        """
+        Get the overall broadcasted shape from two different shapes, without actually constructing any arrays.
+        """
+        # https://stackoverflow.com/a/47244284/774273
+        if not all((l1 == l2) or (l1 == 1) or ((l2 == 1) and not is_result) for l1, l2 in zip(shape1[::-1],
+                                                                                             shape2[::-1])):
+            raise ValueError('operands could not be broadcast together with shapes %s, %s' % (shape1, shape2))
```
mrocklin: Can I request that we wrap code shorter than this when possible?  We allow lines up to 120 in flake8, but in practice we try to aim shorter.  120 is just as lenient as we're willing to be, not our target.

sparse/core.py
```diff
@@ -603,55 +613,238 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
+        return params
+
+    @staticmethod
+    def _get_reduced_coords(coords, params):
+        reduced_params = [bool(param) for param in params]
+
+        return coords[reduced_params]
+
+    @staticmethod
+    def _get_expanded_coords_data(coords, data, params, broadcast_shape):
```
mrocklin: The purpose of this function isn't immediately clear to me.  Can I ask you for a docstring to aid review?  This would generally be appreciated for similar methods as well.

sparse/core.py
```diff
@@ -603,55 +613,238 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
+
+        return coords_list
+
+    @staticmethod
+    def _get_broadcast_coords_data(coords, matched_coords, data, params, broadcast_shape):
+        full_coords, full_data = COO._get_expanded_coords_data(coords, data, params, broadcast_shape)
+        full_coords = np.rec.fromarrays(full_coords)
+        matched_coords = np.rec.fromarrays(matched_coords, dtype=full_coords.dtype)
+
+        matched_coords.sort()
```
mrocklin: Should we be concerned about the cost of this sort here?  Is it necessary?  Are there any cases where it can be avoided?

mrocklin: Generally it might be useful to build a couple of benchmarks and then profile them to see if we're introducing any slow processes.

sparse/core.py
```diff
@@ -746,7 +939,7 @@ def tensordot(a, b, axes=2):
     # Please see license at https://github.com/numpy/numpy/blob/master/LICENSE.txt
     try:
         iter(axes)
-    except:
+    except TypeError:
```
mrocklin: Thanks!

sparse/core.py
```diff
@@ -912,3 +1105,33 @@ def stack(arrays, axis=0):
...
+    al, bl = np.diff(asw), np.diff(bsw)
+    na = len(al)
+    asw, bsw = asw, bsw
+    abunq = np.r_[a[asw[:-1]], b[bsw[:-1]]]
+    m = np.argsort(abunq, kind='mergesort')
+    mv = abunq[m]
+    midx = np.flatnonzero(mv[:-1] == mv[1:])
+    ai, bi = m[midx], m[midx + 1] - na
+    aic = np.r_[0, np.cumsum(al[ai])]
+    a_idx = np.ones((aic[-1],), dtype=int)
```
mrocklin: Do we want to specify a particular dtype here?  Do we mind this being different on different platforms?

sparse/tests/test_core.py
```diff
@@ -359,15 +358,31 @@ def test_addition_not_ok_when_large_and_sparse():
...
+
+
+@pytest.mark.parametrize('shape1,shape2', [((3, 4), (2, 3, 4)),
+                                           ((3, 1, 4), (3, 2, 4)),
+                                           ((3, 4, 1), (3, 4, 2))])
+def test_broadcast_to(shape1, shape2):
+    x = random_x(shape1)
+    a = COO.from_numpy(x)
+
+    assert_eq(np.broadcast_to(x, shape2), a.broadcast_to(shape2))
```
mrocklin: These tests are quite nice and establishing confidence!  Are there any other things we should check on the resulting arrays?  For example the number of non-zeroes, if sortedness is preserved (or should it be?) etc..

sparse/core.py
```diff
@@ -553,10 +563,10 @@ def __add__(self, other):
             return self
         if not isinstance(other, COO):
             return self.maybe_densify() + other
-        if self.shape == other.shape:
+        else:  # if self.shape == other.shape:
             return self.elemwise_binary(operator.add, other)
-        else:
-            raise NotImplementedError("Broadcasting not yet supported")
+        # else:
+        #     raise NotImplementedError("Broadcasting not yet supported")
```
mrocklin: Code like this should probably be removed.

sparse/core.py
```diff
@@ -603,55 +613,238 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
-            raise NotImplementedError("Broadcasting is not supported")
+                             "a dense result: %s" % str(func))
+        self_shape, other_shape = self.shape, other.shape
+
+        result_shape = self._get_broadcast_shape(self_shape, other_shape)
+        self_params = self._get_broadcast_parameters(self.shape, result_shape)
+        other_params = self._get_broadcast_parameters(other.shape, result_shape)
+        combined_params = [p1 and p2 for p1, p2 in zip(self_params, other_params)]
+        self_reduce_params = combined_params[-self.ndim:]
+        other_reduce_params = combined_params[-other.ndim:]
```
mrocklin: Do any of the extra steps that we do meaningfully affect performance in the common case?

hameerabbasi: I haven't carried out a full benchmark, but asymptotically, no. I'll carry out a full benchmark soon and post the results, hopefully by tomorrow.

sparse/core.py
hameerabbasi: Will do this soon.

sparse/core.py
hameerabbasi: Sure, I'll try to cut it down to 80.

sparse/core.py
hameerabbasi: I'll add doctstrings for all methods.

sparse/core.py
hameerabbasi: It turns out that `matched_coords` and `full_coords` aren't necessarily sorted on input to this function. To perform the sorting, I convert them to `recarray`s, because we need to perform matching.

sparse/core.py
hameerabbasi: Yes, `_match_coords` expects sorted arrays, in particular `recarrays`. It's the reason why it matches so fast. In principle, the matching can be made O(n) on sorted arrays if there are no repeats, but on unsorted ones, it's O(n^2). In practice, the O(n log n) solution is much faster. See [my StackOverflow question](https://stackoverflow.com/questions/47831287/get-index-of-matching-elements-in-array-taking-into-account-repetitions) on the topic.

Should be benchmarks be committed or just for testing? For sure, I'll try to roll this back to the last release and perform a benchmark.

I just thought that, for maintainability, it would be easier to treat the non-broadcast case as a special version of the broadcast case. If we change something down the line, both versions would need to be changed.

sparse/core.py
hameerabbasi: I haven't dived into how this function works, but in the worst case it would be `np.min_scalar_type(len(a))`, same for the `b_idx` case for down below.

sparse/tests/test_core.py
hameerabbasi: Well, the number of non-zeros is already minimal (see lines 730, 731). I could add that to the test, if you like.

As a closing note, do I close this pull request and open a new one when I've done the changes, will the changes be automatically merged into this pull request, ...? I hope you don't mind me asking, this is actually my first time contributing to a project without being approved, and I've only ever made one pull request before.

mrocklin:
You can push additional changes to your current branch on GitHub and the PR
will update automatically.

sparse/core.py
hameerabbasi: I just saw the function again. It seems we really do need positive and negative numbers, and I have no idea how big or small they need to be. I'll make them `np.int_` for now.

sparse/core.py
hameerabbasi: It seems my version is 2x slower and most of the time is taken by `np.argsort` in the base case. I am considering moving away from `np.recarray`s and onto simple `np.ndarray`s, and from `np.argsort` to `np.lexsort`. This will require me to understand and modify the function `_match_coords` but the rest should follow. Before I do this though, I should probably evaluate the two against each other and actually confirm `np.lexsort` is indeed faster.

hameerabbasi: It seems `np.lexsort` is indeed faster, and by a large margin.
```
import numpy as np
x = np.random.rand(2, 10**7)
x_rec = np.rec.fromarrays(x)
%timeit np.lexsort(x)
4.59 s ± 65.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit np.argsort(x_rec)
10.7 s ± 91.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

sparse/core.py
```diff
@@ -8,12 +8,21 @@
...
+    from itertools import izip_longest as zip_longest
 
 try:  # Windows compatibility
     int = long
 except NameError:
     pass
 
+# Generated once with np.random.rand(). Used for checking if an elemwise function returns zero for specific inputs.
+_TEST_NONZERO = 0.8173885343707892
+_TEST_NONZERO_2 = 0.32621130891611305
```
hameerabbasi: I needed your input here. Right now, I'm just checking against random values if the function is nonzero for specific combinations. I could check with `(func(self.data, 0, *args, **kwargs) == 0).all()` and `(func(0, other.data, *args, **kwargs) == 0).all`. Do you think that's better? (It's algorithmically correct, but can take longer).

Alternately we can store these results somewhere and use them.

sparse/core.py
mrocklin: You might want to look at the tricks we do in operations like `linear_loc` to avoid making copies, allocating arrays with dtypes that are as small as possible, and so on.

sparse/core.py
mrocklin: This seems fragile to me.  For example what happens if the function has different behavior for positive and negative numbers?

sparse/tests/test_core.py
mrocklin: > Well, the number of non-zeros is already minimal (see lines 730, 731). I could add that to the test, if you like.

Yes, you should definitely test things like this.  You may know that it's true now, but other developers may come by and change your code.  You want them to see failed tests if they change something important.

sparse/core.py
mrocklin: When you get a chance it would be useful to include the examples you're using to benchmark.

hameerabbasi: I've improved the performance quite a bit, now this branch outperforms your `master` branch.

```
import numpy as np
from sparse import COO
from sparse.tests.test_core import random_x
x = COO.from_numpy(random_x((100, 100, 1000)))
y = COO.from_numpy(random_x((100, 100, 1000)))
%timeit x + y

New: 522 ms ± 3.72 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
Old: 1.69 s ± 22.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

sparse/tests/test_core.py
hameerabbasi: It was nice you suggested that we test the `nnz`. Apparently, it was failing for shapes `(5, 1)` and `(1, 5)`. Not sure why yet.

sparse/core.py
```diff
@@ -603,55 +611,346 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
+
+        data_list.append(func(self_data[matched_self],
+                              other_data[matched_other],
+                              *args, **kwargs))
+        coords_list.append(matched_coords)
+
+        self_func = func(self_data, 0, *args, **kwargs)
+
+        # Add unmatched parts as necessary.
+        if (self_func != 0).any():
```
hameerabbasi: Used this logic (same as below) to check if broadcasting "matches" are necessary. Got rid of the magic numbers. This should be solid.

sparse/core.py
```diff
@@ -603,55 +611,346 @@ def elemwise_binary(self, func, other, *args, **kwargs):
...
+        data = np.concatenate(data_list) if len(data_list) else np.empty((0,), dtype=self.dtype)
+        coords = np.concatenate(coords_list, axis=1) if len(coords_list) else \
+            np.empty((0, len(result_shape)), dtype=self.coords.dtype)
 
         nonzero = data != 0
         data = data[nonzero]
-        coords = coords[nonzero]
+        if len(data) != 0:
+            coords, idx = np.unique(coords[:, nonzero], axis=1, return_index=True)
+            data = data[idx]
```
hameerabbasi: Supporting the shapes like `(1, 5)` and `(5, 1)` which have no "common" dimensions is proving to be quite a pain. For some reason, without this check, there were duplicates in `coords` which should not have been summed for these shapes. I still have to figure out why. In the meantime, we have this workaround.

hameerabbasi: As a side comment, the `np.unique call takes up most of the time.`

sparse/core.py
hameerabbasi: I isolated the performance hit of `np.unique` to only where it's needed.
```
274 ms ± 4.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

sparse/core.py
```diff
@@ -396,16 +402,20 @@ def linear_loc(self, signed=False):
...
+    @staticmethod
+    def _linear_loc(coords, shape, signed=False):
+        n = reduce(operator.mul, shape, 1)
         if signed:
             n = -n
         dtype = np.min_scalar_type(n)
-        out = np.zeros(self.nnz, dtype=dtype)
-        tmp = np.zeros(self.nnz, dtype=dtype)
+        out = np.zeros(coords.shape[1], dtype=dtype)
+        tmp = np.zeros(coords.shape[1], dtype=dtype)
```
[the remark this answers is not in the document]
mrocklin: Ah, I see that coords is now an input

sparse/tests/test_core.py
```diff
@@ -381,6 +401,7 @@ def test_scalar_multiplication():
     assert_eq(x // 2.5, a // 2.5)
 
 
+@pytest.mark.filterwarnings('ignore:divide by zero')
```
mrocklin: That's useful.  I didn't know pytest provided this.

sparse/core.py
hameerabbasi: Thanks, it's a pretty useful tool, one that I hadn't seen before. I'm actually a newcomer to Python (only a few months of experience), having fallen in love with its power after leaving MATLAB.

The [PyCharm profiling tool](https://www.jetbrains.com/help/pycharm/optimizing-your-code-using-profilers.html) is pretty useful but it doesn't show stuff in the neat tree format like snakeviz does.

In any case, looks like the function is close to optimal. What should take time intuitively is what is taking time.

mrocklin: > Thanks, it's a pretty useful tool, one that I hadn't seen before. I'm actually a newcomer to Python (only a few months of experience), having fallen in love with its power after leaving MATLAB.

Welcome!  I moved over from MatLab myself a long while ago.

hameerabbasi: Latest commit doesn't change code time significantly.

mrocklin:
Thanks for this.  I'm excited about this work.  My TODO list today is
somewhat long, but I'll try to take a look.

sparse/tests/test_core.py
mrocklin: Did this get resolved?  At first glance it seems fine:

```python
In [26]: x = sparse.COO(random_x((5, 1)))

In [27]: x
Out[27]: <COO: shape=(5, 1), dtype=float64, nnz=3, sorted=True, duplicates=False>

In [28]: x.broadcast_to((5, 5))
Out[28]: <COO: shape=(5, 5), dtype=float64, nnz=15, sorted=True, duplicates=False>
```

Adding it explicitly to tests to avoid future regression might be useful

sparse/core.py
```diff
@@ -912,3 +1284,48 @@ def stack(arrays, axis=0):
 
     return COO(coords, data, shape=shape, has_duplicates=has_duplicates,
                sorted=(axis == 0) and all(a.sorted for a in arrays))
+
+
+# (c) Paul Panzer
+# Taken from https://stackoverflow.com/a/47833496/774273
+# License: https://creativecommons.org/licenses/by-sa/3.0/
```
mrocklin: There was a nice bit of conversation on this question.  Looked like a very productive use of Stack Overflow :)

mrocklin:
This looks great to me.  Thank you for all of your work here @hameerabbasi .  I'm good to merge.  I had a small question and will wait on that, but happy to merge regardless.  Thank you also for your patience with me on this.

sparse/tests/test_core.py
hameerabbasi: Yes, it was resolved. It wasn't `broadcast_to` that was failing, it was `elemwise_binary` where one element was `(1, 5)` and another was `(5, 1)`, or vice versa. I realized it was a wider bug only showing up with those specific shapes, and I made sure it won't occur. It was fixed in 63e68d93827b3fa3dd29975bab4949787617b056.

hameerabbasi:
I replied to your comment. Thanks for taking the time to review this. I'll hopefully continue adding features. This was one of the more difficult ones, so I started here.

sparse/tests/test_core.py
hameerabbasi: I don't think regression will be an issue because I replaced the logic where the bug was occurring entirely.

mrocklin:
OK.  Merging. Thanks @hameerabbasi !

stefanv:
The broadcasting seems to work correctly for all the basic cases I tested.  With some shenanigans, you can break indexing:

```
In [54]: ind0 = np.arange(12)

In [55]: ind0 = np.arange(12)[:, None, None]

In [56]: ind1 = np.arange(3)

In [57]: z = np.zeros((8, 15, 12, 3))

In [58]: p = sparse.core.COO.from_numpy(np.zeros((8, 15, 12, 3)))

In [59]: z[:, ..., ind0, ind1].shape
Out[59]: (8, 15, 12, 1, 3)

In [60]: p[:, ..., ind0, ind1].shape
ValueError                                Traceback (most recent call last)
<ipython-input-61-f3bd65589d07> in <module>()
----> 1 p[:, ..., ind0, ind1].shape

~/src/sparse/sparse/core.py in __getitem__(self, index)
    247         if not isinstance(index, tuple):
    248             index = (index,)
--> 249         if len(index) - index.count(None) - index.count(Ellipsis) > self.ndim:
    250             raise IndexError("too many indices for array")
    251         if index.count(Ellipsis) > 1:

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

hameerabbasi:
Not just that, I think we should also support `np.newaxis` along with `None`. It's the same, from what I understand.

mrocklin:
```python
In [1]: import numpy as np

In [2]: np.newaxis is None
Out[2]: True
```

hameerabbasi:
Ah, apologies. I didn't know they pointed to the same object.

PR #37, opened 2017-12-23
Add slicing support for ellipsis and None

mrocklin:
I ran into this while testing https://github.com/mrocklin/sparse/pull/35 .  

@hameerabbasi if you have a moment to look this over I would welcome your review


hameerabbasi:
Looks great to me. One minor point: A check to see there isn't more than one `Ellipsis` and a subsequent exception + test for that exception would be great. Either way, it doesn't block previous use cases, so should be good to merge.

mrocklin:
Added such a test.  It looks like Python handles this check for us.

mrocklin:
Ah, never mind.  My test was faulty here

sparse/core.py
```diff
@@ -233,10 +233,19 @@ def __sizeof__(self):
     def __getitem__(self, index):
         if not isinstance(index, tuple):
             index = (index,)
-        index = tuple(ind + self.shape[i] if isinstance(ind, numbers.Integral) and ind < 0 else ind
+        if len(index) - index.count(None) > self.ndim:
+            raise IndexError("too many indices for array")
+        if index.count(Ellipsis) > 1:
+            raise IndexError("an index can only have a single sllipsis ('...')")
```
hameerabbasi: There is a spelling error here. `sllipsis` should be `Ellipsis`.

mrocklin:
Thanks.  Fixed

sparse/core.py
```diff
@@ -233,10 +233,19 @@ def __sizeof__(self):
     def __getitem__(self, index):
         if not isinstance(index, tuple):
             index = (index,)
-        index = tuple(ind + self.shape[i] if isinstance(ind, numbers.Integral) and ind < 0 else ind
+        if len(index) - index.count(None) > self.ndim:
+            raise IndexError("too many indices for array")
```
hameerabbasi: This will fail on things like `a[1, ...]` where `a` is `(n,)`-shaped. There should be an additional factor of `- index.count(Ellipsis)` here.

hameerabbasi:
Sorry for requesting so much changes. Here's the last one. It's nitpicky, sure, but I want to make sure all our bases are covered.

mrocklin:
> Sorry for requesting so much changes. Here's the last one. It's nitpicky, sure, but I want to make sure all our bases are covered.

No, this is great.  Thank you for all of your work reviewing here.

sparse/core.py
```diff
@@ -283,7 +283,10 @@ def __getitem__(self, index):
             coords.append(self.coords[j][mask])
             shape.append(self.shape[j])
 
-        coords = np.stack(coords, axis=0)
+        if coords:
+            coords = np.stack(coords, axis=0)
+        else:
+            coords = np.array(coords)
```
hameerabbasi: I may be missing something here, but won't `np.array([])` create an array with a `dtype` of `float64` and a shape of `(0,)`? It would be much nicer to do this: `coords = np.empty((0, len(shape)), dtype=np.uint8)`

hameerabbasi: And also to specify a `dtype` even in the first case...

hameerabbasi:
Some minor comments that will hopefully address the CI failures.

PR #39, opened 2017-12-25
Fix scalar slicing issue.

hameerabbasi:
Basically, the test was failing because we weren't making `coords` and `data` 2-D and 1-D as expected by our code. This fixes those bugs with scalar slicing.


mrocklin:
Thank you for the fix @hameerabbasi !

Issue #33, opened 2017-09-28
Support for triu/tril


jakirkham:
Would be nice to have support for `triu` and `tril` functions to convert a dense Dask Array into a sparse one that only holds the upper or lower triangle respectively.

mrocklin:
Is there a nice implementation of this somewhere for sparse arrays?

jakirkham:
AFAICT SciPy just uses COO for this case. So I'm guessing this would work with what you already have here. These would just be convenience functions for getting that part of the array into COO form.

hameerabbasi:
Implementing `triu` and `tril` for `COO` input/output seems trivial, even without looking at SciPy code, and I may take a look at it sometime. However, doing it on Dask arrays without converting them to `COO` in between seems like an edge use-case. If the original Dask array was dense, this would only result in a density of around `0.5`, in which case it is not a good idea to be using a sparse array anyway.

hameerabbasi:
@jakirkham Okay to close this? We added `triu` and `tril` in #40.

jakirkham:
Thanks for doing this @hameerabbasi. Is it possible to pass a vanilla array (e.g. NumPy) to these functions?

hameerabbasi:
You can do something like `triu(COO.from_numpy(x))` for a dense array. But they don't directly support dense arrays.

jakirkham:
Ok, thanks.

hameerabbasi:
I would however say that if the original array is really dense (has a low number of nonzero elements), then using `COO` to store the upper/lower triangular parts is a bad idea since it will result in a sparsity of 0.5. I usually find `COO` is better in memory with a sparsity of 0.25 and better on most operations with a sparsity of 0.1.

jakirkham:
Yeah I had been debating using this for triangular arrays and in the end decided it was better to just store them in flat 1-D arrays without coordinates as the transformation to 2-D arrays could be easily determined.

PR #40, opened 2017-12-26
Add triu + tril + tests.

hameerabbasi:
Addresses #33. Thought this was trivially simple, so might as well do it immediately.


sparse/tests/test_core.py
```diff
@@ -600,3 +600,31 @@ def test_scalar_slicing():
...
+    ((3, 4, 5), 1),
+    ((4, 2), -1),
+    ((2, 4), -2),
+    ((4, 4), 1000),
+])
+def test_tril(shape, k):
+    x = random_x(shape)
+    s = COO.from_numpy(x)
+
+    assert_eq(np.tril(x, k), s.tril(k))
```
mrocklin: We might include each of these assert lines into a single `test_trilu` test?

sparse/core.py
```diff
@@ -1121,6 +1121,26 @@ def conjugate(self, out=None):
...
+
+        mask = self.coords[-2] + k <= self.coords[-1]
+
+        coords = self.coords[:, mask]
+        data = self.data[mask]
+
+        return COO(coords, data, self.shape, self.has_duplicates, self.sorted)
+
+    def tril(self, k=0):
+        assert self.ndim >= 2
```
mrocklin: We should probably change these to `ValueError`s with informative error messages.  A brief test using `pytest.raises` might also be useful.

mrocklin:
It looks like in NumPy these functions are raw functions, not methods.  We should probably keep to that convention.

sparse/core.py
hameerabbasi: If we follow Numpy strictly strictly, `ndim == 1` works. Not sure if we should support it here, since it is an artefact of the way it is implemented in Numpy rather than an actual feature. I haven't tested scalars (`ndim == 0`) yet. 

I want to mimic numpy behaviour except, of course artefacts. I guess your approach is good but I'd hold off on the test, simply because we're not mimicking Numpy.

sparse/tests/test_core.py
hameerabbasi: I guess I could do that, yes.

sparse/core.py
mrocklin: In that case then we should raise a `NotImplementedError`.  Using `assert` in user code is generally avoided.  For example when people run Python in optimized mode `assert` statements get passed over.  Generally the convention is to always `raise ValueError("informative message")` or `raise NotImplementedError("informative message")` instead of using `assert`.

sparse/core.py
```diff
@@ -1323,6 +1323,60 @@ def stack(arrays, axis=0):
...
+        x : COO
+            The input array.
+        k : int
+            The diagonal above which elements are set to zero.
+
+        Returns
+        -------
+        COO
+            The output lower-triangular matrix.
+        """
```
mrocklin: It looks like the docstring here was accidentally indented an extra four spaces

mrocklin:
Looking good.  Thanks for the changes.

mrocklin:
Looks good to me.  Thanks @hameerabbasi !

sparse/core.py
```diff
@@ -1323,6 +1323,60 @@ def stack(arrays, axis=0):
...
+
+    Returns
+    -------
+    COO
+        The output upper-triangular matrix.
+    """
+    if not x.ndim >= 2:
+        raise NotImplementedError('sparse.triu is not implemented for scalars or 1-D arrays.')
+
+    mask = x.coords[-2] + k <= x.coords[-1]
```
jakirkham: Interesting. Was not aware of this behavior in NumPy. Raised issue ( https://github.com/numpy/numpy/issues/10310 ) to see if we can get it documented in NumPy.

PR #41, opened 2017-12-26
Add random() function

nils-werner:
I think a `random()` function is could be pretty handy for generating some quick testing data.

I tried several methods, like generating `.coords` and `.data` myself, and creating a `COO` object from them. In the end it turned out to be the easiest to create a linear `scipy.sparse` array and simply reshape it to the desired output shape.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1531,7 +1531,6 @@ def random(shape, density=0.01, dtype=None):
 
     """
     elements = np.prod(shape)
-    nnz = int(elements * density)
 
     return COO.from_scipy_sparse(
         scipy.sparse.rand(elements, 1, density, dtype=dtype)
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1583,7 +1583,7 @@ def random(
 
     if random_state is None:
         random_state = np.random
-    elif isinstance(random_state, (int, np.integer)):
+    elif isinstance(random_state, numbers.Integral):
         random_state = np.random.RandomState(random_state)
     if data_rvs is None:
         data_rvs = random_state.rand
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1592,7 +1592,7 @@ def random(
     if elements < 3 * nnz:
         ind = random_state.choice(elements, size=nnz, replace=False)
     else:
-        ind = np.empty(nnz, dtype=np.int64)
+        ind = np.empty(nnz, dtype=np.min_scalar_type(elements - 1))
         selected = set()
         for i in range(nnz):
             j = random_state.randint(elements)
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1516,30 +1516,42 @@ def _grouped_reduce(x, groups, method, **kwargs):
     return result, inv_idx, counts
 
 
-def random(shape, density=0.01, dtype=None, random_state=None):
+def random(shape, density=0.01, dtype=None, canonical_order=False, random_state=None):
     """ Generate a random sparse multidimensional array
 
     Parameters
     ----------
     shape: :obj:`tuple` of :obj:`int`
         Shape of the array
-    density: :obj:`float`
+    density: :obj:`float`, optional
         Density of the generated array.
-    dtype : dtype
+    dtype : numpy.dtype, optional
         Type of the returned array values.
+    canonical_order : bool, optional
+        Whether or not to put the output :obj:`COO` object into canonical
+        order. :code:`False` by default.
     random_state : {numpy.random.RandomState, int}, optional
         Random number generator or random seed. If not given, the
         singleton numpy.random will be used. This random state will be used
         for sampling the sparsity structure, but not necessarily for sampling
         the values of the structurally nonzero entries of the matrix.
 
+    Returns
+    -------
+    COO
+        The generated random matrix.
     """
     elements = np.prod(shape)
 
-    return COO.from_scipy_sparse(
+    ar = COO.from_scipy_sparse(
         scipy.sparse.rand(
             elements, 1, density,
             dtype=dtype,
             random_state=random_state
         )
     ).reshape(shape)
+
+    if canonical_order:
+        ar.sum_duplicates()
+
+    return ar
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1516,7 +1516,7 @@ def _grouped_reduce(x, groups, method, **kwargs):
     return result, inv_idx, counts
 
 
-def random(shape, density=0.01, dtype=None):
+def random(shape, density=0.01, dtype=None, random_state=None):
     """ Generate a random sparse multidimensional array
 
     Parameters
@@ -1527,11 +1527,19 @@ def random(shape, density=0.01, dtype=None):
         Density of the generated array.
     dtype : dtype
         Type of the returned array values.
-
+    random_state : {numpy.random.RandomState, int}, optional
+        Random number generator or random seed. If not given, the
+        singleton numpy.random will be used. This random state will be used
+        for sampling the sparsity structure, but not necessarily for sampling
+        the values of the structurally nonzero entries of the matrix.
 
     """
     elements = np.prod(shape)
 
     return COO.from_scipy_sparse(
-        scipy.sparse.rand(elements, 1, density, dtype=dtype)
+        scipy.sparse.rand(
+            elements, 1, density,
+            dtype=dtype,
+            random_state=random_state
+        )
     ).reshape(shape)
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1514,3 +1514,25 @@ def _grouped_reduce(x, groups, method, **kwargs):
     result = method.reduceat(x, inv_idx, **kwargs)
     counts = np.diff(np.concatenate((inv_idx, [len(x)])))
     return result, inv_idx, counts
+
+
+def random(shape, density=0.01, dtype=None):
+    """ Generate a random sparse multidimensional array
+
+    Parameters
+    ----------
+    shape: :obj:`tuple` of :obj:`int`
+        Shape of the array
+    density: :obj:`float`
+        Density of the generated array.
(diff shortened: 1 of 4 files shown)


mrocklin:
Thanks for this.  Some notes:c

1.  This could use a test
2.  Any thoughts on using this to replace our current `random_x` function currently used in tests?  Your solution is likely far more efficient.

nils-werner:
> 1. This could use a test

Definitely! What would you like to be tested?

> 2. Any thoughts on using this to replace our current random_x function currently used in tests?

Might be worth investigating. I didn't understand what `random_x` was doing at first glance, so I just ignored it for now.

mrocklin:
Well, we might ensure a variety of things we know about sprase matrices.  Here are a few that come to mind:

1.  The shape matches the intended shape
2.  The dtype matches the intended dtype (for a few dtypes)
3.  The density matches the intended density (according to nnz)
4.  Two calls to the same function don't produce the same results
5.  Two calls with the same args and with the same random state seed (we'll need to pass this through) produce equivalent results

mrocklin:
> Might be worth investigating. I didn't understand what random_x was doing at first glance, so I just ignored it for now.

It attempts to do the same thing that you're doing here, just much less efficiently, and much less generally :)

hameerabbasi:
The proper way to test a random function is currently a hot debate topic. There seem to be two ways to test it, both which I'm not happy with:

- Mock/Monkeypatch the underlying random function to see it is getting the correct arguments. In this case, `scipy.sparse.rand`.
- Run it lots of times until it is statistically certain to be in a certain range, then test for that range.

I will also add that `random_x` returns a dense Numpy array, and this returns a sparse one.

mrocklin:
I don't think we need to test the random number generator within this.  We can rely on SciPy/NumPy to produce a decent distribution of numbers.  I'm mostly concerned that we test basic user expectations.

sparse/core.py
```diff
@@ -1380,3 +1380,24 @@ def _match_arrays(a, b):
...
+        Density of the generated array.
+    dtype : dtype
+        Type of the returned array values.
+
+
+    """
+    elements = np.prod(shape)
+
+    return COO.from_scipy_sparse(
+        scipy.sparse.rand(elements, 1, density, dtype=dtype)
```
hameerabbasi: Here, it would be useful to specify `format='coo'` explicitly. Might speed up the conversion.

hameerabbasi:
> (we'll need to pass this through)
It looks like the version of [`scipy.sparse.rand`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.rand.html) is called [`scipy.sparse.random`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.random.html#scipy.sparse.random) and only supports floating point dtypes. So we might lose something if we implement this check.

mrocklin:
> It looks like the version of scipy.sparse.rand that supports passing through a random state is called scipy.sparse.random and only supports floating point dtypes. So we might lose something if we implement this check.

That's unfortunate

sparse/core.py
nils-werner: `scipy.sparse.rand()` defaults to `format='coo'`...

nils-werner:
> It looks like the version of scipy.sparse.rand that supports passing through a random state is called scipy.sparse.random

Really? Looking at the reference it says

    scipy.sparse.rand(m, n, density=0.01, format='coo', dtype=None, random_state=None)

and

    scipy.sparse.random(m, n, density=0.01, format='coo', dtype=None, random_state=None, data_rvs=None)

(note the `random_state` kwarg).

hameerabbasi:
Ah, my mistake, the notes on `rand` were misleading.

> Similar function that allows a user-specified random data source.

nils-werner:
I have replaced `random_x` and `random_x_bool` with calls to `sparse.random` in all the tests. However, I stumbled across a few oddities that need to be discussed. See code review.

sparse/tests/test_core.py
```diff
@@ -9,28 +9,6 @@
 import sparse
```
hameerabbasi: Please remove `random` here, it's unneeded.

sparse/tests/test_core.py
```diff
@@ -9,28 +9,6 @@
...
-
-
-def random_x_bool(shape):
-    x = np.zeros(shape=shape, dtype=np.bool)
-    for i in range(max(5, np.prod(x.shape) // 10)):
-        x[tuple(random.randint(0, d - 1) for d in x.shape)] = True
-    return x
-
-
 @pytest.mark.parametrize('reduction,kwargs', [
```
hameerabbasi: Add an extra newline here.

hameerabbasi:
The issue you were having here earlier was already fixed with #56. I've added comments that should help fix the `flake8` errors.

hameerabbasi:
I'm good to merge as soon as `flake8` is fixed.

nils-werner:
Don't merge yet, the tests still randomly fail due to the input data

hameerabbasi:
You might want to rebase this on master and see if the tests still fail. If we're talking about the one I saw fail (`test_reductions`) then it was fixed in #56.

hameerabbasi:
Since it's just failing for the `np.float16` dtype, I'm pretty sure these are just floating point errors. `np.allclose` checks for a relative error of `1.0e-15`  ([see here](https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html)) and `float16` is nowhere near that accurate. You might want to pass an additional argument called `rtol` to `assert_eq` that defaults to `1.0e-15`, increase that in this case. I think this fails because we're moving from straight integers stored in a floating-point `dtype` to real floating-points... Which increases the likelihood of random errors.

nils-werner:
If I raise `atol` to `1e-2` I stop seeing random errors...

hameerabbasi:
Great! Fix `flake8` and add a test for `random` and we should be good to merge! I could do that too, if you like!

Thanks for your work on all this.

hameerabbasi:
`float16` has just `10` bits of floating point precision, out of which one is essentially a sign bit. That leaves `9` bits. `2 ** -9 = 1.953125e-3`. Account for the fact that the errors can accumulate, and I think we're good with `atol=1.0e-2`.

nils-werner:
While playing around with ways of comparing arrays I realized that `COO(scipy.sparse.random)` always returns unsorted indices. Do we want them to be never sorted/always sorted/optionally sorted?

hameerabbasi:
It would be nice to have an option `sorted` which defaults to `False`. I actually think `sorted=False` is better because we don't want our methods to assume sorted indices, we want them to work regardless of whether the input is or isn't sorted.

hameerabbasi:
During testing I found that the `dtype` argument in `scipy.random.sparse` has no effect. I was considering removing it.

mrocklin:
Long term we might consider copying scipy's logic to produce the coordinates but then create our own data.

hameerabbasi:
Do you think this is good to merge? It's good from my side. I took a look at SciPy's code, it's got a lot of unnecessary branches and I think we're better off just using it.

nils-werner:
Or we could use the coordinates from SciPy and then do

    self.data = numpy.random.rand(elements, dtype=dtype)

hameerabbasi:
[`numpy.random.rand`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.rand.html) doesn't seem to have a dtype parameter either.

nils-werner:
That's right... no Idea why I thought it had one...

nils-werner:
Maybe this is starting to overengineer the solution a bit, but we could allow for a `data_callback` kwarg, and users could inject their own random data generator into the function call:

```
sparse.random((2, 3, 5), density=0.5)  # float
sparse.random((2, 3, 5), density=0.5, data_callback=lambda x: np.random.choice([True, False], size=x))  # True and False
sparse.random((2, 3, 5), density=0.5, data_callback=lambda x: np.random.randint(10, 20, size=x)).todense()  # Integers from 10 to 20
sparse.random((2, 3, 5), density=0.5, data_callback=lambda x: np.full(x, True))  # all True
```

hameerabbasi:
No, I actually really like this solution. However, I think it should be in a separate method rather than in `sparse.random`. I think we should mimic `scipy.sparse.random` with `sparse.random`, and add another utility method `random_callback` (my brain is crapping out on a better name).

hameerabbasi:
Merging if there are no comments until 21:00 German time. This can hold tests on other branches back, and they may need to be rebased on this, so I don't want to hold this out too long.

nils-werner:
Actually, `scipy.sparse.random()` does exactly that, and calls the parameter `data_rvs`. However what they don't do is change the output array `dtype`.

hameerabbasi:
Hmmm... This seems rather inefficient and wasteful to generate the data twice. Maybe it's worth it to copy [this part](https://github.com/scipy/scipy/blob/v1.0.0/scipy/sparse/construct.py#L762-L773) of the Scipy code if we can guarantee consistency internally.

hameerabbasi:
@nils-werner Any more changes? If not, we should merge this.

nils-werner:
Slow down... we've just doubled the LOC in `random()` a minute ago, and nobody had the time to review the code yet...

hameerabbasi:
Good point. I just really don't want to put off changing the test code in the other branches.

sparse/utils.py
```diff
@@ -2,7 +2,7 @@
 from .core import COO
 
 
-def assert_eq(x, y):
+def assert_eq(x, y, rtol=1.e-5, atol=1.e-8):
```
mrocklin: Should we just pass through `**kwargs` here?

mrocklin:
Overall this seems great to me.  It looks like we're able to handle most cases and we're staying within the scipy.sparse API.

sparse/utils.py
hameerabbasi: Yes, that seems like a better option.

PR #42, opened 2017-12-26
More complete gitignore file

nils-werner:
I noticed that the current `.gitignore` file was pretty incomplete, not ignoring `*.egg` or `build` directories etc.

So I took the Gitignore file from [github/gitignore.git](https://github.com/github/gitignore/blob/master/Python.gitignore) and added it to this project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,60 @@
-*.pyc
+#####=== Python ===#####
+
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @nils-werner !

PR #45, opened 2017-12-27
Update MANIFEST.in

### LINKED ISSUE #44: requirements.txt is missing in PyPI
Hi,

I tried to install sparse to a new (conda) environment by `pip install sparse`, but it fails with the following error.
```
Writing /tmp/easy_install-mhkizn5d/sparse-0.1.1/setup.cfg
Running sparse-0.1.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-mhkizn5d/sparse-0.1.1/egg-dist-tmp-jpw672f1
error: [Errno 2] No such file or directory: 'requirements.txt'
```

The installation succeeds if the scipy and numpy are already installed in the environment.
I think we need to include `requirements.txt` in [MANIFEST.in](https://github.com/mrocklin/sparse/blob/master/MANIFEST.in).

mrocklin:
Fixes #44

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,11 @@
-include LICENSE.txt
 recursive-include sparse *.py
+recursive-include sparse *.html
+recursive-include docs *.rst
+
+include setup.py
+include README.rst
+include LICENSE.txt
+include MANIFEST.in
+include requirements.txt
+
+prune docs/_build


fujiisoup:
Maybe README.rst also?
It is referred in setup.py.

mrocklin:
It's already there :)

hameerabbasi:
I would review this but I'm not sure I'm the best person, I'm not good at Python packaging.

Issue #10, opened 2017-04-27
Automatic conversion to dense arrays


mrocklin:
Operations on sparse arrays sometimes produce dense arrays.  This type instability can cause some frustration downstream but may be optimal performance-wise.

In many occasions we actually inherit this behavior from scipy.sparse, which returns `numpy.matrix` objects in some cases.  Currently we also return dense `numpy.ndarray` objects when this happens and when the number of non-zeros is high.  I'm running into cases where I want to do this more and more, especially in parallel computing cases where I tensordot and add together many sparse arrays.  Switching to dense starts to make a lot of sense.

However this will likely cause some frustration downstream as users sometime receive sparse and and sometimes receive dense arrays based on their data at the moment.  Is this performance gain worth it?

jakevdp:
That's a tough one... I'd probably avoid returning sparse or dense conditionally from the same routine, as that's a recipe for hard-to-catch bugs in corner cases.

mrocklin:
Yeah, last week I would have come out strongly in favor of type stability.  As I've been running computations in practice my confidence has decreased somewhat here.  I'm still not entirely in favor of optional conversions, but if I'm unable to optimize down some sparse operations it may become an interesting option.

The saving grace of losing type stability is that this may be come less of an issue if the `__array_ufunc__` protocol emerges in Numpy 1.13.  The type of container may start to matter much less in numpy code in the future.

hameerabbasi:
One issue that I see in the code is that `operator.add` for `COO` is treated differently from other operators, like `operator.mul`, in the way that we do `maybe_densify` for `add` but not for other operators, and when trying to correct this, a unit test failed. I ran into this when I tried to implement `operators.gt`, etc., and wanted to unify behavior across operators and consistently use the same code. Since there is a unit test, I'm assuming someone has a use for it, and I didn't want to change it without asking. We should decide on a concrete strategy to follow:

1. We use `maybe_densify` consistently.
2. We always raise an error.
3. We always densify when needed.
4. We provide a global flag or option that can be used with `with` similar to other libraries.
5. We use `maybe_densify` consistently with global flags/options for `maybe_densify`.

Whatever we choose, we should bake it into `elemwise` and `elemwise_binary` (and possibly provide kwargs). Since `__array_ufunc__` is now here, it should be okay to lose type stability. I'm in favor of option 5, as it is the most general.

hameerabbasi:
Two other things that would be helpful to think about are:

[...]

PR #46, opened 2017-12-27
Remove auto densification and unify operator code.

hameerabbasi:
I've removed the auto-densification in `__add__` and `exp` and I've removed the tests for it. In addition, I've unified the code for all operators so that they only work when not densified. In addition, they work with scalars if the scalars don't densify them.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -651,6 +651,7 @@ class COO(object):
             function. Otherwise, it will be treated as a unary function.
         kwargs : dict, optional
             The kwargs to pass to the function.
+
         Returns
         -------
         COO
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -579,54 +579,51 @@ class COO(object):
 
         return self
 
-    def __add__(self, other):
-        if isinstance(other, numbers.Number) and other == 0:
-            return self
-        if not isinstance(other, COO):
-            return self.maybe_densify() + other
+    def _perform_op(self, other, op):
+        if isinstance(other, COO):
+            return self.elemwise_binary(op, other)
         else:
-            return self.elemwise_binary(operator.add, other)
+            return self.elemwise(op, other)
 
-    def __radd__(self, other):
-        return self + other
+    def __add__(self, other):
+        return self._perform_op(other, operator.add)
+
+    __radd__ = __add__
 
     def __neg__(self):
         return COO(self.coords, -self.data, self.shape, self.has_duplicates,
                    self.sorted)
 
     def __sub__(self, other):
-        return self + (-other)
+        return self._perform_op(other, operator.sub)
 
     def __rsub__(self, other):
-        return -self + other
+        return -(self - other)
 
     def __mul__(self, other):
-        if isinstance(other, COO):
-            return self.elemwise_binary(operator.mul, other)
-        else:
-            return self.elemwise(operator.mul, other)
+        return self._perform_op(other, operator.mul)
 
     __rmul__ = __mul__
 
     def __truediv__(self, other):
-        return self.elemwise(operator.truediv, other)
+        return self._perform_op(other, operator.truediv)
 
     def __floordiv__(self, other):
-        return self.elemwise(operator.floordiv, other)
+        return self._perform_op(other, operator.floordiv)
 
     __div__ = __truediv__
 
     def __pow__(self, other):
-        return self.elemwise(operator.pow, other)
+        return self._perform_op(other, operator.pow)
 
     def __and__(self, other):
-        return self.elemwise_binary(operator.and_, other)
+        return self._perform_op(other, operator.and_)
 
     def __xor__(self, other):
-        return self.elemwise_binary(operator.xor, other)
+        return self._perform_op(other, operator.xor)
 
     def __or__(self, other):
-        return self.elemwise_binary(operator.or_, other)
+        return self._perform_op(other, operator.or_)
 
     def elemwise(self, func, *args, **kwargs):
         check = kwargs.pop('check', True)
@@ -641,11 +638,14 @@ class COO(object):
                    sorted=self.sorted)
 
     def elemwise_binary(self, func, other, *args, **kwargs):
-        assert isinstance(other, COO)
+        if not isinstance(other, COO):
+            raise ValueError("Performing this operation would produce "
+                             "a dense result: %s" % str(func))
+
+        check = kwargs.pop('check', True)
         self_zero = _zero_of_dtype(self.dtype)
         other_zero = _zero_of_dtype(other.dtype)
-        check = kwargs.pop('check', True)
-        func_zero = _zero_of_dtype(func(self_zero, other_zero, * args, **kwargs).dtype)
+        func_zero = _zero_of_dtype(func(self_zero, other_zero, *args, **kwargs).dtype)
         if check and func(self_zero, other_zero, *args, **kwargs) != func_zero:
             raise ValueError("Performing this operation would produce "
                              "a dense result: %s" % str(func))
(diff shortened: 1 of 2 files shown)


sparse/core.py
```diff
@@ -641,11 +656,14 @@ def elemwise(self, func, *args, **kwargs):
                    sorted=self.sorted)
 
     def elemwise_binary(self, func, other, *args, **kwargs):
-        assert isinstance(other, COO)
+        if not isinstance(other, COO):
+            raise ValueError("Performing this operation would produce "
+                             "a dense result: %s" % str(func))
```
mrocklin: Hrm, this isn't always true though.  For example `x > 5` or `x + 0` (which comes up when using the `sum` function) are both common case situations and should both work efficiently.  Is there some way that we can test against zero here?

hameerabbasi: It's already being checked in `_perform_op`. That one handles operations with scalars pretty nicely, and is general. I've also added a test for it (`test_elemwise_scalar`)

hameerabbasi: Though I can move it in here and refactor if you like.

sparse/tests/test_core.py
```diff
@@ -197,6 +200,66 @@ def test_elemwise_binary(func, shape):
...
+    (operator.le, -3),
+    (operator.eq, 1)
+])
+def test_elemwise_scalar(func, scalar):
+    x = random_x((2, 3, 4))
+    y = scalar
+
+    xs = COO.from_numpy(x)
+
+    assert_eq(func(xs, y), func(x, y))
```
mrocklin: Can we also check here that the output is a `COO` object and that the number of non-zeros hasn't significantly increased?

sparse/core.py
```diff
@@ -1067,7 +1071,7 @@ def __abs__(self):
 
     def exp(self, out=None):
         assert out is None
-        return np.exp(self.maybe_densify())
+        return self.elemwise(np.exp)
```
mrocklin: Will this always `err` (given current code)

(not a concern, just asking a question)

sparse/core.py
```diff
@@ -641,11 +650,13 @@ def elemwise(self, func, *args, **kwargs):
                    sorted=self.sorted)
 
     def elemwise_binary(self, func, other, *args, **kwargs):
-        assert isinstance(other, COO)
+        if not isinstance(other, COO):
+            return self.elemwise(func, other, *args, **kwargs)
```
mrocklin: Thanks for this change.  I think that this is nicer, especially given that `elemwise_binary` is public API.  People will probably try to use it and your change here will, I suspect, include fewer surprises for them.

mrocklin: We might at some point reduce this down to a single public `elemwise` function that handles both cases and dispatches to `_elemwise_unary` and `_elemwise_binary` based on inputs.  (not necessarily now though if you'd like to wrap this up quickly).

hameerabbasi: No, properly > quickly. It saves headache down the line.

The problem I see with that is, `elemwise` already exists and does what you are describing as `_elemwise_unary`. I have no problem dispatching between the two, but my question is:

- Is it worthwhile changing the API? Changing what `elemwise` is and hiding `elemwise_binary`? Do we have someone for whom it'll break?
- I was also considering supporting `scipy.sparse.spmatrix`. It's a small change, and well worth it, I expect.

hameerabbasi: It will also make implementing `__array_ufunc__` much easier.

mrocklin: The new `elemwise` would, I think, cover the functionality of the old `elemwise`.  It would now *also* work if the args weren't scalars, but were other things as well.  

In general though this project is young enough and obscure enough that I think API breaks are fine.  Anyone using this project is an early adopter.

I agree that supporting `scipy.sparse.spmatrix` objects would be valuable.

sparse/core.py
hameerabbasi: Yes. But given what we talked about in #10 it's probably best to keep it.

sparse/tests/test_core.py
hameerabbasi: It won't increase, but it can decrease. Let me add that check.

sparse/core.py
```diff
@@ -589,70 +589,103 @@ def __neg__(self):
...
     def __eq__(self, other):
-        return self.elemwise_binary(operator.eq, other)
+        return self.elemwise(operator.eq, other)
 
     def __ne__(self, other):
-        return self.elemwise_binary(operator.ne, other)
+        return self.elemwise(operator.ne, other)
 
     def elemwise(self, func, *args, **kwargs):
+        """
```
hameerabbasi: I needed your opinion here. Should we keep it this way and assume `other=args[0]` or should we add an `other=None` argument for explicitly selecting between binary and unary functions?

hameerabbasi: The first is more compact but the second can help when, for example, the first input to your function itself is supposed to be `COO` or `spmatrix`.

mrocklin: Without thinking about this too much I'm inclined to leave this as `def elemwise(self, func, *args, **kwargs)`.  I would expect `len(args)` to determine between elemwise_unary and elemwise_binary as you've done below.

mrocklin: I'm not sure I understand the second case

hameerabbasi: Suppose you have an inherently unary function that accepts `COO` or `spmatrix` as its second positional argument, instead of it being an operand to a binary function. I agree, it's a corner use case, but it is more explicit.

hameerabbasi: In case you think this won't ever be needed (and I mostly agree), this should be good to merge, if you haven't got any other comments.

mrocklin: I wouldn't worry about this for now.  Short term I might encourage users to use kwrags for this.

hameerabbasi: Fair enough. I overlooked that kwargs can map onto normal args in Python!

sparse/core.py
```diff
@@ -620,32 +611,85 @@ def __pow__(self, other):
...
+        if len(args) == 0:
+            return self._elemwise_unary(func, *args, **kwargs)
+        else:
+            other = args[0]
+            if isinstance(other, COO):
+                return self._elemwise_binary(func, *args, **kwargs)
+            elif isinstance(other, scipy.sparse.spmatrix):
+                other = COO.from_scipy_sparse(other)
+                return self._elemwise_binary(func, other, *args[1:], **kwargs)
```
mrocklin: This branch could use a test

sparse/core.py
```diff
@@ -620,32 +611,85 @@ def __pow__(self, other):
...
+        ----------
+        func
+            The function to apply to one or two arguments.
+        args : tuple, optional
+            The extra arguments to pass to the function. If args[0] is a COO object
+            or a scipy.sparse.spmatrix, the function will be treated as a binary
+            function. Otherwise, it will be treated as a unary function.
+        kwargs : dict, optional
+            The kwargs to pass to the function.
+        Returns
```
mrocklin: I think we need a newline above this docstring title for clean rendering with sphinx

mrocklin:
This looks great to me.  Thanks @hameerabbasi !

I've added you as a collaborator to this fork.  You should now have the ability to merge PRs.  In general I tend to use "Squash and Merge" when merging PRs unless particular care has been taken around the commit history and the individual commits have significant value.  

Care to give this a try and merge your own PR here?

hameerabbasi:
Thanks so much for adding me as a collaborator! Would you prefer I still worked on my own fork in different branches or can I also work with branches here?

mrocklin:
> Thanks so much for adding me as a collaborator!

Heh, you're doing more work on this than anyone else :)

> Would you prefer I still worked on my own fork in different branches or can I also work with branches here?

I tend to prefer keeping development work on different branches when possible.  I plan to continue working from my fork if/when we move this to a more public org.

Issue #47, opened 2017-12-27
nnz not correct for scalars.


hameerabbasi:
The `x.nnz` should be `1` for for a nonzero scalar `x`, yet it is zero. MWE:

```
import numpy as np
x = 1 + np.random.rand(5, 5)
import sparse
xs = sparse.COO.from_numpy(x)
xs[1, 1]
Out[6]: <COO: shape=(), dtype=float64, nnz=0, sorted=True, duplicates=False>
```

hameerabbasi:
Fixed with #48.

PR #48, opened 2017-12-27
Fix nnz for scalars.

hameerabbasi:
Addresses #47.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -299,7 +299,7 @@ class COO(object):
         if coords:
             coords = np.stack(coords, axis=0)
         else:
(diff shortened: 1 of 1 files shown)


mrocklin:
Is there a test that should have failed previously to this change?  Can we add such a test to prevent regressions in the future?

hameerabbasi:
I am hesitant to add a test (although I can add one) because indexing scalars in the future may give us scalars rather than `COO` objects. That's the way numpy/scipy.sparse do it, and that's how I believe it should be done. However, `x[1, 1, ...]` should give us a `COO` object for 2-D `x`.

mrocklin:
OK.  +1 from me then

PR #49, opened 2017-12-28
Add __array_ufunc__ for __call__ and reduce

hameerabbasi:
I've implemented `__array_ufunc__` for the `__call__` and `reduce` methods.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -325,6 +325,12 @@ class COO(object):
         return self.reduce(method, **kwargs)
 
     def reduce(self, method, axis=None, keepdims=False, **kwargs):
+        zero_reduce_result = method.reduce([_zero_of_dtype(self.dtype)], **kwargs)
+
+        if zero_reduce_result != _zero_of_dtype(np.dtype(zero_reduce_result)):
+            raise ValueError("Performing this reduction operation would produce "
+                             "a dense result: %s" % str(method))
+        
         # Needed for more esoteric reductions like product.
         self.sum_duplicates()
 
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -314,7 +314,20 @@ class COO(object):
 
     __repr__ = __str__
 
-    def reduction(self, method, axis=None, keepdims=False, dtype=None):
+    @staticmethod
+    def _reduce(method, *args, **kwargs):
+        assert len(args) == 1
+
+        self = args[0]
+        if isinstance(self, scipy.sparse.spmatrix):
+            self = COO.from_scipy_sparse(self)
+
+        return self.reduce(method, **kwargs)
+
+    def reduce(self, method, axis=None, keepdims=False, dtype=None):
+        # Needed for more esoteric reductions like product.
+        self.sum_duplicates()
+
         if axis is None:
             axis = tuple(range(self.ndim))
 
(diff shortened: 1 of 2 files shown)


sparse/core.py
```diff
@@ -314,40 +314,54 @@ def __str__(self):
...
+
+            flag = np.concatenate(([True], a.coords[0, 1:] != a.coords[0, :-1]))
+            # Partial credit to @shoyer
+            # Ref: https://gist.github.com/shoyer/f538ac78ae904c936844
+            inv_idx, = flag.nonzero()
+            result = method.reduceat(a.data, inv_idx, **kwargs)
+            counts = np.diff(np.concatenate(np.nonzero(flag) + ([a.nnz],)))
+            missing_counts = counts != a.shape[1]
+            result[missing_counts] = method(result[missing_counts],
+                                            _zero_of_dtype(result.dtype))
```
hameerabbasi: This isn't 100% accurate, given that reducing with zeros can potentially change the reduction, but it's a good trade-off and works with most reductions including `max`, `min`, `sum`, `prod`, etc.

sparse/tests/test_core.py
```diff
@@ -149,7 +162,8 @@ def test_dot():
         assert_eq(eval("sa @ sb"), sparse.dot(sa, sb))
 
         # Test that SOO's and np.array's combine correctly
-        assert_eq(eval("a @ sb"), eval("sa @ b"))
+        # Not possible due to https://github.com/numpy/numpy/issues/9028
+        # assert_eq(eval("a @ sb"), eval("sa @ b"))
```
hameerabbasi: I had to get rid of this check because `dense @ sparse` isn't possible.

sparse/core.py
```diff
@@ -314,40 +314,58 @@ def __str__(self):
...
+            self = COO.from_scipy_sparse(self)
+
+        return self.reduce(method, **kwargs)
+
+    def reduce(self, method, axis=None, keepdims=False, **kwargs):
+        zero_reduce_result = method.reduce([_zero_of_dtype(self.dtype)], **kwargs)
+
+        if zero_reduce_result != _zero_of_dtype(np.dtype(zero_reduce_result)):
+            raise ValueError("Performing this reduction operation would produce "
+                             "a dense result: %s" % str(method))
```
mrocklin: It seems to me that we might be ok with producing dense results with reductions, especially if we're losing most of our axes anyway.

sparse/core.py
```diff
@@ -314,40 +314,58 @@ def __str__(self):
...
+            # Ref: https://gist.github.com/shoyer/f538ac78ae904c936844
+            inv_idx, = flag.nonzero()
+            result = method.reduceat(a.data, inv_idx, **kwargs)
+            counts = np.diff(np.concatenate(np.nonzero(flag) + ([a.nnz],)))
+            missing_counts = counts != a.shape[1]
+            result[missing_counts] = method(result[missing_counts],
+                                            _zero_of_dtype(result.dtype), **kwargs)
+
+            a = COO(np.asarray([a.coords[0, inv_idx]]), result, shape=(np.prod(neg_axis),),
+                    has_duplicates=False, sorted=True)
```
mrocklin: cc @shoyer for review of this logic if he has time

sparse/core.py
```diff
@@ -356,21 +374,20 @@ def reduction(self, method, axis=None, keepdims=False, dtype=None):
...
+        assert out is None
+        return self.reduce(np.maximum, axis=axis, keepdims=keepdims)
+
+    def min(self, axis=None, keepdims=False, out=None):
+        assert out is None
+        return self.reduce(np.minimum, axis=axis, keepdims=keepdims)
+
+    def prod(self, axis=None, keepdims=False, dtype=None, out=None):
+        assert out is None
+        return self.reduce(np.multiply, axis=axis, keepdims=keepdims, dtype=dtype)
```
mrocklin: I'm curious, have you benchmarked the difference in performance here?

sparse/core.py
hameerabbasi: I think a `check` kwarg that defaults to `False` might be useful for later user configurability. This check is meant to be for a different reason, though, it checks that if we reduce across an all-zero axis, it shouldn't produce nonzeros. This might be useful in extremely sparse situations. If the overall result is dense because most axes are lost, this logic won't error out.

mrocklin: Ah ok, that makes sense.  Thank you for the explanation.

sparse/core.py
hameerabbasi: Good point, I  just thought of that.
```
import numpy as np, sparse
from sparse.tests.test_core import random_x
x = random_x((10 ** 3, 10 ** 4))
xs = sparse.COO.from_numpy(x)
%timeit xs.sum(axis=1)
```

```
New: 1.31 ms ± 231 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
Old: 11.8 ms ± 39.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I certainly didn't expect that! I expected it to be marginally faster, not this much. With `axis=0`, though:
```
New: 83 ms ± 645 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
Old: 9.8 ms ± 116 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

That's sad. But until `scipy.sparse` implements `__array_ufunc__`, we're stuck with this.

sparse/core.py
shoyer: I would suggest breaking out the grouped sun logic into a helper function if possible.

hameerabbasi: It's mostly the same logic used in `np.unique` combined with `ufunc.reduceat` (which does grouped reduce, exactly what we want). In the end I apply a final operation of a "final" reduction with zero iff the counts don't "fill" the axis.

hameerabbasi: Good point! It will come in useful when implementing `reduceat`.

hameerabbasi:
If the lost performance isn't a big issue, feel free to merge.

hameerabbasi:
@mrocklin, is this good to merge or is the performance hit a big issue?

sparse/core.py
hameerabbasi: @shoyer Broke out the reduce into `_grouped reduce` as per your request.

sparse/core.py
mrocklin: > Edit: Most of the time seems to be taken in sort_indices in the second case, as expected.

Yeah, I was just looking at this as well.  It looks like it remains slow on repeated computation as well.  I suspect that sortedness does not transfer through the transpose operation.  I wonder if there is an efficient way to sort more efficiently in this case.  (this isn't important to do now, I'm just thinking out loud.)

mrocklin:
> is this good to merge or is the performance hit a big issue?

I'm not currently very concerned about the performance issue.  I think that the benefits to organization and increased ability outweigh the costs here.

sparse/core.py
hameerabbasi: Of course, integer sorting can be made `O(n)`, while still keeping it stable. But implementing it will require Cython/Numba, at the very least, as Numpy doesn't have it.

hameerabbasi: See, for example, [this](https://stackoverflow.com/questions/35317442/how-to-sort-an-array-of-integers-faster-than-quicksort)

mrocklin: xref https://github.com/numpy/numpy/issues/6050

hameerabbasi: We can also cache transpose operations. However, it doesn't make sense that the same sum will be done twice, so...

hameerabbasi: Another thing we can do (which is a completely huge project on its own) is to switch to KD trees - they're always sorted, we never have to sort. `__getitem__` will also be faster. However, the cost is slower indexing.

Issue #32, opened 2017-09-01
Problems converting dask chunked sparse arrays greater size than 256x256


ljdursi:
Presumably related to https://github.com/dask/dask/issues/2586 : Computing a sparse array output from a chunked dask sparse array fails if the resulting sparse array is larger than 256x256 - coord type is uint16, but it looks like maybe intermediate calculations overflow:

```python
import dask.array as da
import sparse


def print_sparse_nonidentity(m, name):
    valid = True
    for i, j, v in zip(m.coords[0], m.coords[1], m.data):
        if not i == j:
            print name, i, j, v
            valid = False
    if valid:
        print "OK"


def identity_da(size, chunksize):
    p = {(i, i): 1. for i in range(size)}

    # single chunk
    a = da.from_array(sparse.COO(p), chunks=(chunksize, chunksize), asarray=False)
    return a


print "Should be identity matrix"
size = 256
print "size = ", size
a = identity_da(256, size//2)
c = a.compute()
print c
print_sparse_nonidentity(c, "256 chunked")

size = 258
print "size = ", size, " unchunked"
a = identity_da(size, size)
c = a.compute()
print c
print_sparse_nonidentity(c, "258 unchunked")

print "size = ", size, " chunked"
a = identity_da(size, size//2)
c = a.compute()
print c
print_sparse_nonidentity(c, "258 chunked")
```

Running this works for 256x256 matrix chunked, or 258x258 with only one chunk, but 258x258 broken into 4 chunks fails:

```
$ python fail-conversion.py
Should be identity matrix
size =  256
<COO: shape=(256, 256), dtype=float64, nnz=256, sorted=False, duplicates=False>
OK
size =  258  unchunked
<COO: shape=(258, 258), dtype=float64, nnz=258, sorted=False, duplicates=False>
OK
size =  258  chunked
<COO: shape=(258, 258), dtype=float64, nnz=258, sorted=False, duplicates=False>
258 chunked 256 0 1.0
258 chunked 257 1 1.0
```

Versions are as follow:
```
$ python
Python 2.7.13 (default, Dec 18 2016, 07:03:39)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import dask
>>> print dask.__version__
0.15.2
>>> import sparse
>>> print sparse.__version__
0.1.1
```

hameerabbasi:
[...]

PR #51, opened 2017-12-28
Fix large concatenations and stacks.

hameerabbasi:
Addresses #32.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1301,10 +1301,15 @@ def concatenate(arrays, axis=0):
     assert all(x.shape[ax] == arrays[0].shape[ax]
                for x in arrays
                for ax in set(range(arrays[0].ndim)) - {axis})
-    data = np.concatenate([x.data for x in arrays])
-    coords = np.concatenate([x.coords for x in arrays], axis=1)
-
     nnz = 0
+    dim = sum(x.shape[axis] for x in arrays)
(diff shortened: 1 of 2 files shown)


sparse/core.py
```diff
@@ -1301,19 +1301,22 @@ def concatenate(arrays, axis=0):
...
                for ax in set(range(arrays[0].ndim)) - {axis})
+    nnz = 0
+    dim = sum(x.shape[axis] for x in arrays)
+    shape = list(arrays[0].shape)
+    shape[axis] = dim
+
+    coords_dtype = np.min_scalar_type(max(shape) - 1) if len(shape) != 0 else np.uint8
     data = np.concatenate([x.data for x in arrays])
-    coords = np.concatenate([x.coords for x in arrays], axis=1)
+    coords = np.concatenate([x.coords for x in arrays], axis=1).astype(coords_dtype)
```
mrocklin: Is it better to call `.astype(...)` before concatenation on all of `arrays` or after as is done here?

hameerabbasi: I would think after. Because it would be applied to a bigger array and thus the overhead would be small. Memory would be better, too, as we are just passing already-allocated arrays to `concatenate`, otherwise we'd have to allocate for each temp array.

Incidentally, `astype` just returns the original array if it is already the correct dtype.

PR #55, opened 2017-12-30
Copy constructor

nils-werner:
This is inspired by NumPy's behaviour, where people tend to do `numpy.array(a)` on any iterable `a`, including `ndarray`s, just to make sure it is an `ndarray`.

Do you think something like this would make sense? Would save us having to do type checks in places where we just want to make sure it is a `COO` instance.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -111,6 +111,15 @@ class COO(object):
                 coords = list(coords.items())
                 has_duplicates = False
 
+            if isinstance(coords, COO):
+                self.coords = coords.coords
+                self.data = coords.data
+                self.has_duplicates = coords.has_duplicates
+                self.sorted = coords.sorted
+                self.shape = coords.shape
+                return
+
+
             if isinstance(coords, np.ndarray):
                 result = COO.from_numpy(coords)
                 self.coords = result.coords
@@ -1336,7 +1345,7 @@ def _mask(coords, idx):
 
 
 def concatenate(arrays, axis=0):
-    arrays = [x if type(x) is COO else COO(x) for x in arrays]
+    arrays = [COO(x) for x in arrays]
     if axis < 0:
         axis = axis + arrays[0].ndim
     assert all(x.shape[ax] == arrays[0].shape[ax]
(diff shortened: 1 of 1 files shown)


sparse/core.py
```diff
@@ -111,6 +111,15 @@ def __init__(self, coords, data=None, shape=None, has_duplicates=True,
...
                 has_duplicates = False
 
+            if isinstance(coords, COO):
+                self.coords = coords.coords
+                self.data = coords.data
+                self.has_duplicates = coords.has_duplicates
+                self.sorted = coords.sorted
+                self.shape = coords.shape
+                return
+
```
hameerabbasi: `flake8` is crapping out here. You can have a maximum of one blank line inside of a function.

sparse/core.py
```diff
@@ -1366,7 +1375,7 @@ def concatenate(arrays, axis=0):
 
 def stack(arrays, axis=0):
     assert len(set(x.shape for x in arrays)) == 1
-    arrays = [x if type(x) is COO else COO(x) for x in arrays]
+    arrays = [COO(x) for x in arrays]
```
hameerabbasi: I would leave these as-is, or maybe change them to `arrays = [x if isinstance(x, COO) else COO(x) for x in arrays]`. Generally, we want to avoid making copies where possible.

I realize what you are doing is a shallow copy, not a deep one, but in the future, if we allow manipulation of arrays (if we make the `COO` type mutable), we might need to change this to a deep copy, and that might be expensive.

hameerabbasi:
Some minor comments.

hameerabbasi:
Looks good to me. Merging if there are no further comments or changes until 21:00 German time.

hameerabbasi:
Actually, it'd be nice to have a test for equality. Something like generating a random matrix and then `COO(x) `, asserting they're not the same object, and an `assert_eq`.

mrocklin:
Do we also want to ensure that the data and coords arrays are not the same?

hameerabbasi:
I don't know the convention in Python here. I know Numpy makes a deep copy but we might get away with a shallow one since we're not mutating the original `data` and `coords` in any way. I'm good with this until it raises issues (which I suspect won't be for a long time since `COO` is immutable).

mrocklin:
OK, sounds good to me

PR #57, opened 2017-12-31
Add lots of indexing features.

hameerabbasi:
Adds support for:

- Returning scalars when slicing, e.g. `x[1, 1]` if `x` is 2-D is now scalar.
- Returning arrays when the last index is an `Ellipsis`, e.g. `x[1, 1, ...]` if `x` is 2-D is now a `()`-shaped `COO` object.
- Slices with steps other than `None` and `1` (even negative steps).
- Inbdexing with custom `dtype`s. So if `x` has a custom `dtype`, `x['field']` is now supported.
- Now throws `IndexError`s consistent with Numpy for string and out of range indices.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -250,13 +250,8 @@ class COO(object):
             if isinstance(index, str):
                 data = self.data[index]
                 idx = np.where(data)
-                coords = []
-
-                for i in range(self.ndim):
-                    coords.append(self.coords[i, idx[0]])
-
-                for i in range(1, np.ndim(data)):
-                    coords.append(idx[i])
+                coords = list(self.coords[:, idx[0]])
+                coords.extend(idx[1:])
 
                 return COO(coords, data[idx].flatten(),
                            shape=self.shape + self.data.dtype[index].shape,
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -311,7 +311,7 @@ class COO(object):
                         start = stop
                     shape.append((start - stop - step - 1) // (-step))
 
-                dt = np.min_scalar_type(min(-(l - 1) if l != 0 else -1 for l in shape))
+                dt = np.min_scalar_type(min(-(dim - 1) if dim != 0 else -1 for dim in shape))
                 coords.append((self.coords[i, mask].astype(dt) - start) // step)
                 i += 1
             elif isinstance(ind, list):
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -8,6 +8,8 @@ import operator
 import numpy as np
 import scipy.sparse
 
+from .slicing import normalize_index
+
 # zip_longest with Python 2/3 compat
 try:
     from itertools import zip_longest
@@ -263,25 +265,15 @@ class COO(object):
             else:
                 index = (index,)
 
-        last_ellipsis = False
-
-        if len(index) - index.count(None) - index.count(Ellipsis) > self.ndim:
-            raise IndexError("too many indices for array")
-        if index.count(Ellipsis) > 1:
-            raise IndexError("an index can only have a single ellipsis ('...')")
-        if any(ind is Ellipsis for ind in index):
-            loc = index.index(Ellipsis)
-            n = self.ndim - (len(index) - 1 - index.count(None))
-            if loc == len(index) - 1:
-                last_ellipsis = True
-            index = index[:loc] + (slice(None),) * n + index[loc + 1:]
-        if len(index) != 0 and all(ind == slice(None) for ind in index):
+        last_ellipsis = len(index) > 0 and index[-1] is Ellipsis
+        index = normalize_index(index, self.shape)
+        if len(index) != 0 and all(not isinstance(ind, Iterable) and ind == slice(None) for ind in index):
             return self
         mask = np.ones(self.nnz, dtype=np.bool)
         for i, ind in enumerate([i for i in index if i is not None]):
-            if ind == slice(None):
+            if not isinstance(ind, Iterable) and ind == slice(None):
                 continue
-            mask &= _mask(self.coords[i], ind, self.shape[i], i)
+            mask &= _mask(self.coords[i], ind, self.shape[i])
 
         n = mask.sum()
         coords = []
@@ -292,7 +284,6 @@ class COO(object):
                 i += 1
                 continue
             elif isinstance(ind, slice):
-
                 step = ind.step if ind.step is not None else 1
                 if step > 0:
                     start = ind.start if ind.start is not None else 0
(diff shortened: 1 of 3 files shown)


hameerabbasi:
@nils-werner, your input here would be valuable.

hameerabbasi:
cc @mrocklin Since you wrote the original `__getindex__`, it'd be helpful if you took a look at this.

mrocklin:
Thank you for doing this.  I look forward to reviewing.  Just as a warning,
I'll probably be busy for the next 24 hours or so.

sparse/core.py
```diff
@@ -245,26 +245,43 @@ def __sizeof__(self):
...
+            if isinstance(index, str):
+                data = self.data[index]
+                idx = np.where(data)
+                coords = []
+
+                for i in range(self.ndim):
+                    coords.append(self.coords[i, idx[0]])
+
+                for i in range(1, np.ndim(data)):
+                    coords.append(idx[i])
```
mrocklin: I don't understand the purpose of this.  Can you help explain?

sparse/core.py
```diff
@@ -245,26 +245,43 @@ def __sizeof__(self):
...
     def __getitem__(self, index):
         if not isinstance(index, tuple):
-            index = (index,)
+            if isinstance(index, str):
+                data = self.data[index]
+                idx = np.where(data)
+                coords = []
+
+                for i in range(self.ndim):
+                    coords.append(self.coords[i, idx[0]])
```
mrocklin: This is just `self.coords[:, idx[0]]`?

sparse/core.py
```diff
@@ -245,26 +245,43 @@ def __sizeof__(self):
...
             n = self.ndim - (len(index) - 1 - index.count(None))
-            index = index[:loc] + (slice(None, None),) * n + index[loc + 1:]
-        if all(ind == slice(None) for ind in index):
+            if loc == len(index) - 1:
+                last_ellipsis = True
+            index = index[:loc] + (slice(None),) * n + index[loc + 1:]
+        if len(index) != 0 and all(ind == slice(None) for ind in index):
             return self
-        mask = np.ones(self.nnz, dtype=bool)
+        mask = np.ones(self.nnz, dtype=np.bool)
```
mrocklin: This change doesn't do anything

```python
In [5]: np.bool is bool
Out[5]: True

In [6]: np.bool_
Out[6]: numpy.bool_
```

sparse/core.py
```diff
@@ -275,10 +292,27 @@ def __getitem__(self, index):
...
             elif isinstance(ind, slice):
-                start = ind.start or 0
-                stop = ind.stop if ind.stop is not None else self.shape[i]
-                shape.append(min(stop, self.shape[i]) - start)
-                coords.append(self.coords[i][mask] - start)
+
+                step = ind.step if ind.step is not None else 1
+                if step > 0:
+                    start = ind.start if ind.start is not None else 0
+                    start = max(start, 0)
```
mrocklin: What if someone does something like `x[-10:]` ?

mrocklin: It may be useful to separate out a `normalize_slice` function that reduces slice objects to a canonical form.  It looks like there might be some useful functions in `dask/array/slicing.py` that might be helpful here.

mrocklin: The `normalize_*` and `check_index` functions there might be helpful in particular

sparse/core.py
```diff
@@ -245,26 +245,43 @@ def __sizeof__(self):
...
+                idx = np.where(data)
+                coords = []
+
+                for i in range(self.ndim):
+                    coords.append(self.coords[i, idx[0]])
+
+                for i in range(1, np.ndim(data)):
+                    coords.append(idx[i])
+
+                return COO(coords, data.flatten(),
```
mrocklin: The `data.flatten()` call seems odd to me here.  This is for struct dtypes?

sparse/core.py
hameerabbasi: Custom `dtype`s can have multiple dimensions inside a single scalar 'field'. For example, `np.dtype([('grades', np.float64, (2, 2))])` will have a `(2, 2)` element "inside" each "scalar" element. `self.data['grades']` gets a `(nnz, 2, 2)` array in this case, but the resulting array has to be `self.shape + (2, 2)`. We first use `np.where` to get the nonzero indices. Then we append the matching parts of the original coordinates above, we get those from `idx[0]`. After this, we get the parts that come from the dimensions of the field itself, and append those to `coords`.

Finally, we flatten the data since we're done calculating all the `coords` of the resulting array.

sparse/core.py
hameerabbasi: You're right, it should be `data[idx].flatten()`. We have to filter out nonzero entries.

sparse/core.py
hameerabbasi: My intent was to make it more performant but looks like Numpy treats them the same...

sparse/core.py
hameerabbasi: Is there anything similar for `bool` indexing?

sparse/core.py
mrocklin: Right, I see.  I had expected the dtype to continue being of shape `(2, 2)` but I see that NumPy rolls this into the normal array's shape

```python

In [1]: import numpy as np

In [2]: np.ones((5,), dtype=np.dtype([('grades', np.float64, (2, 2))]))
Out[2]: 
array([([[ 1.,  1.], [ 1.,  1.]],), ([[ 1.,  1.], [ 1.,  1.]],),
       ([[ 1.,  1.], [ 1.,  1.]],), ([[ 1.,  1.], [ 1.,  1.]],),
       ([[ 1.,  1.], [ 1.,  1.]],)],
      dtype=[('grades', '<f8', (2, 2))])

In [3]: x = np.ones((5,), dtype=np.dtype([('grades', np.float64, (2, 2))]))

In [4]: x.shape
Out[4]: (5,)

In [5]: x['grades'].shape
Out[5]: (5, 2, 2)

In [6]: x['grades'].dtype
Out[6]: dtype('float64')

```

sparse/core.py
mrocklin: Not just Numpy, these are exactly the same objects, the `is` operator is object identity testing in Pyhton.

sparse/core.py
mrocklin: On a brief look I'm not seeing much.

sparse/core.py
hameerabbasi: I meant `np.bool_` and `bool`. If you create arrays with both they end up being the same `dtype`.

hameerabbasi:
Whoops. Looks like some old build results need to be cleaned.

hameerabbasi:
It turns out boolean slices were already handled in `slicing.py`, but the exceptions weren't, so I added those.

hameerabbasi:
Merging at 20:00 German time if there are no other proposed changes.

PR #58, opened 2017-12-31
Out-of-place sort_indices and sum_duplicates

nils-werner:
As far as I can tell `COO.sort_indices()` and `COO.sum_duplicates()` are the only two functions that specifically modify the array inplace (and in some cases also needlessly `return self`).

However, specifically for `sort_indices()` an out-of-place counterpart would be very interesting, e.g. when comparing two `COO` arrays: Temporarily sort both, then compare `.coords` and `.data` to see if they are equal.

Having both an inplace and an out-of-place function for the same operation is like Python offering `sorted(mylist)` and `mylist.sort()`.

What do you think?

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -4,6 +4,7 @@ from collections import Iterable, defaultdict, deque
 from functools import reduce
 import numbers
 import operator
+import copy
 
 import numpy as np
 import scipy.sparse
@@ -553,6 +554,8 @@ class COO(object):
         return csc
 
     def sort_indices(self):
+        """ Sort indices inplace
+        """
         if self.sorted:
             return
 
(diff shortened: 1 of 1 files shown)


hameerabbasi:
I, personally, find this to be an edge use-case. Usually, calling `sum_duplicates()` puts the data into canonical form. I find being in canonical form is usually better and more beneficial, and preserving the non-canonical version doesn't have much use.

It would help to know a use-case where this might be needed.

nils-werner:
While messing around with `assert_eq` I realized that it is actually never comparing the sparse data structures, and essentially always boils down to

    def assert_eq(x, y):
        assert np.allclose(xx.todense(), yy.todense())

while comparing the two data structures in sparse form should be much more efficient

    def sparse_assert_eq(x, y):
        x = x.sorted_indices()
        y = y.sorted_indices()
        assert np.array_equal(x.coords, y.coords) and np.allclose(x.data, y.data)

If instead inside this function I now call `x.sort_indices()` I will mess around with the object that was passed to me, introducing some potentially unexpected behaviour.

hameerabbasi:
Well, I argue that sorting indices is a rather desirable operation. In many cases, having unsorted indices introduces an overhead. For a lot of our functions, the majority of the performance penalty is sorting indices and if a method sorts them as a side-effect, we should document that side effect instead of not-in-place sorts, which would essentially translate to requiring multiple sorts later on for the same array.

hameerabbasi:
cc @mrocklin If he can give his input on this.

mrocklin:
I agree with @hameerabbasi that I don't see myself using this personally, but it does seem reasonable to add if others find it useful.

*If* we were to build this then we might do the following:

1. Bulid a copy method
2. add an `inplace=` keyword to these operations that calls copy and then returns that operation applied to a copied object instead

This would be out of alignment with the Numpy API but in line with APIs like Pandas.

hameerabbasi:
I agree with @mrocklin's design here. However, if we add `inplace`, we would need to return the modified object every time to maintain consistency.

nils-werner:
OK, I'm convinced.

PR #59, opened 2018-01-01
flake8 and coverage in pytest

nils-werner:
I totally missed that you guys were using `flake8` to keep the code style clean... To make it easier for first time contributors we could, instead of running `flake8` manually, integrate it into the `pytest` run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,10 @@ install:
   - conda create -n test-sparse python=$TRAVIS_PYTHON_VERSION pytest numpy scipy flake8 nomkl
   - source activate test-sparse
 
-  - python setup.py install
+  - pip install -e .[tests]
 
 script:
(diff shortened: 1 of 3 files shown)


hameerabbasi:
I really like having coverage like this. Merging.

hameerabbasi:
This caused me to run into pytest-dev/pytest#3077, and also [this issue with PyCharm/pydev](https://youtrack.jetbrains.com/issue/PY-27829).

nils-werner:
A workaround is to use

    py.test -k test_two

instead of 

    py.test pytest_test/tests/test_pytest.py::test_two

hameerabbasi:
@nils-werner I would if I programmed on the command line. Unfortunately, I use PyCharm to run and debug tests and it selects them using the second method.

PR #61, opened 2018-01-03
Validate .transpose() axes

nils-werner:
Before this change `COO` arrays allowed many variants of illegal `.transpose()` parameters:

    sparse.random((2, 3, 4)).transpose((0, 0, 1))  # duplicate axis
    sparse.random((2, 3, 4)).transpose((0, 1))  # too few
    sparse.random((2, 3, 4)).transpose((0, 1, 2, 3))  # too many

This PR adds some sanity checks and compares the exceptions of some scenarios with the exceptions from NumPy.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -425,6 +425,18 @@ class COO(object):
         if axes is None:
             axes = reversed(range(self.ndim))
 
+        # Normalize all axe indices to posivite values
+        try:
+            axes = np.arange(self.ndim)[list(axes)]
+        except IndexError as e:
+            raise ValueError("invalid axis for this array") from None
+
+        if len(np.unique(axes)) < len(axes):
+            raise ValueError("repeated axis in transpose")
+
+        if not len(axes) == self.ndim:
+            raise ValueError("axes don't match array")
+
         axes = tuple(axes)
 
         if axes == tuple(range(self.ndim)):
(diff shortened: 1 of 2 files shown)


sparse/core.py
```diff
@@ -429,7 +429,7 @@ def transpose(self, axes=None):
         try:
             axes = np.arange(self.ndim)[list(axes)]
         except IndexError as e:
-            raise ValueError("invalid axis for this array") from None
+            raise ValueError("invalid axis for this array")
```
hameerabbasi: This should just be `except IndexError:`

hameerabbasi:
It might be easier and simpler to just:

1. Add `ndim` to all negative axes.
2. Check if `tuple(sorted(axes)) == tuple(range(ndim))`

nils-werner:
> 1. Add ndim to all negative axes.

incorporated that idea

> 2. Check if tuple(sorted(axes)) == tuple(range(ndim))

that would check for all kinds of errors simultaneously, and wouldn't allow us to print individual error messages, like NumPy does

hameerabbasi:
That's a good point. This seems fairly trivial to me, so I'm merging immediately. :-)

nils-werner:
You were too quick, I hadn't pushed the last commit yet :-D

nils-werner:
You can cherry-pick 1f7ea72 to get the last change

hameerabbasi:
Done!

PR #64, opened 2018-01-06
Use higher density for sparse.random in tests

### LINKED ISSUE #63: Tests using sparse.random needs higher `density`
Hi, all

I noticed that many tests with `sparse.random` uses default density value, 0.01, 
which is too small for small matrices.

For example,
https://github.com/mrocklin/sparse/blob/2c0483d6bce1cd2a49624e72d396872f2aa7e23f/sparse/tests/test_core.py#L605-L607
would be all zero with high probability.

I will send a PR

fujiisoup:
Fixes #63.

Actually, one test fails with a higher density.

### DIFF
--- a/sparse/tests/test_core.py
+++ b/sparse/tests/test_core.py
@@ -92,7 +92,7 @@ def test_transpose_error(axis):
     [(), ()],
 ])
 def test_reshape(a, b):
-    s = sparse.random(a)
+    s = sparse.random(a, density=0.5)
     x = s.todense()
 
     assert_eq(x.reshape(b), s.reshape(b))
@@ -111,13 +111,13 @@ def test_large_reshape():
 
 
 def test_reshape_same():
-    s = sparse.random((3, 5))
+    s = sparse.random((3, 5), density=0.5)
 
     assert s.reshape(s.shape) is s
 
 
 def test_to_scipy_sparse():
-    s = sparse.random((3, 5))
+    s = sparse.random((3, 5), density=0.5)
     a = s.to_scipy_sparse()
     b = scipy.sparse.coo_matrix(s.todense())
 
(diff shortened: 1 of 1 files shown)


sparse/tests/test_core.py
```diff
@@ -111,13 +111,13 @@ def test_large_reshape():
...
 def test_reshape_same():
-    s = sparse.random((3, 5))
+    s = sparse.random((3, 5), density=0.5)
 
     assert s.reshape(s.shape) is s
 
 
 def test_to_scipy_sparse():
-    s = sparse.random((3, 5))
+    s = sparse.random((3, 5), density=0.5)
```
fujiisoup: This test fails with higher density.

sparse/tests/test_core.py
```diff
@@ -121,8 +121,9 @@ def test_to_scipy_sparse():
     a = s.to_scipy_sparse()
     b = scipy.sparse.coo_matrix(s.todense())
 
-    assert_eq(a.data, b.data)
     assert_eq(a.todense(), b.todense())
+    assert_eq(a.data[np.lexsort((a.row, a.col))],
+              b.data[np.lexsort((b.row, b.col))])
```
fujiisoup: Just added a workaround, but I think this check is not necessary here.
(Do we want to preserve the order of the entries?)

hameerabbasi: No, we do not want to preserve the order of entries. It's worth looking into why this test failed. I may take a look later today... I suspect it's because of out of order entries but I can't say much more.

hameerabbasi: I looked into this. I'd say it's safe to remove this check, it was meant for when the entries were sorted. As long as `todense` is okay, we should be good.

hameerabbasi:
Some comments on the failing test.

fujiisoup:
Thanks. I just removed this check.

PR #67, opened 2018-01-06
Support left-side np.number elemwise operations.

### LINKED ISSUE #66: left-side np.scalar multiplication
Hi,

I noticed multiplication with `np.scalar` such as `np.float32` fails,

```python
In [1]: import numpy as np
   ...: import sparse
   ...: x = sparse.random((2, 3, 4), density=0.5)
   ...: x * np.float32(2.0)  # This succeeds
   ...: 
Out[1]: <COO: shape=(2, 3, 4), dtype=float64, nnz=12, sorted=False, duplicates=True>

In [2]: np.float32(2.0) * x  # fails
AttributeError                            Traceback (most recent call last)
<ipython-input-2-e1d24f7d85b3> in <module>()
----> 1 np.float32(2.0) * x  # fails

~/Dropbox/projects/sparse/sparse/core.py in __array_ufunc__(self, ufunc, method, *inputs, **kwargs)
    491     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
    492         if method == '__call__':
--> 493             return COO._elemwise(ufunc, *inputs, **kwargs)
    494         elif method == 'reduce':
    495             return COO._reduce(ufunc, *inputs, **kwargs)

~/Dropbox/projects/sparse/sparse/core.py in _elemwise(func, *args, **kwargs)
    735             other = args[1]
    736             if isinstance(other, COO):
--> 737                 return self._elemwise_binary(func, *args[1:], **kwargs)
    738             elif isinstance(other, scipy.sparse.spmatrix):
    739                 other = COO.from_scipy_sparse(other)

AttributeError: 'numpy.float32' object has no attribute '_elemwise_binary'
```

fujiisoup:
Fixes #66.

This adds a workaround when the np.number multiplication from left-side.

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 from collections import Iterable, defaultdict, deque
-from functools import reduce
+from functools import reduce, partial
 import numbers
 import operator
 
@@ -728,8 +728,13 @@ class COO(object):
         self = args[0]
         if isinstance(self, scipy.sparse.spmatrix):
             self = COO.from_numpy(self)
-        if isinstance(self, np.number):
-            self = COO.from_numpy(np.atleast_1d(self))
+        elif np.isscalar(self) or (isinstance(self, np.ndarray)
+                                   and self.ndim == 0):
+            func = partial(func, self)
+            other = args[1]
+            if isinstance(other, scipy.sparse.spmatrix):
+                other = COO.from_numpy(other)
+            return other._elemwise_unary(func, *args[2:], **kwargs)
 
         if len(args) == 1:
             return self._elemwise_unary(func, *args[1:], **kwargs)
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -728,6 +728,8 @@ class COO(object):
         self = args[0]
         if isinstance(self, scipy.sparse.spmatrix):
             self = COO.from_numpy(self)
+        if isinstance(self, np.number):
+            self = COO.from_numpy(np.atleast_1d(self))
 
         if len(args) == 1:
             return self._elemwise_unary(func, *args[1:], **kwargs)
--- a/sparse/tests/test_core.py
+++ b/sparse/tests/test_core.py
@@ -266,8 +266,11 @@ def test_op_scipy_sparse():
     (operator.le, -3),
     (operator.eq, 1)
 ])
-def test_elemwise_scalar(func, scalar):
+@pytest.mark.parametrize('convert_to_np_number', [True, False])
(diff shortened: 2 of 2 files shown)


hameerabbasi:
Hello, the correct way to fix this would be to modify `__rmul__` to return `scalar * COO`. Fixing it this way causes the broadcasting path to be taken which will be much, much slower and will fail for scalar `COO` objects, returning 1-D instead of scalar.

fujiisoup:
Hi, @hameerabbasi 
Thanks for the quick response.

> the correct way to fix this would be to modify __rmul__ to return scalar * COO

`np.number` has `__mul__` method and thus `COO.__rmul__` will not be called.
Instead, `np.number.__mul__` calls `COO.__array_ufunc__`.

Another option would be to have a special treatment for the `func is operator.mul` case in `__array_ufunc__`, but the code would be more complicated.

hameerabbasi:
Hey! I came up with a long-term solution to this, hope it will be of use.

```
@staticmethod
    def _elemwise(func, *args, **kwargs):
        assert len(args) >= 1
        self = args[0]
        if isinstance(self, scipy.sparse.spmatrix):
            self = COO.from_numpy(self)
        elif np.isscalar(self):
            func = partial(func, self)
            other = args[1]
            if isinstance(other, scipy.sparse.spmatrix):
                other = COO.from_scipy_sparse(other)

            return other._elemwise_unary(func, *args[2:], **kwargs)

        if len(args) == 1:
            return self._elemwise_unary(func, *args[1:], **kwargs)
        else:
            other = args[1]
            if isinstance(other, scipy.sparse.spmatrix):
                other = COO.from_scipy_sparse(other)

            if isinstance(other, COO):
                return self._elemwise_binary(func, *args[1:], **kwargs)
            else:
                return self._elemwise_unary(func, *args[1:], **kwargs)
```

If you check this in, maybe @mrocklin can review it.

hameerabbasi:
@fujiisoup I've edited the elemwise to account for the unit test failure.

fujiisoup:
Cool!!

I added more tests for left-side operations.
Additionally, I added a support also for 0-dimensional `np.ndarray`.

hameerabbasi:
Would you be kind enough to rebase this on `master`? Your other PR caused conflicts with this one.

hameerabbasi:
Just waiting on either of @nils-werner or @mrocklin to review these changes... Since I was involved in adding code, I shouldn't be the one to review it. :-)

mrocklin:
Thank you for the fix @fujiisoup .  I'm sorry I wasn't around to review (I'm in conference prep mode).  Thank you for handling this @hameerabbasi .

PR #68, opened 2018-01-10
len() function for COO objects

nils-werner:
Quite trivial really, implements the `len()` operation following NumPy's convention of

    len(array) == array.shape[0]

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -269,7 +269,7 @@ class COO(object):
 
         Examples
         --------
-        >>> x = np.zeros(10, 10)
+        >>> x = np.zeros((10, 10))
         >>> s = COO.from_numpy(x)
         >>> len(s)
         10
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -253,6 +253,9 @@ class COO(object):
     def nbytes(self):
         return self.data.nbytes + self.coords.nbytes
 
+    def __len__(self):
+        return self.shape[0]
+
     def __sizeof__(self):
         return self.nbytes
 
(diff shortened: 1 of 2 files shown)


nils-werner:
Added them

hameerabbasi:
Well, this looks good. Let me know if you're finished working on it, and I'll review and merge (don't want to be too fast this time).

nils-werner:
This is ready to be merged.

mrocklin:
Agreed.

mrocklin:
Thanks @nils-werner !

nils-werner:
One unexpected sideeffect of this:

    numpy.array(cooarr)

now magically works!

    x = sparse.random((20, 30))
    np.allclose(x.todense(), np.array(x))    # True

hameerabbasi:
I'm going to assume it's really slow. Do we have any benchmarks for this?

nils-werner:
Yes, it's very very slow.

If I understand it correctly, `numpy.array(x)` looks for `buffer` and `__array_interface__` when creating an array from an object. If it doesn't find one it simply iterates over it.

What if we implement `__array_interface__` that calls `self.todense()`?

    @property
    def __array_interface__(self):
        return {
            'shape': self.shape,
            'data': self.todense(),
            'typestr': self.dtype.str,
        }

Before

    x = sparse.random((20, 30, 20))

    %timeit numpy.array(x)
    # 1.48 s ± 25.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
    %timeit x.todense()
    # 11.6 µs ± 40.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

After

    %timeit numpy.array(x)
    # 23 µs ± 392 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

hameerabbasi:
A pull request would be welcome. Also, your data is too small. Maybe it's just the overhear.

PR #43, opened 2017-12-26
First steps toward documentation

nils-werner:
Navigating this library without any docs is a bit tedious... so I thought I could fix that by starting to document the API.

If you are OK with me doing this I can continue pushing more and more docs commits into this PR until every function is properly documented.

For now you can generate the docs using

    pip install -e .[docs]
    python setup.py build_sphinx

and then opening `build/sphinx/html/index.html`.

To have a publicly accessible docs page you could register and set up a project at https://readthedocs.org/

PR #69, opened 2018-01-10
Size and density properties

nils-werner:
NumPy has a `.size` property that gives you the number of elements in an array. 

Additionally for sparse arrays a `.density` property may be handy, which is the ratio of nonzero elements to total elements in the array (a value between `0` and `1`).

This PR implements both properties for `COO` arrays

### DIFF
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -551,11 +551,11 @@ class COO(object):
 
         Examples
         --------
-        >>> x = np.zeros((10, 10))
+        >>> x = np.zeros((8, 8))
         >>> x[0, :] = 1
         >>> s = COO.from_numpy(x)
         >>> s.density
-        0.10000000000000001
+        0.125
         """
         return self.nnz / self.size
 
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -511,6 +511,14 @@ class COO(object):
         """
         return self.shape[0]
 
+    @property
+    def size(self):
+        return np.prod(self.shape)
+
+    @property
+    def density(self):
+        return self.nnz / self.size
+
     def __sizeof__(self):
         return self.nbytes
 
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -237,7 +237,7 @@ class COO(object):
         if self.coords.ndim == 1:
             self.coords = self.coords[None, :]
 
-        if shape and not np.prod(self.coords.shape):
+        if shape and not self.coords.size:
             self.coords = np.zeros((len(shape), 0), dtype=np.uint64)
 
         if shape is None:
@@ -748,7 +748,7 @@ class COO(object):
 
         if set(axis) == set(range(self.ndim)):
             result = method.reduce(self.data, **kwargs)
-            if self.nnz != np.prod(self.shape):
+            if self.nnz != self.size:
                 result = method(result, _zero_of_dtype(self.dtype)[()], **kwargs)
         else:
             axis = tuple(axis)
@@ -1277,7 +1277,7 @@ class COO(object):
         if self.shape == shape:
             return self
         if any(d == -1 for d in shape):
-            extra = int(np.prod(self.shape) /
+            extra = int(self.size /
                         np.prod([d for d in shape if d != -1]))
             shape = tuple([d if d != -1 else extra for d in shape])
 
@@ -1289,7 +1289,7 @@ class COO(object):
                 if sh == shape:
                     return value
 
-        # TODO: this np.prod(self.shape) enforces a 2**64 limit to array size
+        # TODO: this self.size enforces a 2**64 limit to array size
         linear_loc = self.linear_loc()
 
         max_shape = max(shape) if len(shape) != 0 else 1
@@ -1480,7 +1480,7 @@ class COO(object):
         # See https://github.com/scipy/scipy/blob/master/LICENSE.txt
         if not self.has_duplicates:
             return
-        if not np.prod(self.coords.shape):
+        if not self.coords.size:
             return
 
         self.sort_indices()
@@ -2029,9 +2029,7 @@ class COO(object):
             ...
         NotImplementedError: Operation would require converting large sparse array to dense
         """
-        elements = np.prod(self.shape)
-
-        if elements <= allowed_nnz or self.nnz >= elements * allowed_fraction:
+        if self.size <= allowed_nnz or self.density >= allowed_fraction:
             return self.todense()
         else:
             raise NotImplementedError("Operation would require converting "
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -513,10 +513,50 @@ class COO(object):
 
     @property
     def size(self):
+        """
+        The number of all elements (including zeros) in this array.
+
+        Returns
+        -------
+        int
+            The number of elements.
+
+        See Also
+        --------
+        numpy.ndarray.size : Numpy equivalent property.
+
+        Examples
+        --------
+        >>> x = np.zeros(10, 10)
+        >>> s = COO.from_numpy(x)
+        >>> s.size
+        100
+        """
         return np.prod(self.shape)
 
     @property
     def density(self):
+        """
+        The ratio of nonzero to all elements in this array.
+
+        Returns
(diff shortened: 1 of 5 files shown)


hameerabbasi:
Seems fine, however it would be really nice to have docstrings for these, and if possible, tests.

nils-werner:
Added them

hameerabbasi:
Ignore the Travis CI failure. Numpy 1.14 changed the way it converts arrays to strings, and that's causing a doctests failure. I've addressed it over in #43.

nils-werner:
Yup, I already noticed that

sparse/tests/test_core.py
```diff
@@ -875,3 +875,13 @@ def test_scalar_shape_construction():
     s = COO(coords, x, shape=5)
 
     assert_eq(x, s)
+
+
+def test_density():
+    s = sparse.random((20, 30, 40), density=0.1)
+    assert 0.09 < s.density < 0.11
```
hameerabbasi: It might be wise to use `np.isclose` here.

sparse/core.py
```diff
@@ -253,6 +253,54 @@ def nnz(self):
...
+        COO.size : Number of elements.
+        COO.nnz : Number of nonzero elements.
+
+        Examples
+        --------
+        >>> x = np.zeros((10, 10))
+        >>> x[0, :] = 1
+        >>> s = COO.from_numpy(x)
+        >>> s.density
+        0.10000000000000001
```
hameerabbasi: This might fluctuate in different system floating-point implementations. Perhaps `np.isclose(density, 0.1)`?

mrocklin:
Or perhaps change 10 to 8 above so that the result is evenly divisible by
two

nils-werner:
@mrocklin that's clever! Done.

hameerabbasi:
This looks ready. If you rebase, I can do a final review and merge.

nils-werner:
Did you see my [code comment](https://github.com/mrocklin/sparse/pull/69/files/a3adb4c816b6bfd71de8d8966f3cb87ac3642a73)?

hameerabbasi:
Hmmm, that's strange, it isn't showing any of your comments in that link, or on any commits.

nils-werner:
Aha! after commenting on lines you have to also open the code review thingy.

sparse/core.py
```diff
@@ -1277,7 +1325,7 @@ def astype(self, dtype, out=None):
 
     def maybe_densify(self, allowed_nnz=1e3, allowed_fraction=0.25):
         """ Convert to a dense numpy array if not too costly.  Err othrewise """
-        if reduce(operator.mul, self.shape) <= allowed_nnz or self.nnz >= np.prod(self.shape) * allowed_fraction:
+        if self.size <= allowed_nnz or self.density >= allowed_fraction:
```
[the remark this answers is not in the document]
hameerabbasi: It actually does make sense. If the density is low, then the cost of a dense array is higher compared to that of a sparse array, and we should avoid densifying.

If, however, density is high, the overhead of a sparse matrix far outweighs the benefits and we should densify.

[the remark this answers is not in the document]
nils-werner: True. Maybe I got confused that one `allowed` means `max` and the other one means `min`... What about renaming them to

    maybe_densify(self, max_nnz=1e3, min_density=0.25)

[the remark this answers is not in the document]
hameerabbasi: I would make sure that no tests break either here or in Dask before doing that. We can change unit tests here... But it's best to send a PR to Dask if any tests break there.

hameerabbasi:
Is `NotImplementedError` the best way to show that densfication is forbidden? What would be the right exception here?

sparse/core.py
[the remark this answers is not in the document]
hameerabbasi: Let's change it anyway and I'll send a PR to Dask if anything fails for them.

nils-werner:
I would say `ValueError` is the right one, as it's the function arguments that make the operation fail. A `NotImplementedError` sounds more like "whatever you pass to this function, it will never succeed".

sparse/core.py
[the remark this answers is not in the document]
nils-werner: Actually, shouldn't `max_nnz` be actually named `max_size`? We are not comparing it to `.nnz` anyways...

hameerabbasi:
Best to switch that to a `ValueError` then.

sparse/core.py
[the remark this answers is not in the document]
hameerabbasi: Yes, `max_size` is a good idea as well.

sparse/core.py
```diff
@@ -2412,19 +2412,19 @@ def maybe_densify(self, allowed_nnz=1000, allowed_fraction=0.25):
...
         an error.
 
         >>> x = np.zeros((5, 5), dtype=np.uint8)
         >>> x[2, 2] = 1
         >>> s = COO.from_numpy(x)
-        >>> s.maybe_densify(allowed_nnz=5, allowed_fraction=0.25)
+        >>> s.maybe_densify(max_size=5, min_fraction=0.25)
         Traceback (most recent call last):
             ...
         NotImplementedError: Operation would require converting large sparse array to dense
```
hameerabbasi: You need to update this line in the docstring. [See here.](https://docs.python.org/3/library/doctest.html#what-about-exceptions)

sparse/core.py
```diff
@@ -2381,17 +2381,17 @@ def astype(self, dtype, out=None):
         assert out is None
         return self.elemwise(np.ndarray.astype, dtype)
 
-    def maybe_densify(self, allowed_nnz=1000, allowed_fraction=0.25):
+    def maybe_densify(self, max_size=1000, min_fraction=0.25):
```
hameerabbasi: Best keep this as `min_density`, it's clearer that it represents the density.

sparse/core.py
```diff
@@ -2364,25 +2412,23 @@ def maybe_densify(self, allowed_nnz=1000, allowed_fraction=0.25):
...
         an error.
 
         >>> x = np.zeros((5, 5), dtype=np.uint8)
         >>> x[2, 2] = 1
         >>> s = COO.from_numpy(x)
-        >>> s.maybe_densify(allowed_nnz=5, allowed_fraction=0.25)
+        >>> s.maybe_densify(max_size=5, min_density=0.25)
         Traceback (most recent call last):
             ...
         NotImplementedError: Operation would require converting large sparse array to dense
```
hameerabbasi: Line 2425...

```
NotImplementedError: Operation would require converting large sparse array to dense
```

Need to change this to `ValueError`.

hameerabbasi: It's causing doctests to fail. :)

nils-werner: Currently there are a million tests failing on `master`, it is hard to spot my own mistakes...

hameerabbasi: That's strange, [Travis shows everything is working on master](https://travis-ci.org/mrocklin/sparse/branches).

mrocklin:
> Is NotImplementedError the best way to show that densfication is forbidden? What would be the right exception here?

To me `NotImplementedError` means "We have not *yet* implemented this, but it is a reasonable request".  If it is a case of a type not being supported like "we don't think that densifying scipy.sparse matrices is a good idea" then I would raise a TypeError.  If it is a case of the values being inappropriate like "sometimes we do densify sparse arrays, but this particular array is very very sparse" then I would use a ValueError.

sparse/core.py
nils-werner: I checked, they only fail on Numpy 1.12. Latest Numpy works fine...

hameerabbasi: That is really weird, I tested with Numpy 1.13.3 on my local system and it still works fine.

hameerabbasi:
So, we all agree that it should be `ValueError` in `maybe_densify` then. :-)

hameerabbasi:
Are you still making changes here? If not, this gets a +1 from me.

nils-werner:
No, I'm done.

hameerabbasi:
A rebase would be nice.

PR #72, opened 2018-01-10
Deal with np.array(sparsearr) densification

nils-werner:
Picking up from #68:

Implementing `len()` means that NumPy can create `np.ndarray`s from `COO` arrays using

    numpy.array(x)

But it's very very slow.

If I understand it correctly, `numpy.array(x)` looks for `buffer` and `__array_interface__` when creating an array from an object. If it doesn't find one it simply iterates over it (hence the slowness)

What if we implement `__array_interface__` that calls `self.todense()`?

```python
@property
def __array_interface__(self):
    return {
        'shape': self.shape,
        'data': self.todense(),
        'typestr': self.dtype.str,
    }

```

Before

```python
x = sparse.random((20, 30, 20))

%timeit numpy.array(x)
# 1.48 s ± 25.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit x.todense()
# 11.6 µs ± 40.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
numpy.allclose(x, x.todense)
# True
```

After

```python
%timeit numpy.array(x)
# 23 µs ± 392 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
numpy.allclose(x, x.todense)
# True
```

I can't explain where the 2x slowdown (11 vs 23 us) comes from though. The factor is also there for `x = sparse.random((20, 30, 40000))`: `78ms` vs `184ms`.

**But please be aware that I am phrasing this suggestion as a question!** I don't exactly know what NumPy does with the value returned from `__array_interface__`. Does it copy it? Does it share it? Are we leaking memory? Are we double-freeing it?

### DIFF
--- a/docs/generated/sparse.COO.rst
+++ b/docs/generated/sparse.COO.rst
@@ -68,6 +68,7 @@ COO
       COO.to_scipy_sparse
       COO.tocsc
       COO.tocsr
+      COO.__array__
 
    .. rubric:: :doc:`Other operations <../user_manual/operations/other>`
    .. autosummary::
--- a/docs/generated/sparse.COO.__array__.rst
+++ b/docs/generated/sparse.COO.__array__.rst
@@ -1,0 +1,6 @@
+COO\.\_\_array\_\_
+==================
+
+.. currentmodule:: sparse
+
+.. automethod:: COO.__array__
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -488,13 +488,30 @@ class COO(object):
         """
         return self.data.nbytes + self.coords.nbytes
 
-    @property
-    def __array_interface__(self):
-        return {
-            'shape': self.shape,
-            'data': self.todense(),
-            'typestr': self.dtype.str,
-        }
+    def __array__(self, dtype=None):
+        """
+        Helper function to speed up :code:`np.array(x)` conversion
+
+        Parameters
+        ----------
+        dtype: type
+            Datatype of dense array. The sparse array is cast before densification.
+
+        Returns
+        -------
+        array_like
+            The dense array.
+
+        See Also
+        --------
+        numpy.ndarray.__array__ : Numpy equivalent function.
+
+        """
+        if dtype is not None:
+            tmp = self.astype(dtype)
+        else:
+            tmp = self
+        return tmp.todense()
 
     def __len__(self):
         """
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -488,6 +488,14 @@ class COO(object):
         """
         return self.data.nbytes + self.coords.nbytes
 
+    @property
+    def __array_interface__(self):
+        return {
+            'shape': self.shape,
+            'data': self.todense(),
+            'typestr': self.dtype.str,
+        }
+
     def __len__(self):
         """
         Get "length" of array, which is by definition the size of the first
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -503,7 +503,9 @@ class COO(object):
         Returns
         -------
         NotImplemented
-            We do not implement this function, so this is what we return.
+            We do not implement this function, so we return
+            :obj:`NotImplemented`.
+
         See Also
         --------
         numpy.ndarray.__array__ : Numpy equivalent function.
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -490,27 +490,26 @@ class COO(object):
 
     def __array__(self, dtype=None):
         """
-        Helper function to speed up :code:`np.array(x)` conversion
+        Helper function that gets called during :code:`np.array(x)` conversion.
+        We deliberately return :code:`NotImplemented` to prevent accidental
+        densification.
 
         Parameters
         ----------
         dtype: type
-            Datatype of dense array. The sparse array is cast before densification.
+            Datatype requested by :code:`np.array(x)`. Has no effect on
+            output.
 
         Returns
         -------
-        array_like
-            The dense array.
-
+        NotImplemented
+            We do not implement this function, so this is what we return.
         See Also
         --------
         numpy.ndarray.__array__ : Numpy equivalent function.
 
         """
-        if dtype is not None:
-            return self.astype(dtype).maybe_densify()
-        else:
-            return self.maybe_densify()
+        return NotImplemented
 
     def __len__(self):
         """
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -2397,7 +2397,8 @@ class COO(object):
         >>> s.maybe_densify(allowed_nnz=5, allowed_fraction=0.25)
         Traceback (most recent call last):
             ...
-        NotImplementedError: Operation would require converting large sparse array to dense
+        NotImplementedError: Operation would require converting large sparse \
+array to dense. Use .todense() to force densification.
         """
         elements = np.prod(self.shape)
 
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -508,10 +508,9 @@ class COO(object):
 
         """
         if dtype is not None:
-            tmp = self.astype(dtype)
+            return self.astype(dtype).todense()
         else:
-            tmp = self
-        return tmp.todense()
+            return self.todense()
 
     def __len__(self):
         """
--- a/sparse/core.py
+++ b/sparse/core.py
@@ -508,9 +508,9 @@ class COO(object):
 
         """
         if dtype is not None:
-            return self.astype(dtype).todense()
+            return self.astype(dtype).maybe_densify()
         else:
-            return self.todense()
+            return self.maybe_densify()
 
     def __len__(self):
         """
@@ -2406,7 +2406,8 @@ class COO(object):
             return self.todense()
         else:
             raise NotImplementedError("Operation would require converting "
-                                      "large sparse array to dense")
+                                      "large sparse array to dense. Use "
+                                      ".todense() to force densification.")
 
 
 def tensordot(a, b, axes=2):
--- a/sparse/tests/test_core.py
+++ b/sparse/tests/test_core.py
@@ -892,5 +892,11 @@ def test_len():
 def test_numpy_todense():
     # make sure that .astype(int) is never all zeros
     s = sparse.random((20, 30, 40)) * 100
-    assert np.allclose(np.array(s, dtype=int), s.astype(int).todense())
-    assert np.allclose(np.array(s), s.todense())
+    with pytest.raises(ValueError):
+        assert np.array(s)
+
+    with pytest.raises(ValueError):
+        assert np.array(s, dtype=int)
+
+    with pytest.raises(ValueError):
+        assert np.allclose(s, s.todense())
--- a/sparse/tests/test_core.py
+++ b/sparse/tests/test_core.py
@@ -887,3 +887,10 @@ def test_scalar_shape_construction():
 def test_len():
     s = sparse.random((20, 30, 40))
     assert len(s) == 20
+
+
+def test_numpy_todense():
+    # make sure that .astype(int) is never all zeros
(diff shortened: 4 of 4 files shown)


nils-werner:
[Lets see what SO has to say](https://stackoverflow.com/questions/48193107/using-array-interface-to-help-numpy-create-array-from-object)

hameerabbasi:
From the docs for [`__array_interface__`](https://docs.scipy.org/doc/numpy/reference/arrays.interface.html) it seems data sharing is done automatically, so we should be good as long as Numpy is free of bugs.

You might also want to do `'data': (self.todense(), True)`, since we're not writing to the original object. I suspect (but am not sure) that the slowness is due to the double allocation and copy. If we used `False`, it would be shared.

Since we're passing a complete Numpy array, the data sharing will be done for us. However, when using `__array_interface__`, the double allocation/copy can be a cause for slowness. If we used `False`, the data from the original array produced from `todense()` would be shared and (I suspect) be faster.

But since we're marking it as "writable" that might cause other side effects I'm not aware of. Common sense says this should not be the case, though, since a new copy is produced every time `__array_interface__` is called.

nils-werner:
One answer (now deleted) on SO suggested to use [`__array__`](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ndarray.__array__.html).

hameerabbasi:
Interesting. He was probably right to delete the answer, as it didn't answer the original question. That said, it does solve our problem.

I'm happy to merge either way (as this is clearly an improvement) but it'd be nice to see benchmarks either way.

nils-werner:
Where would you like to see them? Here as a comment, or as part of the codebase and docs?

Also, one thing to keep in mind is that

    x = sparse.random((200, 200, 200, 200, 200, 200), density=0.00000001)
    np.allclose(x, 0)

 now "just works" by silently converting the sparse array to a dense one. This could be a point of frustration for users, if they accidentally consume all their RAM in one innocent looking single line.

hameerabbasi:
Here, as a comment. Benchmarks should only be included in docs if there are changes across releases.

Also, `sparse.allclose` would be simple.

- Check shape.
- Call `sum_duplicates`
- Call `np.array_equal(coords1, coords2)`
- Call `np.allclose(data1, data2)`

Although this would end up comparing sparsity structure exactly. In any case, I think it's close to what we want. If you don't want to compare sparsity structure:

- Match coords1 and coords2
- Call np.allclose(data_matched1, data_matched2)
- np.isclose(data_unmatched1, 0).all()
- np.isclose(data_unmatched2, 0).all()

You might want to look at the logic we use in `_elemwise_binary` if you take this approach.

hameerabbasi:
On second thought, this PR in general might be a bad idea, until we implement a suitable framework for auto-densification (#10), as it would densify at most times, and densification would be implicit instead of explicit.

nils-werner:
> On second thought, this PR in general might be a bad idea

Not implementing this is even worse, as NumPy tries to densify the array anyways...

nils-werner:
Would be using `maybe_densify` a solution?

hameerabbasi:
It's okay for now, however we will need to find a resolution to #10 for the future. I'm looking into it now. Since Dask (and other parallel libraries) may use this, we might need thread local storage for the configuration options.

hameerabbasi:
There's another problem with this... If we do `x + y` where `x` is a `scipy.sparse.spmatrix`, and `y` is `COO`, the addition "indirectly" calls `np.asanyarray` and that tries to densify the `COO`... In a purely sparse situation!

I fear that with this solution we're moving more and more towards implicit densification. At this point, I think it's just best if we just raise a `NotImplementedError` in `__array__`, or return `NotImplemented`, so the operation above fails over to `COO.__radd__` and can be handled properly.

hameerabbasi:
If possible, could you add tests for the cases `scipy.sparse.spmatrix op COO`?

nils-werner:
What do you mean?

hameerabbasi:
Tests of the form `x op y` where `x` is `scipy.sparse.spmatrix` and `y` is `COO`. We would like to check that the result is `COO` as well.

docs/generated/sparse.COO.rst
```diff
@@ -68,6 +68,7 @@ COO
       COO.to_scipy_sparse
       COO.tocsc
       COO.tocsr
+      COO.__array__
```
hameerabbasi: Do we actually want to document double-underscore functions in the API docs? I'm not sure what the convention is for the Python community. I know documenting them in code is good for potential contributors.

hameerabbasi:
One minor question, otherwise I think this is good to merge.

hameerabbasi:
I just tested... Prior to merging #68, `np.array(COO)` returns a `COO` object. After that we run into weird bugs such as https://github.com/mrocklin/sparse/pull/72#issuecomment-357434676 and #78. I wonder if there's a way to go back to the old behavior while keeping the old behavior...

docs/generated/sparse.COO.rst
nils-werner: [Numpy does document it](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.__array__.html)

And I think in case of `__array__` we should, because it is likely to show unexpected behaviour that users should be able to look up.

nils-werner:
Another alternative would be

    def __array__(self, dtype=None):
        if dtype is None:
            dtype = self.dtype
        arr = np.array(1, dtype=dtype)
        # assign `self` to array to prevent recursive call of __array__()
        arr[()] = self
        return arr

which would be in line with what happens when you `np.array()` a non-rectangular list of values:

    a = sparse.random((20, 30, 40))
    np.array(a)               # ValueError: setting an array element with a sequence.
    np.array(a, dtype=object)
    np.array(a, dtype=int)    # ValueError: setting an array element with a sequence.

    b = [1, [1, 1]]
    np.array(b)               # ValueError: setting an array element with a sequence.
    np.array(b, dtype=object)
    np.array(b, dtype=int)    # ValueError: setting an array element with a sequence.

nils-werner:
Or

    def __array__(self, dtype=object):
        arr = np.array(1, dtype=dtype)
        # assign `self` to array to prevent recursive call of __array__()
        arr[()] = self
        return arr

Which allows straight up `np.array(a)` casting

    a = sparse.random((20, 30, 40))
    np.array(a)
    np.array(a, dtype=object)
    np.array(a, dtype=int)    # ValueError: setting an array element with a sequence.

hameerabbasi:
The way I see it, we have two ways we can really go:

1. Raise a `TypeError` (I think a `TypeError` is more appropriate here) and drop Numpy 1.12 support.
2. Implement the hack in https://github.com/mrocklin/sparse/pull/72#issuecomment-357890921 or https://github.com/mrocklin/sparse/pull/72#issuecomment-357771582 (we would need to see what NumPy 1.12 does when the `dtype` isn't `object` and mimic that) and support Numpy 1.12.

I'm open to both but slightly in favour of the first. The case with https://github.com/mrocklin/sparse/pull/72#issuecomment-357434676 succeeds with both 1 and 2, and even if we return `NotImplemented`, or raise `NotImplementedError` or `ValueError`.

Personally, I think if we implement 2, and run into problems again, we drop Numpy 1.12 support completely.

hameerabbasi:
I wish `np.ndarray` implemented some sort of abstract metaclass that we could inherit from to deal with all these problems. Inheriting from `np.ndarray` isn't really an option at this point.

hameerabbasi:
cc @mrocklin Your input here would be valuable.

hameerabbasi:
@nils-werner What do you think? I'm not sure how widespread Numpy 1.13 adoption is and whether we should put in the effort to support 1.12.

nils-werner:
I don't know. I don't understand all this new ufunc magic well enough to give a useful answer...

mrocklin:
> cc @mrocklin Your input here would be valuable.

I'm not sure I know enough about NumPy internals to quickly know the right thing to do here.  cc'ing @njsmith in case he has time to comment here (Also, Nathaniel, meet @hameerabbasi and @nils-werner , both of whom have been doing a lot of great work on this project over the last month.)

hameerabbasi:
@njsmith, just a quick run-down. We want `np.exp(COO)`, etc. to not call `np.array(COO)` but to default to our own `COO.exp()`. The same goes for `scipy.sparse.spmatrix + COO`, which calls `np.array` and then densifies our matrix instead of calling `COO.__radd__`.

We're basically looking for a way to tell Numpy 1.12- to NOT use `np.array` results.

hameerabbasi:
I've opened #84 to deal with this for now.

njsmith:
Definitely don't inherit from `np.ndarray`, that way lies nothing but suffering.

For `np.exp(COO)`, I think `__array_ufunc__` is pretty much the best/only solution you'll find. If that means people need to upgrade numpy then oh well. If they wanted to use old software they'd be using `scipy.sparse`, right?

In general, the plan is for numpy to continue adding stuff like `__array_ufunc__`, e.g. there have been rumblings about `__array_concatenate__`, and I think another priority is some kind of `asduckarray`. The ideal is that eventually you'll be able to implement all of these `__array_whatever__` functions and get something that basically acts as an array without being automatically densified. I would probably make `__array__` raise an error for now though – eventually it might make sense, but if you make it an error now then you can always implement it later; if you make it work now, then you're stuck with it. And you *definitely* don't want people to get in the habit of calling it and getting back an `ndarray([COO], dtype=object)`, that's just a mess.

For interaction with `scipy.sparse`, I dunno, it depends on the details of how that code works, which I'm not very familiar with. You'd have to talk to them. I would seriously consider simply not supporting it at all and telling people to convert all their matrices to `sparse` objects.

hameerabbasi:
Thanks a lot @njsmith!

hameerabbasi:
xref numpy/numpy#4164

hameerabbasi:
@nils-werner, as discussed in #84, would you mind adding the fix in https://github.com/mrocklin/sparse/pull/72#issuecomment-357771582 (with the slight exception that we should mimic numpy behaviour for `dtype != object`)?

It might be worthwhile converting the dtype with `np.result_type` before erroring.

mrocklin:
OK, I've taken a bit of time and read through things here.  In general I apologize for being absent recently.  Some thoughts.

1.  I'm in favor of `np.asarray(my_coo)` densifying.  This is dumb in some cases but I'd rather not get in people's way, even as they're being dumb.
2.  It would be nice to maintain compatibility with `scipy.sparse`.

It's also worth noting that Numpy saves us a bit here with `MemoryError`s.  So if we don't do anything here then we're just relying on their policies, which I like.

```python
In [1]: from sparse import COO, random

In [2]: x = random((100000, 100000), density=0.000001)

In [3]: x
Out[3]: <COO: shape=(100000, 100000), dtype=float64, nnz=10000, sorted=False, duplicates=True>

In [4]: import numpy as np

In [5]: np.array(x)
MemoryError                               Traceback (most recent call last)
<ipython-input-5-152940d1f14f> in <module>()
----> 1 np.array(x)

/home/mrocklin/workspace/sparse/sparse/coo.py in __array__(self, dtype, **kwargs)
   1230 
   1231     def __array__(self, dtype=None, **kwargs):
-> 1232         x = self.todense()
   1233         if dtype and x.dtype != dtype:
   1234             x = x.astype(dtype)

/home/mrocklin/workspace/sparse/sparse/coo.py in todense(self)
    353         """
    354         self.sum_duplicates()
--> 355         x = np.zeros(shape=self.shape, dtype=self.dtype)
    356 
    357         coords = tuple([self.coords[i, :] for i in range(self.ndim)])

MemoryError: 
```

mrocklin:
Here is a strawman proposal: https://github.com/mrocklin/sparse/pull/87

PR #75, opened 2018-01-12
Support mixed sparse-dense support when result is sparse

hameerabbasi:
Lots of changes here, but small ones.

- Added a lot more operators, even if they fail consistently.
- Added support for operators from both sides.
- Moved a lot of functions out below the class.
- (Most important) Where sparse/dense mixed operations are sparse (such as with `operator.mul` and `operator.and_`), support that with broadcasting and return a `COO` array.


hameerabbasi:
This also supports cases like `sparse / dense` where `(dense != 0).all()` holds and `sparse ** dense` where `(dense > 0).all()` holds.

hameerabbasi:
Merging at 18:00 German time if there are no further proposed changes.

PR #76, opened 2018-01-13
Update contributing.rst.

hameerabbasi:
Updated the contributing guidelines for the repository.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -23,6 +23,7 @@ To get started, you can do something along the lines of the following, after
 creating a fork:
 
 .. code-block:: bash
+
    git pull https://github.com/<username>/sparse.git
    git remote add upstream https://github.com/mrocklin/sparse.git
    git install -e .[docs, tests]
@@ -46,6 +47,7 @@ yourself. Once you've configured your environment, you can just :code:`cd` to
 the root of your repository and run
 
 .. code-block:: bash
+
    py.test
 
 Adding/Building the Documentation
@@ -67,6 +69,7 @@ To build the documentation, you can :code:`cd` into the :code:`docs` directory
 and run
 
 .. code-block:: bash
+
    sphinx-build -b html . _build/html
 
 After this, you can find an HTML version of the documentation in :code:`docs/_build/html/index.html`.
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -26,7 +26,7 @@ creating a fork:
 
    git pull https://github.com/<username>/sparse.git
    git remote add upstream https://github.com/mrocklin/sparse.git
-   git install -e .[docs, tests]
+   pip install -e .[docs, tests]
 
 Then you can make changes to a branch on your fork and open a pull request
 on GitHub. There, it can be discussed, and if the community decides it's
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -1,7 +1,72 @@
 Contributing to sparse
 ======================
 
+General Guidelines
+------------------
 sparse is a community-driven project on GitHub. You can find our
 `repository on GitHub <https://github.com/mrocklin/sparse>`_. Feel
 free to open issues for new features or bugs, or open a pull request
 to fix a bug or add a new feature.
+
+If you haven't contributed to open-source before, we recommend you read
+`this excellent guide by GitHub on how to contribute to open source
+<https://opensource.guide/how-to-contribute/>`_. The guide is long,
+so you can gloss over things you're familiar with.
+
+Guidelines Specific to this Project
+-----------------------------------
+If you're not already familiar with it, we follow the `fork and pull model
+<https://help.github.com/articles/about-collaborative-development-models/>`_
+on GitHub.
+
+To get started, you can do something along the lines of the following, after
+creating a fork:
+
+.. code-block:: bash
+   git pull https://github.com/<username>/sparse.git
+   git remote add upstream https://github.com/mrocklin/sparse.git
+   git install -e .[docs, tests]
+
+Then you can make changes to a branch on your fork and open a pull request
+on GitHub. There, it can be discussed, and if the community decides it's
+best, it will be merged into the main codebase. Afterwards, you can
+`sync your fork on GitHub <https://help.github.com/articles/syncing-a-fork/>`_
+
+Running/Adding Unit Tests
+-------------------------
+It is best if all new functionality and/or bug fixes have unit tests added
+with each use-case.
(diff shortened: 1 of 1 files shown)


hameerabbasi:
cc @mrocklin It'd be nice if you looked this over and see if there's something I missed.

docs/contributing.rst
```diff
@@ -1,7 +1,72 @@
...
+<https://help.github.com/articles/about-collaborative-development-models/>`_
+on GitHub.
+
+To get started, you can do something along the lines of the following, after
+creating a fork:
+
+.. code-block:: bash
+   git pull https://github.com/<username>/sparse.git
+   git remote add upstream https://github.com/mrocklin/sparse.git
+   git install -e .[docs, tests]
```
mrocklin: git->pip

mrocklin: However I'm also somewhat concerned about having basic github usage documentation in our docs.  I would like to reserve the right to delete all of this in the future if it becomes something we have to maintain.

docs/contributing.rst
```diff
@@ -1,7 +1,72 @@
...
+Since we support both Python 2.7 and Python 3.5 and newer, it is recommended
+to test with at least these two versions before committing your code or opening
+a pull request. We use `pytest <https://docs.pytest.org/en/latest/>`_ as our unit
+testing framework, with the pytest-cov extension to check code coverage and
+pytest-flake8 to check code style. You don't need to configure these extensions
+yourself. Once you've configured your environment, you can just :code:`cd` to
+the root of your repository and run
+
+.. code-block:: bash
+   py.test
```
mrocklin: I usually have to run `py.test sparse`

Also, do you need a blank line between code-block and py.test?  I've always put this there, but I'm not sure if its necessary.

docs/contributing.rst
```diff
@@ -1,7 +1,72 @@
...
+
+We use the `Sphinx Autosummary extension <http://www.sphinx-doc.org/en/stable/ext/autosummary.html>`_
+to generate API references. In particular, you may want do look at the :code:`docs/generated`
+directory to see how these files look and where to add new functions, classes or modules.
+
+To build the documentation, you can :code:`cd` into the :code:`docs` directory
+and run
+
+.. code-block:: bash
+   sphinx-build -b html . _build/html
```
mrocklin: This also does the autogeneration of the rst files?

docs/contributing.rst
hameerabbasi: Yes, I put those in in a later commit. It's necessary.

`py.test` works for me at least. I just tested... Are you running it from the root of the repo?

docs/contributing.rst
hameerabbasi: No, it doesn't. You need to create (at the very least) `generated/sparse.rst` yourself with stock Sphinx. I've checked the `generated/` directory in to avoid building those, also because I customized `sparse.COO.rst` (it was just one big list, I wanted to add some structure to it), so the auto-generation would modify that file.

To actually auto-generate the files in `generated/`, you'd need to modify Sphinx itself and I don't want to put that in the docs.

docs/contributing.rst
hameerabbasi: Good point, I'll remove that, and fix that.

nils-werner:
If we support Python 2 and Python 3, why not [use tox to run tests for both versions](https://github.com/mrocklin/sparse/compare/master...nils-werner:tox)?

docs/contributing.rst
```diff
@@ -1,7 +1,59 @@
...
+We use the `Sphinx Autosummary extension <http://www.sphinx-doc.org/en/stable/ext/autosummary.html>`_
+to generate API references. In particular, you may want do look at the :code:`docs/generated`
+directory to see how these files look and where to add new functions, classes or modules.
+
+To build the documentation, you can :code:`cd` into the :code:`docs` directory
+and run
+
+.. code-block:: bash
+
+   sphinx-build -b html . _build/html
```
nils-werner: What's the advantage of this over `python setup.py build_sphinx`?

nils-werner: Also, `_build/` is not in `.gitignore`.

hameerabbasi: [It's what RTD runs](https://docs.readthedocs.io/en/latest/builds.html#understanding-what-s-going-on), so it automatically ensures compatibility for the most part.

hameerabbasi:
I would prefer to have Python 2 and 3 failures separate in Travis CI... Gives us an immediate indication of what configurations fail. It's helped me before.

nils-werner:
Travis CI would run `py.test` separately on two images, but users could run `tox` on their local machines.

hameerabbasi:
In that case, I have no objections.

Also, we should move `py.test` options from `setup.cfg`, it's deprecated in version 3.x and will be removed in 4.x.

docs/contributing.rst
mrocklin: I think that we need documentation on how to build the documentation.  It sounds like right now there are procedures that currently reside only within your head.  This makes me concerned because at some point in the future you may decide to move on to other projects and then people will have to spend a lot of time to figure out what needs to be done.

I think that we should either document the process or else change the process so that it is simpler.

hameerabbasi: I may have been a bit opaque in my original description, so let me clarify...

* To build the docs, all you need is stock Sphinx and that one command.
* To update the docs after an API change, all you need is some knowledge of the [Sphinx autosummary extension](http://www.sphinx-doc.org/en/stable/ext/autosummary.html). Basically, just updating the lists in `generated/`.

The modification to Sphinx was a one-time thing and isn't needed anymore.

I will make clear the second point because that's all that's needed at this point.

mrocklin: If I were to add a new function then is that one command sufficient?  If not, what should we do in the future if someone adds a new function?

hameerabbasi: That is exactly my second point. You update the lists in `docs/generated` for that specific module/class that you place the function in, and everything else will be generated for you. For example, if I were to add a function to `COO`, I would go to `sparse.COO.rst`, add in the name of the function (just the name, not the whole docstring) where appropriate (standard autosummary stuff, nothing specific to us), and it will generate it for us.

mrocklin: OK, that sounds good.  Thank you for the explanation.

docs/contributing.rst
hameerabbasi: `docs/_build` is in gitignore, that's where this command is supposed to be run.

PR #77, opened 2018-01-15
Tox for cross Python-version testing

nils-werner:
Note that the way Travis CI and `py.test` run their tests remain unchanged:

 - Travis CI runs multiple Python version tests in seprarate images
 - `py.test` works exactly as before

The only difference: Users can run `tox` on their local machines to test Python 2 and 3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@ setup(name='sparse',
       install_requires=list(open('requirements.txt').read().strip().split('\n')),
       extras_require={
           'tests': [
+              'tox',
               'pytest',
               'pytest-cov',
               'pytest-flake8',
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,8 @@
+[tox]
+envlist = py27,py36
+[testenv]
+commands=
+    py.test {posargs}
+extras=
+    docs
+    tests


hameerabbasi:
Can you also move the `py.test` options from `setup.cfg` to [`pytest.ini`](https://docs.pytest.org/en/latest/customize.html#adding-default-options), since it's deprecated in 3.x? (I can't find a link in the docs, but see pytest-dev/pytest#3086)

nils-werner:
The way I understood it was `[pytest]` was deprecated in favor of `[tool:pytest]`

hameerabbasi:
Ah, in that case... We should be good.

nils-werner:
Related to [this issue](https://github.com/mrocklin/sparse/pull/69#discussion_r161530365) and #71 it seems that `NumPy<=1.12.0` has some issues that are not present in NumPy latest. It may make sense to test for multiple Python versions and multiple NumPy versions:

```
[tox]
envlist = {py27,py36}-{numpy12,numpylatest}
[testenv]
deps =
   numpy12: numpy==1.12.0
basepython =
   py27: python2.7
   py36: python3.6
commands=
   pip list
   py.test {posargs}
extras=
   docs
   tests
```

Issue #78, opened 2018-01-15
NumPy 1.12.0 has failing tests


nils-werner:
Somewhere there is auto-densification happening, as I have many failing tests on my setup:

    sparse/tests/test_core.py::test_elemwise[func0] FAILED
    sparse/tests/test_core.py::test_elemwise[func1] FAILED
    sparse/tests/test_core.py::test_elemwise[func2] FAILED
    sparse/tests/test_core.py::test_elemwise[func3] FAILED
    sparse/tests/test_core.py::test_elemwise[func4] FAILED
    sparse/tests/test_core.py::test_elemwise[func5] FAILED
    sparse/tests/test_core.py::test_elemwise[func6] FAILED
    sparse/tests/test_core.py::test_elemwise[func7] FAILED
    sparse/tests/test_core.py::test_elemwise[func8] FAILED
    sparse/tests/test_core.py::test_elemwise[func9] FAILED
    sparse/tests/test_core.py::test_elemwise[func10] PASSED
    sparse/tests/test_core.py::test_elemwise[func11] FAILED
    sparse/tests/test_core.py::test_elemwise[func12] PASSED
    sparse/tests/test_core.py::test_elemwise[func13] FAILED
    sparse/tests/test_core.py::test_elemwise[func14] FAILED
    sparse/tests/test_core.py::test_elemwise[func15] PASSED
    sparse/tests/test_core.py::test_elemwise[func16] PASSED

with

    func = <ufunc 'floor'>

        @pytest.mark.parametrize('func', [np.expm1, np.log1p, np.sin, np.tan,
                                          np.sinh, np.tanh, np.floor, np.ceil,
                                          np.sqrt, np.conj, np.round, np.rint,
                                          lambda x: x.astype('int32'), np.conjugate,
                                          np.conj, lambda x: x.round(decimals=2), abs])
        def test_elemwise(func):
            s = sparse.random((2, 3, 4), density=0.5)
            x = s.todense()

            fs = func(s)

    >       assert isinstance(fs, COO)
    E       assert False
    E        +  where False = isinstance(array([[[ 0.,  0.,  0.,  0.],\n        [ 0.,  0.,  0.,  0.],\n        [ 0.,  0.,  0.,  0.]],\n\n       [[ 0.,  0.,  0.,  0.],\n        [ 0.,  0.,  0.,  0.],\n        [ 0.,  0.,  0.,  0.]]]), COO)

Additionally, the two tests

[...]

PR #79, opened 2018-01-15
Test NumPy 1.12 and latest in both Travis and tox

nils-werner:
I have now run into issues with NumPy 1.12 twice: #71 #78. Maybe it makes sense to test multiple Python versions and multiple NumPy versions?

This PR creates a testmatrix in both Travis and tox that tests all combinations of

 - Python 2.7
 - Python 3.6

and

 - NumPy 1.12.0
 - NumPy latest

In tox you can run a specific test on a version combination using

    tox -e py36-numpy12 -- -k testname

for `testname` on Python 3.6 and NumPy 1.12.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ matrix:
     - python: 2.7
     - python: 3.6
 
+env:
+  - NUMPY=1.12
+  - NUMPY=*
+
 install:
   # Install conda
   - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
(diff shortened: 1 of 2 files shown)


hameerabbasi:
Since we're making extensive use of the `__array_ufunc__` interface which was introduced in Numpy 1.13, we should require Numpy 1.13.0 rather than supporting legacy versions. It should be okay, since most packages (at least in the Anaconda ecosystem) have been updated to support Numpy 1.13.

nils-werner:
Wait but in #71 it was the other way round, no? :-)

hameerabbasi:
Merging #72 might help fix this... I'll re-run once that's merged. Otherwise I'm just in favor of dropping Numpy 1.12 support.

#71 was different because it was a minor behavioral change leading to a test failure. Here, it's a whole paradigm shift between how Numpy 1.12 and 1.13 handle operations in general.

hameerabbasi:
We're no longer supporting Numpy 1.12 (see #84 for all the benefits).

Issue #81, opened 2018-01-18
Operations of the form scipy.sparse.spmatrix operator COO return a dense result


hameerabbasi:
Should ideally be fixed with the final resolution for #72.

hameerabbasi:
Fixed with #84, except where Scipy makes it impossible.

PR #84, opened 2018-01-19
Drop support for Numpy 1.12.

hameerabbasi:
- Drop Numpy 1.12 support
- Implement all operators via `NDArrayOperatorsMixin`.
- Support for `scipy.sparse.spmatrix` operations from the left (mostly).
- Raise `TypeError` for `__array__`.


hameerabbasi:
Fixes #81.

hameerabbasi:
In addition, closes #78 as wontfix.

hameerabbasi:
I've updated this PR with the excellent suggestions from @njsmith.

@nils-werner @mrocklin Nice if you gave it a look.

nils-werner:
This is a radical change and needs a proper code review, but it would be a very sleek solution, if all works well.

mrocklin:
I agree with @nils-werner that this is sleek.  

The dependence on 1.13 is interesting.  I'm fine going forward with this, but we might want to release before we merge, just so we can point people not yet on latest numpy to that version.

> Also, it should warrant a major version bump, as it is pretty backwards-incompatible :-)

We haven't released in a good long while.  I'll add this (and instructions) to my TODO list if no one has objections.

hameerabbasi:
Here's what I propose (in no particular order):

- We tidy up #72 and merge it to a `0.2` (or whatever) branch.
- We discuss and make this one proper and merge this to master.

Even if we do the latter first, we can always make a `0.2` branch from the previous commit. Git is awesome that way. :)

hameerabbasi:
Or we can branch off from hameerabbasi/sparse@293ea067bd64a69a1d70e296348ac0a8b1035e4b :) That has working Scipy.sparse support and Numpy 1.12 support.

mrocklin:
Any objection to merging #72 to master, releasing, and then merging this one on top of it?  I have a slight desire to keep releases in the mainline history.  Happy either way though.  I'd also be happy if you wanted to just skip directly to dropping numpy 1.13.

hameerabbasi:
Releases will diverge from mainline history if we make backwards incompatible changes (as we are doing here) and then have to fix bugs in old releases. This is the reason why I would prefer any backwards-incompatible changes happen on master and the older version be in a separate branch.

This applies particularly to this PR and #72 because we're effectively undoing #72 by merging this, and it makes more sense to me intuitively that they should be in separate branches.

However, I would still prefer to release with #72 merged for the reason that not only are we dropping support for Numpy 1.12 (I have no problem with that part), but we're _also_ dropping support for inter-compatibility with `scipy.sparse` and requiring that all objects be explicitly converted to `COO` before any operations.

hameerabbasi:
I personally am fine with both, releasing with #72 or this PR. I don't know if if you have reservations, though. Because dropping Numpy 1.12 support only requires users to update Numpy, dropping scipy.sparse inter-op requires... well... Rewriting code.

mrocklin:
> This applies particularly to this PR and #72 because we're effectively undoing #72 by merging this, and it makes more sense to me intuitively that they should be in separate branches.

I'm totally fine having a commit in mainline history than we undo in a subsequent commit.  

> However, I would still prefer to release with #72 merged for the reason that not only are we dropping support for Numpy 1.12 (I have no problem with that part), but we're also dropping support for inter-compatibility with scipy.sparse and requiring that all objects be explicitly converted to COO before any operations.

Hrm, I hadn't realized the change with support for scipy.sparse.  Is this critical to achieve the nice refactor that you have here?  Was it desired?

mrocklin:
> I assume that's a typo. We're dropping Numpy 1.12, not 1.13. I'm not in favor of dropping old versions as soon as new ones come out. ;-)

Not a typo, just me being distracted and forgetting which versions of Numpy are out :)

hameerabbasi:
> I'm totally fine having a commit in mainline history than we undo in a subsequent commit.

In that case, we should be good.

> Hrm, I hadn't realized the change with support for scipy.sparse. Is this critical to achieve the nice refactor that you have here? Was it desired?

I added it back, I realized it was too drastic. In any case, supporting `scipy.sparse operator COO` seems to be a hit-and-miss. `add` and `sub` works fine, but `mul` tries to densify and `lt`/`gt`/`ne` all raise `TypeError`s. We should document, at the very least, that it is recommended to convert all arrays to `COO` before starting.

mrocklin:
Ah, mul for scipy.sparse will be matrix multiplication due to the np.matrix
interface I think

mrocklin:
I would not be surprised to see that behave oddly

hameerabbasi:
This seems okay to merge from my side.

hameerabbasi:
I removed calls of the form `COO.ufunc` because `np.ndarray` doesn't have them either. `np.ufunc(COO, ...)` still works.

hameerabbasi:
> Also, it should warrant a major version bump, as it is pretty backwards-incompatible :-)

We're only dropping support for Numpy 1.12. Scipy.sparse inter-op has been added back. This slims down our code quite a bit.

hameerabbasi:
I think this is ready for a code review + merge. :-)

PR #85, opened 2018-01-21
Add implementation for DOK.

hameerabbasi:
I've made a small DOK implementation. So far, it supports:

- Setting items with integers and slices, including setting with arrays.
- Getting single items only.
- Converting to/from COO with `COO(DOK)`/`DOK(COO)`
- Has a `todense()` for debugging.

I haven't added tests or docs (yet) or tested performance, but it seems to work fine.

Edit: xref #15


hameerabbasi:
You can do, for example,
```
x[1:3, 1:3] = 5
x[2, 2] = 3
x[1:3, 1:3] = [4, 6]
x[1:3, 1:3] = [[4, 5], [6, 7]]
```

And all of them should work. :)

sparse/dok.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @property
+    def nnz(self):
+        return len(self.dict)
+
+    def __getitem__(self, item):
+        item = normalize_index(item, self.shape)
+
+        for i in item:
+            if not isinstance(i, Integral):
```
mrocklin: perhaps `if not all(isinstance(i, Integral) for i in item):`

sparse/dok.py
```diff
@@ -0,0 +1,147 @@
...
+                raise IndexError('All indices must be integers'
+                                 ' when getting an item.')
+
+        if not len(item) == self.ndim:
+            raise IndexError('Can only get single elements.')
```
mrocklin: Might want to include the item in the error message with `"expected key of length %d, got %s" % (self.ndim, str(item))`

sparse/dok.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @property
+    def ndim(self):
+        return len(self.shape)
+
+    @property
+    def nnz(self):
+        return len(self.dict)
+
+    def __getitem__(self, item):
```
mrocklin: I would use `key` or `index` for this variable name

sparse/dok.py
```diff
@@ -0,0 +1,147 @@
...
+
+        for i in item:
+            if not isinstance(i, Integral):
+                raise IndexError('All indices must be integers'
+                                 ' when getting an item.')
+
+        if not len(item) == self.ndim:
+            raise IndexError('Can only get single elements.')
+
+        if item in self.dict.keys():
```
mrocklin: You want `if item in self.dict`.  This is much faster.  The `.keys()` result is either a list (python 2) or iterator (python 3), each of which have linear time membership checks.

sparse/dok.py
```diff
@@ -0,0 +1,147 @@
...
+                raise IndexError('All indices must be slices or integers'
+                                 ' when setting an item.')
+
+        self.dict[tuple(key_list)] = value[()]
+
+    def todense(self):
+        result = np.zeros(self.shape, dtype=self.dtype)
+
+        for c, d in six.iteritems(self.dict):
+            result[c] = d
```
mrocklin: You might be able to do something like `result[self.dict.keys()] = self.dict.values()` here

hameerabbasi: This works in Python 2, but not 3 (where these are iterators). In addition, it'll take up more memory for the resulting lists even if we were to do this.

hameerabbasi:
Tests added, and a few bugs caught and fixed. :)

Will do docs soon.

hameerabbasi:
I think this is ready for a final review and merge.

Keep in mind we can keep adding features later. Right now, we just need an implementation. :)

sparse/dok.py
```diff
@@ -0,0 +1,312 @@
...
+    <COO: shape=(5, 5), dtype=int64, nnz=4, sorted=False, duplicates=False>
+    >>> s2.todense()  # doctest: +NORMALIZE_WHITESPACE
+    array([[0, 0, 0, 0, 0],
+           [0, 4, 5, 0, 0],
+           [0, 6, 7, 0, 0],
+           [0, 0, 0, 0, 0],
+           [0, 0, 0, 0, 0]])
+    """
+    def __init__(self, shape, values=None, dtype=None):
+        """
```
mrocklin: I recommend only including the docstring on the main class and removing it from the `__init__` method.  People tend to check for docstrings like `DOK?` or `help(DOK)` rather than `help(DOK.__init__)`.  Also, the two versions here will almost certainly grow out of sync.

sparse/dok.py
```diff
@@ -0,0 +1,312 @@
...
+        >>> values = {
+        ...     (1, 2, 3): 4,
+        ...     (3, 2, 1): 0,
+        ... }
+        >>> s4 = DOK((5, 5, 5), values)
+        >>> s4
+        <DOK: shape=(5, 5, 5), dtype=int64, nnz=1>
+        """
+        from .coo import COO
+        self.dict = {}
```
mrocklin: It seems odd to me to use the name of a core collection as a variable name.  I'm not sure I have a better solution though.  Maybe `.data`?

docs/user_manual/building_dok.rst
```diff
@@ -0,0 +1,38 @@
...
+Building :obj:`COO` Arrays from :obj:`DOK` Arrays
+=================================================
+It's possible to build :obj:`COO` arrays from :obj:`DOK` arrays, if it is not
+easy to construct the :code:`coords` and :obj:`data` in a simple way. :obj:`DOK`
+arrays provide a simple builder interface to build :obj:`COO` arrays, but at
+this time, they can do little else.
+
+You can get started by defining the shape (and optionally, datatype) of the
+:obj:`DOK` array. If you do not specify a dtype, it is inferred from the first
+element.
```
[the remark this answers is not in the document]
hameerabbasi: I'm mimicking Numpy semantics here. Numpy doesn't change the data type of the whole array for just one element, it has to be decided beforehand.

docs/user_manual/building_dok.rst
```diff
@@ -0,0 +1,38 @@
...
+.. code-block:: python
+
+   s[1:3, 3:1:-1] = [[6, 5]]
+
+At the end, you can convert the :obj:`DOK` array to a :obj:`COO` array, and
+perform arithmetic or other operations on it.
+
+.. code-block:: python
+
+   s2 = COO(s)
```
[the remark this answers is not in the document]
hameerabbasi: I'll add `to_coo` and `from_coo`. Doesn't hurt, and they're tiny.

docs/user_manual/building_dok.rst
```diff
@@ -0,0 +1,38 @@
...
+arrays provide a simple builder interface to build :obj:`COO` arrays, but at
+this time, they can do little else.
+
+You can get started by defining the shape (and optionally, datatype) of the
+:obj:`DOK` array. If you do not specify a dtype, it is inferred from the first
+element.
+
+.. code-block:: python
+
+   s = DOK((6, 5, 2), [dtype=np.float64])
```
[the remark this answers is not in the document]
hameerabbasi: I'll change that real quick.

sparse/dok.py
```diff
@@ -0,0 +1,312 @@
...
+from .slicing import normalize_index
+from .utils import _zero_of_dtype
+
+try:  # Windows compatibility
+    int = long
+except NameError:
+    pass
+
+
+class DOK:
```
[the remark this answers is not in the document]
hameerabbasi: Tests succeed with Python 2.7... Which version of Python has issues with this?

sparse/dok.py
hameerabbasi: `.data` sounds good to me. I thought of that at first, but wanted to make sure users don't confuse it with `COO.data`

sparse/dok.py
hameerabbasi: That's how we're doing it in `COO`. I tend to keep the class docstring somewhat general (like an overview) and the `__init__` docstring specific to constructing objects.

sparse/dok.py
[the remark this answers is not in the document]
mrocklin: This is just general cleanliness in Python 2.  `class Foo:` creates an "old style class" which very few people expect these days.  People have grown used to new style classes (`class Foo(object)`) and their class methods.  These are also the default in Python 3.  

See https://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python

sparse/dok.py
mrocklin: Happy with either.  I don't have strong thoughts here.

sparse/dok.py
mrocklin: I recommend merging the docstring in `__init__` up to the class-level docstring.  I find that people rarely look at the `__init__` docstring in practice.

hameerabbasi:
Another round of reviews, maybe? :)

docs/user_manual/building_dok.rst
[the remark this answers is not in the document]
mrocklin: Yeah, I agree with that behavior.  My concern is that they might not realize that they are selecting a dtype, and then have the rest of their data silently truncated.  Consider the following example:

```python
dok = DOK((5, 5))

data = {(1, 1): 1, (2, 2), 2.5, (3, 3): 3.5}
for k, v in data.items():
    dok[k] = v
```

The first entry defines the dtype as int, but then the subsequent values are silently truncated.

It looks like scipy defaults dtype to `np.float64`.  We might want to do the same.

```python
In [1]: from scipy.sparse import dok_matrix

In [2]: dok_matrix((5, 5))
Out[2]: 
<5x5 sparse matrix of type '<class 'numpy.float64'>'
	with 0 stored elements in Dictionary Of Keys format>
```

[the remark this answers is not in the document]
hameerabbasi: Looks like Numpy does, too:
```
import numpy as np
x = np.zeros((5, 5))
x.dtype
Out[4]: dtype('float64')
y = np.array([3, 1.5])
y.dtype
Out[6]: dtype('float64')
```

I could use, for example, [numpy.result_type](https://docs.scipy.org/doc/numpy/reference/generated/numpy.result_type.html) on the initial data.

```
d = {0: x, 1: y}
np.result_type(*d.values())
Out[11]: dtype('float64')
```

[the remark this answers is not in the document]
hameerabbasi: I've fixed this now. It defaults to `float64`, or to the maximum dtype of `values` if it is provided.

sparse/dok.py
```diff
@@ -0,0 +1,380 @@
...
+
+        if not data:
+            data = {}
+
+        if isinstance(data, dict):
+            if not dtype:
+                if not len(data):
+                    self.dtype = np.dtype('float64')
+                else:
+                    self.dtype = np.result_type(*map(lambda x: np.asarray(x).dtype, data.values()))
```
hameerabbasi: I had to do `map` here as otherwise it would not work on `dicts` of `int`.

hameerabbasi:
And another round! 💯

mrocklin:
+1 from me

PR #87, opened 2018-01-25
Implement __array__ protocol

mrocklin:
One approach to the problems described in #72

### DIFF
--- a/sparse/coo.py
+++ b/sparse/coo.py
@@ -1232,8 +1232,6 @@ class COO(object):
         x = self.todense()
         if dtype and x.dtype != dtype:
             x = x.astype(dtype)
-        if not isinstance(x, np.ndarray):
-            x = np.array(x)
         return x
 
     def linear_loc(self, signed=False):
--- a/sparse/coo.py
+++ b/sparse/coo.py
@@ -1228,6 +1228,14 @@ class COO(object):
         else:
             return NotImplemented
 
+    def __array__(self, dtype=None, **kwargs):
+        x = self.todense()
+        if dtype and x.dtype != dtype:
+            x = x.astype(dtype)
+        if not isinstance(x, np.ndarray):
+            x = np.array(x)
+        return x
+
     def linear_loc(self, signed=False):
         """
         The nonzero coordinates of a flattened version of this array. Note that
--- a/sparse/tests/test_coo.py
+++ b/sparse/tests/test_coo.py
@@ -894,7 +894,18 @@ def test_scipy_sparse_interface():
 
     assert_eq(x, xx)
     assert_eq(x.T.dot(x), xx.T.dot(xx))
-    assert isinstance(x + xx, (COO, scipy.sparse.spmatrix))
+    assert isinstance(x + xx, COO)
+    assert isinstance(xx + x, COO)
+
+
+@pytest.mark.parametrize('scipy_format', ['coo', 'csr', 'dok', 'csc'])
+def test_scipy_sparse_interaction(scipy_format):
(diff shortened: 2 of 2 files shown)


mrocklin:
I haven't thought through this very deeply.  Criticism welcome from @nils-werner and @hameerabbasi

sparse/tests/test_coo.py
```diff
@@ -894,6 +894,7 @@ def test_scipy_sparse_interface():
 
     assert_eq(x, xx)
     assert_eq(x.T.dot(x), xx.T.dot(xx))
+    assert isinstance(x + xx, (COO, scipy.sparse.spmatrix))
```
hameerabbasi: Perhaps test _only_ for `COO`? Since we're "higher level" than `scipy.sparse`, it makes sense that mixed binary operations don't return `scipy.sparse.spmatrix`.

sparse/coo.py
```diff
@@ -1228,6 +1228,14 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         else:
             return NotImplemented
 
+    def __array__(self, dtype=None, **kwargs):
+        x = self.todense()
+        if dtype and x.dtype != dtype:
+            x = x.astype(dtype)
+        if not isinstance(x, np.ndarray):
```
hameerabbasi: Maybe [`np.asarray`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.asarray.html)? It passes through things that are already arrays.

But, in any case, do we really need this check, since `.todense()` already returns an `np.ndarray`.

sparse/tests/test_coo.py
hameerabbasi: In addition (just as a minor quip) I would parametrize this with `format` for scipy.sparse and check for (at least) `csr`, `csc`, `coo` and `dok`.

hameerabbasi:
Hmm. I just tested this some more:

```Jupyter Notebook
x * xx
  File "/Users/hameerabbasi/anaconda/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 2910, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-7-d2266dae7fac>", line 1, in <module>
    x * xx
  File "/Users/hameerabbasi/anaconda/lib/python3.6/site-packages/scipy/sparse/base.py", line 476, in __mul__
    raise ValueError('dimension mismatch')
ValueError: dimension mismatch
xx * x
Out[8]: <COO: shape=(100, 10), dtype=uint8, nnz=96, sorted=False, duplicates=False>
```

mrocklin:
Can you explain more about how you're getting that error?  I'm not able to reproduce.

sparse/tests/test_coo.py
mrocklin: Done

sparse/coo.py
mrocklin: Removed

hameerabbasi:
`add` and `sub` work fine, `mul` gives that error.

mrocklin:
Yeah, I mentioned this in the other PR, but just for completeness `__mul__` in scipy.sparse matrices will be matrix-multiplication, which I expect would be a bit stranger.  Personally I'm ok seeing this fail.

hameerabbasi:
In that case, this gets a +1 from me. :)

hameerabbasi:
We should merge this, I'll rebase #84 on top of it, and then I'm totally fine releasing.

mrocklin:
@hameerabbasi ok to release current master as 0.2.0 ?

hameerabbasi:
Of course. :)

PR #88, opened 2018-01-25
Add densification manager.

### LINKED ISSUE #10: Automatic conversion to dense arrays

hameerabbasi:
Fixes #10

A densification manager for `COO`.

You can do, for example,

```
with s.configure_densification(densify=True), s2.configure_densification(densify=True):
    x = np.exp(s)
    x2 = s + 1
    x3 = -(s + s2) + 1
```

And it should work automagically.

### DIFF
--- a/sparse/coo.py
+++ b/sparse/coo.py
@@ -1237,7 +1237,7 @@ class COO(SparseArray):
             return NotImplemented
 
     def __array__(self, dtype=None, **kwargs):
-        x = self.todense()
+        x = self.maybe_densify()
         if dtype and x.dtype != dtype:
             x = x.astype(dtype)
         return x
@@ -2016,10 +2016,19 @@ class COO(SparseArray):
 
     @contextmanager
     def configure_densification(self, densify, max_size=None, min_density=None):
-        old_densification_config = self._densification_config
-        self._densification_config = DensificationConfig(densify, max_size, min_density)
+        if isinstance(self._densification_config, DensificationConfig):
+            old_densification_config = DensificationConfig(self._densification_config)
+            self._densification_config.make_copy_of(DensificationConfig(densify, max_size, min_density))
+        else:
+            old_densification_config = self._densification_config
+            self._densification_config = DensificationConfig(densify, max_size, min_density)
+
         yield
-        self._densification_config = old_densification_config
+
+        if isinstance(old_densification_config, DensificationConfig):
+            self._densification_config.make_copy_of(old_densification_config)
+        else:
+            self._densification_config = old_densification_config
 
     @property
     def densification_config(self):
--- a/sparse/coo.py
+++ b/sparse/coo.py
@@ -2043,11 +2043,9 @@ class COO(SparseArray):
         return self.elemwise(np.ndarray.astype, dtype)
 
     @contextmanager
-    def configure_densification(self, densify, max_size=None, min_density=None):
+    def configure_densification(self, **kwargs):
         old_densification_config = self.densification_config
-        self.densification_config = DensificationConfig(densify=densify,
-                                                        max_size=max_size,
-                                                        min_density=min_density)
+        self.densification_config = DensificationConfig(**kwargs)
 
         yield
 
--- a/sparse/coo.py
+++ b/sparse/coo.py
@@ -2510,14 +2510,14 @@ def _elemwise_binary(func, self, other, *args, **kwargs):
 
     func_zero = _zero_of_dtype(func(self_zero, other_zero, *args, **kwargs).dtype)
 
-    if isinstance(self, np.ndarray):
+    if not isinstance(self, COO):
         if np.array_equiv(func(self, other_zero, *args, **kwargs), func_zero):
             return _elemwise_binary_self_dense(func, self, other, *args, **kwargs)
         else:
             other.densification_config.check(str(func), self, other)
             return func(self, other.todense(), *args, **kwargs)
 
-    if isinstance(other, np.ndarray):
+    if not isinstance(other, COO):
         if np.array_equiv(func(self_zero, other, *args, **kwargs), func_zero):
             temp_func = _reverse_self_other(func)
             return _elemwise_binary_self_dense(temp_func, other, self, *args, **kwargs)
@@ -2526,8 +2526,12 @@ def _elemwise_binary(func, self, other, *args, **kwargs):
             return func(self.todense(), other, *args, **kwargs)
 
     if func(self_zero, other_zero, *args, **kwargs) != func_zero:
-        raise ValueError("Performing this operation would produce "
-                         "a dense result: %s" % str(func))
+        DensificationConfig.from_many([
+            self.densification_config,
+            other.densification_config
+        ]).check(str(func), self, other)
+
+        return func(self.todense(), other.todense(), *args, **kwargs)
 
     self_shape, other_shape = self.shape, other.shape
     result_shape = _get_broadcast_shape(self_shape, other_shape)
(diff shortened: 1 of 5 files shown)


hameerabbasi:
I think this is ready for code review before I get to the docs. See the tests for example use.

nils-werner:
I find it weird that the `._densification_config` attribute of an output array may changes from being a set `set()` to a `DensificationConfig` instance and back:

    s = sparse.random((20, 30, 40))

    with s.configure_densification(densify=True):
        x = s ** 2
        print(type(s._densification_config))  # DensificationConfig
        print(type(x._densification_config))  # set

    with x.configure_densification(densify=True):
        print(type(s._densification_config))  # DensificationConfig
        print(type(x._densification_config))  # DensificationConfig, WTF?

    print(type(s._densification_config))  # DensificationConfig
    print(type(x._densification_config))  # set, WTF?!

Isn't it possible to make it a `DensificationConfig` instance for every case?

nils-werner:
Also, the `DensificationConfig` instance of each array is not disconnected from the parent once the context manager is left. This means that entering another `configure_densification` context of a parent, all children will be changed as well:

    s = sparse.random((20, 30, 40))

    with s.configure_densification(densify=True):
        x = s ** 2
        print(s._densification_config.densify)           # True
        print(list(x._densification_config)[0].densify)  # True

    print(s._densification_config.densify)               # False
    print(list(x._densification_config)[0].densify)      # False

    with s.configure_densification(densify=True):
        print(s._densification_config.densify)           # True
        print(list(x._densification_config)[0].densify)  # True, this should be False!

    print(s._densification_config.densify)               # False
    print(list(x._densification_config)[0].densify)      # False

hameerabbasi:
Yes, both of these bugs are one and the same. I could, for example, just store a `DensificationConfig` once instead of storing the parents and this would resolve all of these issues.

I have, however, provided a `COO.densification_config` property that will always be the reduced `DensificationConfig`.

However, my question is: Do we actually want the children to be linked to the parent or not? Linked in the sense that once the parent's config changes, the child's does as well.

Another thing I was considering was automatic densification (above a certain density or below a certain size). This would help performance a lot.

hameerabbasi:
Or I could store the children in the context manager and reduce the children's configs once the manager has exited.

sparse/densification.py
```diff
@@ -0,0 +1,140 @@
...
+            self.make_copy_of(densify)
+            return
+
+        if max_size is None:
+            max_size = 10000
+
+        if min_density is None:
+            min_density = 0.25
+
+        if not isinstance(densify, bool) and densify is not None:
```
nils-werner: Instead of allowing `densify` to be of mixed type (`None` and `bool`), you could [create a `Maybe` class that encapsulates the behaviour](https://stackoverflow.com/questions/9501148/booleans-have-two-possible-values-are-there-types-that-have-three-possible-valu)

sparse/densification.py
```diff
@@ -0,0 +1,140 @@
...
+        for manager in managers:
+            densify = _three_way_and(densify, manager.densify)
+            max_size = min(max_size, manager.max_size)
+            min_density = max(min_density, manager.min_density)
+
+        return DensificationConfig(densify, max_size, min_density)
+
+    def __str__(self):
+        if isinstance(self.densify, bool):
+            return '<DensificationConfig: densify=%s>' % self.densify
```
nils-werner: I think there should be one unified `repr`.

sparse/densification.py
```diff
@@ -0,0 +1,140 @@
...
+
+        DensificationConfig._validate_single(configs)
+
+    @staticmethod
+    def _validate_single(config):
+        if not isinstance(config, DensificationConfig):
+            raise ValueError('Invalid DensificationConfig.')
+
+    @staticmethod
+    def combine(*configs):
```
nils-werner: What's the difference between `combine` and `from_many`?

sparse/densification.py
```diff
@@ -0,0 +1,140 @@
...
+
+try:
+    _max_size = sys.maxsize
+except NameError:
+    _max_size = sys.maxint
+
+
+class DensificationConfig(object):
+    def __init__(self, densify=None, max_size=None, min_density=None):
+        if isinstance(densify, (Iterable, DensificationConfig)):
```
nils-werner: I don't like how you abuse the `densify` parameter here. If you want a copy constructor, we should have one specifically for it.

sparse/coo.py
```diff
@@ -2834,20 +2862,21 @@ def _cartesian_product(*arrays):
...
     data_func = func(self.data, *args, **kwargs)
     nonzero = data_func != func_zero
 
+    densification_config = DensificationConfig.combine(self._densification_config)
+
     return COO(self.coords[:, nonzero], data_func[nonzero],
                shape=self.shape,
                has_duplicates=self.has_duplicates,
-               sorted=self.sorted)
+               sorted=self.sorted, densification_config=densification_config)
```
nils-werner: 💄 all other parameters got their own line.

sparse/densification.py
```diff
@@ -0,0 +1,140 @@
...
+            return False
+        else:
+            return self.max_size >= size or self.min_density <= density
+
+    def _raise_if_fails(self, size, density, name):
+        if not self._should_densify(size, density):
+            raise ValueError("Performing this operation would produce "
+                             "a dense result: %s" % name)
+
+    def check(self, name, *arrays):
```
nils-werner: `name` could be made optional. varargs like `*arrays` can become a huge PITA, I think people should pass an iterable instead.

sparse/coo.py
```diff
@@ -2006,7 +2014,39 @@ def astype(self, dtype, out=None):
...
+
+        yield
+
+        if isinstance(old_densification_config, DensificationConfig):
+            self._densification_config.make_copy_of(old_densification_config)
+        else:
+            self._densification_config = old_densification_config
+
+    @property
+    def densification_config(self):
```
nils-werner: I don't like how `COO` knows how to deal with all the different `DensificationConfig` flavors. All of this should be entirely encapsulated in `DensificationConfig`.

hameerabbasi:
Thanks for the great feedback, @nils-werner!

sparse/densification.py
hameerabbasi: Name is needed, we use it for the exception.

sparse/densification.py
hameerabbasi: I feel otherwise... Printing fluff that isn't relevant to the repr can be useless. This just prints the required info.

sparse/coo.py
hameerabbasi: Done.

sparse/densification.py
hameerabbasi: I `combine`d them (pun intended).

Seriously, though. `from_parents` builds the parents, and reduces if none of the parents are in a context manager mode.

`_reduce_from_parents` gets the current rules without regards to what the parents are.

hameerabbasi:
I feel I should be writing tests for the `DensificationConfig` class as well.

sparse/densification.py
hameerabbasi: I'm willing to change this, but it is a huge ease to simply do `config.check(name, arr)` or `config.check(name, arr1, arr2)` instead of putting everything into a list. If we need to pass a list, we can always do `config.check(name, *arrays)`. Could you tell me why it's difficult to support?

mrocklin:
I might be a bit late to this discussion but, are we sure that we want to do this?  Automatic densification may not be a good idea long term.  If possible it would be good to make this change easy to remove in the future without having to touch all of the codebase.  When I see densification commands going into each of the small functions I become concerned.  This also raises the bar for contributions from new developers.

hameerabbasi:
I'll work on that... I'm not entirely sure I need this either, at least for any of my use cases. The best I can do here is pass the parent `COO` objects (instead of `DensificationConfig` objects), and handle it in the constructor. That way, if we need to undo it, all we need to to (on the `COO` side) is to remove the code from the constructor.

hameerabbasi:
Okay, I'm convinced that a densification manager is unwise, and that densification should always be explicit. I'll close this PR but let the branch stay in case someone else can build on my work.

PR #90, opened 2018-01-26
Flatten documentation hierarchy

### LINKED ISSUE #89: Documentation thoughts
Here are some initial thoughts on the documentation.  Feedback welcome.

1.  The user manual section has a lot of useful information like installation and quickstart that it would be useful to have visible from the main page.  I'm inclined to flatten out this subsection and just push everything to top-level
2.  We might want to avoid referring to this library as `sparse` when possible, instead maybe using terms like "this library".  The term `sparse` is fairly generic and awkward (at least to me) given that it is usually an adjective, not a proper noun.  

    Some examples:
    - The section header "Contributing to sparse" might be renamed to just "Contributing"
    - "`sparse` can be obtained from pip via" 

        could be changed to this text 

        "This library can be installed with `pip`:

3.  I wonder if small pages like converting.html could be rolled into some other larger page like operations.html .  That page might also just include all of its subpages on the same html document.  The hierarchy might not be doing us much good here.  We can still have a table of contents for the page, but it might be nice to flatten things out a bit.  This can help people who just want to read linearly through a narrative.

    Not that this is a particularly good example, but consider the [Pandas documentation on groupby](http://pandas.pydata.org/pandas-docs/stable/groupby.html).  It is a long linear page with a detailed table of contents on the left.  People can easily navigate to a particular subsubsection, but that subsubsection doesn't need to live in isolation.

I can do some of this work if desired.  I just wanted to check in with @hameerabbasi to make sure that he is ok with these changes to the documentation.

mrocklin:
Fixes #89


mrocklin:
Some additional things that would be nice to fix (but I'm not sure how):

1.  Flatten the API docs to require fewer clicks to get to a function of interest.  Currently the API page has a single option that you have to click on.  This seems silly, but is perhaps necessary for the machinery we're using now?
2.  Remove scrollbars from code blocks

docs/quickstart.rst
```diff
@@ -0,0 +1,72 @@
...
+   <COO: shape=(100, 100, 100), dtype=float64, nnz=100246, sorted=True, duplicates=False>
+
+For more efficient ways to construct sparse arrays,
+see documentation on :doc:`Constructing Arrays <construct>`.
+
+Compute
+-------
+
+Many of the normal Numpy operations work on :obj:`COO` objects just like on :obj:`numpy.ndarray` objects.
+This includes arithmentic, :obj:`numpy.ufunc` operations, or functions like tensordot and transpose.
```
hameerabbasi: arithmentic -> arithmetic

docs/contribute.rst
```diff
@@ -1,5 +1,5 @@
-Contributing to sparse
-======================
+Contributing
+============
```
hameerabbasi: This probably has a link to the GitHub page that might need to be changed.

docs/index.rst
```diff
@@ -1,4 +1,4 @@
-sparse
+Sparse
```
hameerabbasi: This might also have a GitHub page link.

hameerabbasi:
1 can be fixed by moving `generated/sparse.rst` to `api.rst`, and redoing the headings. 2, I'm not so sure about.

docs/conf.py
```diff
@@ -19,6 +19,7 @@
 #
 import os
 import sys
+import sparse
```
hameerabbasi: This should be moved below the `sys.path.insert`, otherwise it won't work. RTD runs Sphinx from the `docs` directory.

hameerabbasi:
This gets a +1 from me, but you should still do https://github.com/pydata/sparse/pull/90#issuecomment-360849671 if possible.

mrocklin:
Thoughts on how to change the title of the page and TOC line from `sparse` to `API`?  I can change the file, but presumably that file gets auto-generated and so the change would be lost.

hameerabbasi:
I apologize in advance for stating things you may already know here.

Actually, it doesn't. It only gets auto-generated if it's deleted. If it's not deleted, the change sticks. That's a feature so we can modify auto-generated pages.

Also, it only gets auto-generated if there's an `autosummary` directive somewhere upstream. We would be changing the main `autosummary` directive from the modules page to the `sparse` page.

I would suggest the following:

- Delete `api.rst`
- Move `generated/sparse.rst` -> `api.rst`
- Change the heading.
- Add a `toctree` directive of `generated` like the current `api.rst`.
- [Add `:ref:` roles in the flattened document.](http://www.sphinx-doc.org/en/stable/markup/inline.html#ref-role) (see below).
- Modify `generated/sparse.COO.rst` to match the new labels.

You can also do this for inter-links:
```
.. _my-label:

Some section heading
----------------------


# Somewhere else, maybe even in another file entirely!
:ref:`my-label` # Automatic title for link text!
:ref:`A link <my-label>` # Custom link text
```

You may need to delete your build directory, otherwise it doesn't fully update the TOC on every page, and you will only see a fraction of the warnings (such as the ones that should have happened for missing cross-links when you changed the documentation structure). RTD does this for us.

If this is all too much for you, I can take over and push a commit or two. If not, it's a great learning experience!

hameerabbasi:
If you want to put the `COO` and `DOK` classes on the main API page, copy over their contents and get rid of the classes part in `sparse.rst`

hameerabbasi:
Alternatively, you could just not move them and rename `sparse.rst` to `api.rst` and do the changes there. In this case, you don't need a `:toctree:` directive.

mrocklin:
Rebased on master.  Merging this soon if there are no objections.

hameerabbasi:
Minor comment... The original `api.rst` should be deleted.

hameerabbasi:
And there's still the problem of missing/broken cross references in `generated/sparse.COO.rst`. :-) I can do that if you have issues.

mrocklin:
> And there's still the problem of missing/broken cross references in generated/sparse.COO.rst. :-) I can do that if you have issues.

If you wanted to handle that that would be welcome.  It might also be useful to have the autosummary references come before the autoclass and large docstring.  That way people can more easily see the options they have on this page.

hameerabbasi:
True. I'm a bit busy today, I might not be responsive. I'll handle it tomorrow, if you're okay with that.

mrocklin:
No rush on my end.  I wanted to get this in to avoid having to rebase further and handle conflicts with other PRs that touch documentation.  But if you're busy then those are unlikely to arise anyway :)

PR #92, opened 2018-01-27
Inheritance hierarchy

hameerabbasi:
- Move from `six` to `future` (I think it's cleaner, with a few imports we can write Python 2/3 compatible code instead of modifying everything)
- Add a `SparseArray` class with the properties common to both `COO` and `DOK`.
- Make `COO` and `DOK` inherit from `SparseArray`.
- Update docs.


sparse/coo.py
```diff
@@ -270,6 +263,29 @@ def __init__(self, coords, data=None, shape=None, has_duplicates=True,
...
+        >>> x = np.zeros((10, 10))
+        >>> s = COO.from_numpy(x)
+        >>> s.shape
+        (10, 10)
+        """
+        return self._shape
+
+    @shape.setter
+    def shape(self, value):
+        self._shape = value
```
mrocklin: Perhaps we should avoid the property altogether and just have this be a normal attribute?

hameerabbasi: There's no way to inherit it if we do that. And we need `shape` directly for `ndim`, `size`,  and indirectly for`density`.

mrocklin: Maybe we just avoid listing it as an abstract method then?

hameerabbasi: Then there's no point to the base class. We would have to remove all of the properties I mentioned.

The whole point of the class is to access the "common" stuff for all types.

hameerabbasi: Or we can make it have an `__init__` method with just the shape.

mrocklin: This seems like it might work?  The super class doesn't necessarily need to be functional on its own.

```python
class SparseArray(object):
    @property
    def ndim(self):
        return len(self.shape)

    @property
    def size(self):
        return reduce(operator.mul, self.shape, 1)
  
    @property
    def density(self):
        return self.nnz / self.size

class COO(SparseArray):
    def __init__(...):
        self.shape = ...
```

hameerabbasi: It doesn't make sense in an structural way to be using things undefined in the parent class. However, I made it into an attribute of the parent class. This keeps it as an attribute and still makes sense in the inheritance hierarchy.

sparse/dok.py
```diff
@@ -99,46 +89,45 @@ class DOK(object):
...
 
-            for c, d in six.iteritems(data):
+            for c, d in data.items():
                 self[c] = d
         else:
             raise ValueError('data must be a dict.')
 
+    @property
+    def dtype(self):
+        return self._dtype
```
mrocklin: Thoughts on using an attribute rather than a property here?  That seems simpler.

My guess is that we're doing this because the superclass has a property.  In this case I would be inclined to just drop the property on the superclass.

sparse/dok.py
```diff
@@ -377,7 +343,7 @@ def todense(self):
         """
         result = np.zeros(self.shape, dtype=self.dtype)
 
-        for c, d in six.iteritems(self.data):
+        for c, d in self.data.items():
```
mrocklin: I'm +1 on this change, even if it causes more memory issues in Python 2.

hameerabbasi: With `from builtins import dict` it won't, unless the user purposefully passes an old-style array. :-)

sparse/dok.py
hameerabbasi: `COO` uses a property, whereas `DOK` uses an attribute. We could make `COO` use an attribute as well, however that seems bad since it might not match `data.dtype`. I chose the lowest common denominator and went for a property.

mrocklin: Is it important that they match?

sparse/dok.py
```diff
@@ -1,31 +1,21 @@
-import six
+from builtins import dict, int
```
mrocklin: What is `dict` here in Python 2?  It's not a normal builtin dict object?

Ah, ok, I'm now reading about the `future` package.  I have to admit that I am a bit hesitant to add it as a new dependency just because I haven't heard of it before.  Happy to deal with this in the future though.

hameerabbasi: `future` is a dependency for many packages. It's more common than `six`, I believe. I may be wrong.

It's `pip` installable.

mrocklin: > future is a dependency for many packages. It's more common than six, I believe. I may be wrong.

This would make me much more comfortable.  I would be somewhat surprised to learn this though.

I hesitate to have a `dict` type anywhere in the code that is not the standard builtin dict type, both for performance and (more importantly) consistency reasons with the ecosystem.

sparse/dok.py
mrocklin: > Anything that is common should be in the base class

I would agree with this to the extent that it's convenient.  It looks like by trying to make these two subclasses adhere to the same interface we're forced to increase the level of technology used by one of the subclasses.  I have a (probably irrational) aversion to using more advanced features than are necessary, mostly on the grounds of avoiding complex situations and increasing maintainability.

sparse/dok.py
hameerabbasi: I just took a peek through the code for `future`.

```
    def items(self):
        """
        On Python 2.7+:
            D.items() -> a set-like object providing a view on D's items
        On Python 2.6:
            D.items() -> an iterator over D's items
        """
        if ver == (2, 7):
            return self.viewitems()
        elif ver == (2, 6):
            return self.iteritems()
        elif ver >= (3, 0):
            return self.items()
```

mrocklin: What if some other infrastrucural library (like cloudpickle, Cython, PyPy, Numba, ...) has an optimiziation that works well on objects where `type(obj) is dict` (this is almost certainly the case in all of those projects).  Then presumably that optimization won't apply to the DOK class.

This library is aimed to be infrastructural for the ecosystem.  I think that it is important that we be as non-fancy as possible.  As another point, if you ever want to see this upstreamed into `scipy` then you'll need to convince them about this point as well and they are, in general, far more conservative than I am.

sparse/dok.py
hameerabbasi: If I were to do something like:
```
assert isinstance(x, SparseArray)
```

Type inference for any IDE would allow me to automatically see `dtype`, which I feel is kind of important.

Although you may be right in thinking it incurs a slight performance penalty, or incurs a small maintenance cost, this would make things easier in the future, providing a checklist of sorts for what any inherited sparse array should provide.

sparse/dok.py
mrocklin: I had this blogpost sitting on my hard drive as a draft; I just pushed it to my website.  

http://matthewrocklin.com/blog/work/2018/01/27/write-dumb-code

It has my thoughts on the topic of avoiding technology in programming whenever possible.  I may be a bit dogmatic on this topic, I apologize if this causes friction.

hameerabbasi: `builtins` basically aims to map Python 3 behavior everywhere. It also has a `past` module for mapping Python 2 behavior anywhere.

sparse/dok.py
mrocklin: I'll admit that I don't use IDEs as much, and so don't fully understand what needs to happen to support nice linting and recommendations.  I do see the value in this though.  Thank you for bringing it up.  

Are there other ways to say that objects of type SparseArray have certain attributes other than to create properties?  For example if we were to include a class-level attribute?

```python
class SparseArray(object):
    dtype = None
```

sparse/dok.py
mrocklin: Yeah, it's a nice trick, don't get me wrong.  I've just seen many many things go wrong once code like this interacts with 1000 other libraries (which again, I think is our intention).  I would definitely use a library like this in application code, but for core ecosystem code I tend to be more conservative.  I think that many other maintainers of core ecosystem code also hold this belief.

sparse/dok.py
mrocklin: Or can type annotations get us out of this problem?

hameerabbasi: Although that would work for type inference, inheriting from `SparseArray` won't fail if we forget to implement `dtype`. Failures are good if we fail to add basic functionality.

In addition, `SparseArray.dtype` would be inferred as `NoneType` in all IDEs instead of a `np.dtype` object.

mrocklin: > Although that would work for type inference, inheriting from SparseArray won't fail if we forget to implement dtype. Failures are good if we fail to add basic functionality.

I'm not particularly concerned about us forgetting to specify a dtype.

> In addition, SparseArray.dtype would be inferred as NoneType in all IDEs instead of a np.dtype object.

What happens now?  We don't specify the dtype in the property.

hameerabbasi: We could, for example, add a docstring and it'd be inferred as a dtype.

sparse/dok.py
hameerabbasi: Hmm. I'd be okay going back to `six`, but the memory constrains could get bad. What do you think?

hameerabbasi: > I apologize if this causes friction.

Not at all. :-) We're just discussing.

mrocklin: I would wait for someone to arrive with a memory issue before optimizing for memory constraints in this way.  I don't expect this to be a significant issue in common use.  Also, I don't mind Python 2 being less efficient.

sparse/dok.py
mrocklin: I asked about the right way to do this on stack overflow:

https://stackoverflow.com/questions/48481041/build-type-information-about-attributes-into-superclass-for-type-hinting/48481359

hameerabbasi: Great idea. However, I had another: How about A class docstring with an Attributes section?

sparse/dok.py
hameerabbasi: I removed the `six` dependency entirely and added our own basic Python 2/3 compat.

hameerabbasi:
If there are no other objections, I think this gets a +1 from me.

sparse/dok.py
mrocklin: Glad to hear it!

mrocklin:
I've renamed py23.py to compatibility.py and pushed to this branch (I hope that's ok).  +1 from me to merge

hameerabbasi:
Thanks!

PR #97, opened 2018-01-29
Merge README content into the main index doc page

mrocklin:
This adds some informative content to the documentation
and centralizes our prose in one place.


hameerabbasi:
This seems fine to me. I'm sorry to piggy back on this issue, but I didn't want to open another one, what about the PyData domain?

hameerabbasi:
On second thought, could we add developer information to the GitHub `README`? Kind of like [SciPy](https://github.com/scipy/scipy/blob/master/README.rst).

Just two links would be enough, with some very short dev intro, IMO:

- Contributing
- Docs

Also, can we add a section on filing issues in  contributing?

mrocklin:
This seems like a good idea to me.  I may not get to it until later today
if you'd like to take a shot at it before then.

hameerabbasi:
@mrocklin, I could use your input here. :-)

docs/operations.rst
```diff
@@ -9,7 +9,7 @@ Operators
 ---------
 
 :obj:`COO` objects support a number of operations. They interact with scalars,
-:obj:`numpy.ndarray` objects, other :obj:`COO` objects, and
+:doc:`Numpy arrays <reference/generated/numpy.ndarray>`, other :obj:`COO` objects, and
```
mrocklin: I'm curious, why was this change required?

mrocklin:
In general this seems fine to me.  I'm in favor of merging.

docs/operations.rst
hameerabbasi: Saying "numpy dot enn dee array objects" is a mouthful. I thought it would be easier for readers to just change it to "Numpy arrays."

Issue #1, opened 2017-04-17
Support Everything that XArray Expects


mrocklin:
https://github.com/bolt-project/bolt/issues/58

- [ ]  single argument ufuncs (sin, exp, etc.) and ufunc like functions (pd.isnull, notnull, astype, around, isclose)
- [ ]  broadcasting binary operations (e.g., ndarray arithmetic)
- [ ]  three argument version of where (preferably with broadcasting)
- [ ]  aggregation methods (argmax/argmin/max/min/prod/std/var)
- [ ]  nan-skipping aggregations (e.g., nanmean, nansum)
- [ ]  indexing with integer arrays, booleans, slices, None
- [ ]  transpose
- [ ]  insert or setting with fancy indexing
- [ ]  broadcast_to (NumPy 1.10)
- [ ]  concatenate and stack (NumPy 1.10)

shoyer:
I updated the check-list (apparently I'm an Owner for pydata) to drop "insert/fancy indexing" and add a three-item checklist for indexing instead.

hameerabbasi:
Do we have partial XArray compatibility or does XArray fail with `sparse` arrays outright?

shoyer:
@hameerabbasi Not yet, see https://github.com/pydata/xarray/issues/1375. But `sparse` is close enough that it might be worth trying soon.

From a usability perspective, the most helpful additional feature would be nan-skipping aggregations. I know `sparse` will probably rarely be used with NaNs, but xarray's aggregation methods default to `skipna=True` so this would be valuable for consistency.

In addition to the direct uses, xarray uses `where` and outer/vectorized indexing internally primarily for alignment with `join='outer'` (e.g., in `concat` or `merge`). This sort of alignment would require making sparse arrays dense, unless they have a fill value of NaN rather than 0. So these operations are less essential than they might otherwise seem.

hameerabbasi:
I just tested with your example code, NaN-skipping aggregations use `ufunc`s with `reduce`, so they should already be supported in theory.
```python
>>> class A(np.ndarray):
...     def __array_ufunc__(self, *args, **kwargs):
...         print('ufunc', args, kwargs)
...         
>>> np.nansum(np.arange(5).view(A))
ufunc (<ufunc 'add'>, 'reduce', A([0, 1, 2, 3, 4])) {'axis': None, 'dtype': None, 'keepdims': False}
```

Although I have absolutely now idea how to discern that this was NaN skipping from the args/kwargs.

[...]

PR #98, opened 2018-02-10
Add N-ary broadcasting operations.

hameerabbasi:
This PR adds N-ary broadcasting operations (in preparation for `where`) and simplifies code for the N-ary case.

Discussed in https://github.com/pydata/sparse/issues/1

### DIFF
--- a/sparse/coo.py
+++ b/sparse/coo.py
@@ -243,7 +243,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):
 
         super(COO, self).__init__(shape)
         if self.shape:
-            dtype = np.min_scalar_type(max(self.shape))
+            dtype = np.min_scalar_type(max(max(self.shape) - 1, 0))
         else:
             dtype = np.uint8
         self.coords = self.coords.astype(dtype)
@@ -1464,6 +1464,9 @@ class COO(SparseArray, NDArrayOperatorsMixin):
         --------
         :obj:`numpy.broadcast_to` : NumPy equivalent function
         """
+        if shape == self.shape:
+            return self
+
         result_shape = _get_broadcast_shape(self.shape, shape, is_result=True)
         params = _get_broadcast_parameters(self.shape, result_shape)
         coords, data = _get_expanded_coords_data(self.coords, self.data, params, result_shape)
@@ -2017,8 +2020,8 @@ def elemwise(func, *args, **kwargs):
     """
     # Because we need to mutate args.
     args = list(args)
-
-    scalars = 0
+    posargs = []
+    pos = []
     for i in range(len(args)):
         if isinstance(args[i], scipy.sparse.spmatrix):
             args[i] = COO.from_scipy_sparse(args[i])
@@ -2031,12 +2034,16 @@ def elemwise(func, *args, **kwargs):
 
             # The -scalars factor is there because we need to account for already
             # added scalars in the function.
-            func = _posarg_partial(func, i - scalars, args[i])
-            scalars += 1
+            pos.append(i)
+            posargs.append(args[i])
         elif isinstance(args[i], SparseArray) and not isinstance(args[i], COO):
             args[i] = COO(args[i])
+        elif not isinstance(args[i], COO):
+            raise ValueError("Performing this operation would produce "
+                             "a dense result: %s" % str(func))
 
     # Filter out scalars as they are 'baked' into the function.
+    func = _posarg_partial(func, pos, posargs)
     args = list(filter(lambda arg: not isscalar(arg), args))
 
     if len(args) == 0:
@@ -2071,8 +2078,6 @@ def _elemwise_n_ary(func, *args, **kwargs):
     ValueError
         If the input shapes aren't compatible or the result will be dense.
     """
-    check = kwargs.pop('check', True)
-
     args = list(args)
 
     for arg in args:
@@ -2083,70 +2088,22 @@ def _elemwise_n_ary(func, *args, **kwargs):
 
     func_value = func(*args_zeros, **kwargs)
     func_zero = _zero_of_dtype(np.dtype(func_value))
-    if check and func_value != func_zero:
+    if func_value != func_zero:
         raise ValueError("Performing this operation would produce "
                          "a dense result: %s" % str(func))
-
-    # zero_partials = [_posarg_partial(func, i, args_zeros[i])
-    #                  for i in range(len(args))]
-    #
-    # partial_args = [args[:i] + args[i + 1:] for i in range(len(args))]
-    # results_with_zero = [zero_partials[i](*partial_args[i], **kwargs)
-    #                      for i in range(len(args))]
-    # results_with_zero = [x if isinstance(x, COO) else COO.from_numpy(x)
-    #                      for x in results_with_zero]
-
-    for i in range(len(args)):
-        if not isinstance(args[i], COO):
-            zero_partial = _posarg_partial(func, i, args_zeros[i])
-            partial_args = args[:i] + args[i + 1:]
-            result_with_zero = zero_partial(*partial_args, **kwargs)
-            data_with_zero = result_with_zero.data if isinstance(result_with_zero, COO) \
-                else result_with_zero
-            if np.array_equiv(data_with_zero, func_zero):
-                args[i] = COO.from_numpy(args[i])
-            else:
-                raise ValueError("Performing this operation would produce "
-                                 "a dense result: %s" % str(func))
-
-            del zero_partial
-            del partial_args
-            del result_with_zero
-            del data_with_zero
-
-    # matched_coords, matched_data, result_shape = _match_coo(*args)[1:]
-    #
-    # # Start with an empty list. This may reduce computation in many cases.
     data_list = []
     coords_list = []
-    #
-    # data_list.append(func(*matched_data, **kwargs))
-    # coords_list.append(matched_coords)
 
     for mask in product([True, False], repeat=len(args)):
         if not any(mask):
             continue
 
-        di, ci = _unmatch_coo(func, args, mask, **kwargs)
+        ci, di = _unmatch_coo(func, args, mask, **kwargs)
 
-        data_list.extend(di)
         coords_list.extend(ci)
+        data_list.extend(di)
 
-    # for i in range(len(args)):
-    #     zero_func = results_with_zero[i]
-    #
-    #     if len(zero_func.data):
-    #         # Get just the matched indices for the excluded index.
-    #         zero_matches = _match_coo(zero_func, args[i])[0][0]
-    #
-    #         unmatched_coords, unmatched_func = \
-    #             _get_unmatched_coords_data(zero_func.coords,
-    #                                        zero_func.data,
-    #                                        zero_func.shape, result_shape,
-    #                                        zero_matches, matched_coords)
-    #
-    #         data_list.extend(unmatched_func)
-    #         coords_list.extend(unmatched_coords)
+    result_shape = _get_nary_broadcast_shape(*[arg.shape for arg in args])
 
     # Concatenate matches and mismatches
     data = np.concatenate(data_list) if len(data_list) else np.empty((0,), dtype=args[0].dtype)
@@ -2178,6 +2135,9 @@ def _match_coo(*args):
     matched_data : list[ndarray]
         The matched data.
     """
+    if len(args) == 1:
+        return [np.arange(args[0].nnz)], [args[0]]
+
     shapes = [arg.shape for arg in args]
     matched_shape = _get_nary_broadcast_shape(*shapes)
     broadcast_params = [_get_broadcast_parameters(shape, matched_shape)
@@ -2186,11 +2146,9 @@ def _match_coo(*args):
     combined_params = [all(params) for params in zip(*broadcast_params)]
 
     reduced_coords = [_get_reduced_coords(arg.coords,
-                                          arg.shape,
                                           combined_params[-arg.ndim:])
                       for arg in args]
-    reduced_shape = _get_reduced_shape(args[0].coords,
-                                       args[0].shape,
+    reduced_shape = _get_reduced_shape(args[0].shape,
                                        combined_params[-args[0].ndim:])
 
     reduced_linear = [_linear_loc(coord, reduced_shape)
@@ -2208,12 +2166,14 @@ def _match_coo(*args):
     matched_coords = [c[:, idx] for c, idx in zip(coords, matched_indices)]
 
     # Add the matched part.
-    matched_coords = _get_nary_matching_coords(matched_coords, broadcast_params)
+    matched_coords = _get_nary_matching_coords(matched_coords, broadcast_params, matched_shape)
     matched_data = [d[idx] for d, idx in zip(data, matched_indices)]
 
     matched_indices = [sidx[midx] for sidx, midx in zip(sorted_indices, matched_indices)]
 
-    return matched_indices, matched_coords, matched_data, matched_shape
+    matched_arrays = [COO(matched_coords, d, shape=matched_shape) for d in matched_data]
+
+    return matched_indices, matched_arrays
 
 
 def _unmatch_coo(func, args, mask, **kwargs):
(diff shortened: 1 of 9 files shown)


hameerabbasi:
cc @shoyer Your input would be valuable, if you have time.
cc @mrocklin Your input is welcome, as always.

sparse/coo.py
```diff
@@ -2149,101 +2117,153 @@ def _elemwise_binary(func, self, other, *args, **kwargs):
...
-    assert isinstance(other, COO)
+def _match_coo(*args):
+    """
+    Matches the coordinates for any number of input :obj:`COO` arrays.
+    Equivalent to "sparse" broadcasting for all arrays.
 
-    result_shape = _get_broadcast_shape(self.shape, other.shape)
+    Parameters
+    ----------
+    args : tuple[COO]
```
mrocklin: If you're trying for parametrized python type annotations then I think it's supposed to be standard to use capitalized types like `List[COO]` or `Tuple[np.ndarray]`

mrocklin: I don't know though, this is somewhat new to me.

sparse/coo.py
```diff
@@ -2034,113 +1988,127 @@ def _grouped_reduce(x, groups, method, **kwargs):
...
-        _get_reduced_coords(other_coords, other_shape,
-                            other_reduce_params)
-    other_reduced_linear = _linear_loc(other_reduced_coords, other_reduced_shape)
-    i = np.argsort(other_reduced_linear)
-    other_reduced_linear = other_reduced_linear[i]
-    other_coords = other_coords[:, i]
-    other_data = other_data[i]
+    # Filter out scalars as they are 'baked' into the function.
+    func = _posarg_partial(func, pos, posargs)
+    args = list(filter(lambda arg: not isscalar(arg), args))
```
mrocklin: You might consider `toolz.remove` here

sparse/coo.py
```diff
@@ -2034,113 +1988,127 @@ def _grouped_reduce(x, groups, method, **kwargs):
...
+    See Also
+    --------
+    :obj:`numpy.ufunc` : A similar Numpy construct. Note that any :code:`ufunc` can be used
+        as the :code:`func` input to this function.
+    """
+    # Because we need to mutate args.
+    args = list(args)
+    posargs = []
+    pos = []
+    for i in range(len(args)):
```
mrocklin: You might consider `for i, arg in enumerate(args)`, which might be a bit more idiomatic for Python readers

sparse/coo.py
```diff
@@ -1954,6 +1894,20 @@ def tril(x, k=0):
     return COO(coords, data, x.shape, x.has_duplicates, x.sorted)
 
 
+def _nary_match(*arrays):
```
mrocklin: I'm not able to quickly figure out what this function does.  Can I ask you for a small docstring?  If possible I find small example sections in docstrings to be very helpful when learning codebases that others have written.

sparse/coo.py
```diff
@@ -2034,113 +1988,127 @@ def _grouped_reduce(x, groups, method, **kwargs):
...
+    for mask in product([True, False], repeat=len(args)):
+        if not any(mask):
+            continue
 
-        coords_list.extend(other_unmatched_coords)
-        data_list.extend(other_unmatched_func)
+        ci, di = _unmatch_coo(func, args, mask, **kwargs)
+
+        coords_list.extend(ci)
+        data_list.extend(di)
```
mrocklin: This confuses me and seems concerning.  I see that this was a main point of your conversation with @shoyer earlier.  I probably have some thinking to do on this problem before I'm able to reasonably comment on this.

mrocklin:
I think that it would be good to see a more comprehensive test suite that fully explains the complexity of what we're trying to accomplish here.  I think that that will make it more clear as we discuss different possibilities here.  We might ask "why are we doing X" and the answer can be "see test_X".  I get the sense that you've thought deeply about this problem and know all of the problems that might arise.  It would be very valuable to encode that deep thinking and all of those corner cases into a test suite.

sparse/coo.py
```diff
@@ -2511,8 +2512,39 @@ def _linear_loc(coords, shape, signed=False):
...
+
+        def __str__(self):
+            return str(func)
+
+        def __repr__(self):
+            return repr(func)
+
+        __doc__ = func.__doc__
+
+    return Partial()
```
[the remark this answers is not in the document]
hameerabbasi: I could have, but our situation is slightly unique:

1. We're using `str(func)` for exceptions. `functools.wraps` doesn't work on that for all callables (e.g. `ufunc`s), and breaks a few docstrings. This leads to illegible names in exceptions like `_posarg_partial.<locals>.wrapper` (and the same for debugging).
2. We're replacing a number of arguments in different positions.

I guess I could turn it into a class rather than a decorator style function.

sparse/coo.py
hameerabbasi: `tuple` and `list` tend to work better with intersphinx and code type annotations, so I tend to prefer those. Of course, I could import in something, but then that gives me PEP8 failures as I don't use it in code, just in docstrings.

sparse/coo.py
hameerabbasi: I'd prefer not to introduce a dependency for something as simple as this.

hameerabbasi:
I plan to make more comprehensive tests, yes. But the issue is some of the complexity can't be directly tested: For example, the optimizations are just that: Optimizations. We can design the tests so the optimizations are hit but we can't know that they kicked in without weird monkey-patching of some sort.

sparse/coo.py
hameerabbasi: Whoops, must have missed that one.

sparse/coo.py
hameerabbasi: Done!

sparse/coo.py
```diff
@@ -2426,80 +2468,39 @@ def _elemwise_unary(func, self, *args, **kwargs):
...
+    coords : list[numpy.ndarray]
+        The input coordinates.
+    params : list[Union[bool, none]]
+        The broadcast parameters.
     Returns
     -------
-    matching_coords : np.ndarray
-        The coordinates of the output array for which both inputs will be nonzero.
+    numpy.ndarray
+        The broacasted coordinates.
```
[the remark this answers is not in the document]
hameerabbasi: Done!

sparse/coo.py
```diff
@@ -2426,80 +2468,39 @@ def _elemwise_unary(func, self, *args, **kwargs):
                sorted=self.sorted)
 
 
-def _get_matching_coords(coords1, coords2, shape1, shape2):
+def _get_nary_matching_coords(coords, params, shape):
```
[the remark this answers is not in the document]
hameerabbasi: Done!

sparse/coo.py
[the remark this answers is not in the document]
hameerabbasi: I turned it into a callable class.

hameerabbasi:
It seems there's a slight bug for number of inputs >2 and broadcasting, nothing unfixable, but will have to think a bit. I'm on it.

mrocklin:
> I plan to make more comprehensive tests, yes. But the issue is some of the complexity can't be directly tested: For example, the optimizations are just that: Optimizations. We can design the tests so the optimizations are hit but we can't know that they kicked in without weird monkey-patching of some sort.

I think that there are probably a lot of correctness tests that could be written as well.  In #1 you discussed many situations that might arise for which a system like this would be necessary to catch.  Ideally we would encode all of those situations as tests to ensure that future developers don't change code to alter correct behavior here.

sparse/coo.py
```diff
@@ -2034,113 +2006,124 @@ def _grouped_reduce(x, groups, method, **kwargs):
...
-        _get_reduced_coords(other_coords, other_shape,
-                            other_reduce_params)
-    other_reduced_linear = _linear_loc(other_reduced_coords, other_reduced_shape)
-    i = np.argsort(other_reduced_linear)
-    other_reduced_linear = other_reduced_linear[i]
-    other_coords = other_coords[:, i]
-    other_data = other_data[i]
+    # Filter out scalars as they are 'baked' into the function.
+    func = PositinalArgumentPartial(func, pos, posargs)
+    args = list(filter(lambda arg: not isscalar(arg), args))
```
shoyer: optional: consider using a list comprehension instead

sparse/coo.py
```diff
@@ -2149,101 +2132,153 @@ def _elemwise_binary(func, self, other, *args, **kwargs):
...
-    shape : tuple[int]
-        The shape corresponding to these coordinates.
-    result_shape : tuple[int]
-        The result broadcasting shape.
-    matched_idx : np.ndarray
-        The indices into the coords array where it matches with the other array.
-    matched_coords : np.ndarray
-        The overall coordinates that match from both arrays.
+    args : tuple[COO]
+        The input :obj:`COO` arrays.
```
shoyer: add in `func`, `mask` and `**kwargs` to the docstring?

sparse/coo.py
```diff
@@ -2149,101 +2132,153 @@ def _elemwise_binary(func, self, other, *args, **kwargs):
...
+    matched_args = [a for a, m in zip(args, mask) if m]
+    unmatched_args = [a for a, m in zip(args, mask) if not m]
 
-    unmatched &= nonzero
-    matched &= nonzero
+    matched_arrays = _match_coo(*matched_args)[1]
 
-    coords_list = []
-    data_list = []
+    pos, = np.where([not m for m in mask])
```
shoyer: maybe use `np.flatnonzero()`?

shoyer:
I agree with @mrocklin that a more extensive test suite is vital here. This logic is complicated and fixing bugs later will be hard. I haven't seriously tried to follow it yet.

I would suggest parametric tests verifying proper broadcasting with 2 or 3 arguments with:
- `*` for sparse broadcasting
- `+` for dense broadcasting
- an order dependent function (e.g., `-`) to verify you get order right
- with pathological values `np.infty` and `np.nan`

sparse/coo.py
hameerabbasi: Done!

sparse/coo.py
hameerabbasi: Done!

sparse/coo.py
hameerabbasi: This isn't really a numerical operation. I've converted it to a `tuple(generator comprehension)` form and avoided `np.where` altogether. The exact code is
```
pos = tuple(i for i, m in enumerate(mask) if not m)
```

hameerabbasi:
I think this PR is now ready for a comprehensive review + merge.

sparse/coo.py
mrocklin: It looks like this function is no longer used.  Delete?

sparse/coo.py
```diff
@@ -2034,113 +2006,125 @@ def _grouped_reduce(x, groups, method, **kwargs):
...
+        elif isscalar(arg) or (isinstance(arg, np.ndarray)
+                               and not arg.shape):
+            # Faster and more reliable to pass ()-shaped ndarrays as scalars.
+            if isinstance(arg, np.ndarray):
+                args[i] = arg[()]
+
+            pos.append(i)
+            posargs.append(args[i])
+        elif isinstance(arg, SparseArray) and not isinstance(arg, COO):
+            args[i] = COO(arg)
```
mrocklin: This line doesn't get hit by tests.  Should we add a small DOK test?

sparse/coo.py
```diff
@@ -2034,113 +2006,125 @@ def _grouped_reduce(x, groups, method, **kwargs):
...
+        elif not isinstance(arg, COO):
             raise ValueError("Performing this operation would produce "
                              "a dense result: %s" % str(func))
```
mrocklin: Same here.  No test triggers this error-handling code.

sparse/coo.py
```diff
@@ -2034,113 +2006,125 @@ def _grouped_reduce(x, groups, method, **kwargs):
...
-
-    # Start with an empty list. This may reduce computation in many cases.
-    data_list = []
-    coords_list = []
+    # Filter out scalars as they are 'baked' into the function.
+    func = PositinalArgumentPartial(func, pos, posargs)
+    args = [arg for arg in args if not isscalar(arg)]
+
+    if len(args) == 0:
+        return func(**kwargs)
```
mrocklin: Also here.  No test operates on no args

sparse/coo.py
```diff
@@ -2149,101 +2133,173 @@ def _elemwise_binary(func, self, other, *args, **kwargs):
...
+        ``return_midx`` is ``False``.
+    """
+    return_midx = kwargs.pop('return_midx', False)
+    cache = kwargs.pop('cache', None)
+
+    if kwargs:
+        raise ValueError('Unknown kwargs %s' % kwargs.keys())
+
+    if return_midx and (len(args) != 2 or cache is not None):
+        raise NotImplementedError('Matching only supported for two args, and no cache.')
```
mrocklin: Do we still need this option?

sparse/tests/test_coo.py
```diff
@@ -234,6 +234,113 @@ def test_elemwise_binary(func, shape):
...
+     ], lambda x, y, z: x - y + z),
+])
+def test_nary_broadcasting(shapes, func):
+    args = [sparse.random(s, density=0.5) for s in shapes]
+    dense_args = [arg.todense() for arg in args]
+
+    fs = sparse.elemwise(func, *args)
+    assert isinstance(fs, COO)
+
+    assert_eq(fs, func(*dense_args))
```
mrocklin: It would be nice to test and verify that we are not creating unnecessary zeroes in the data attribute.  We might either test that explicitly here, or we might put it into `assert_eq`.  I've gone ahead and pushed a commit to your branch that adds a check into `assert_eq`.  Please remove if you prefer not to add this here.

sparse/tests/test_coo.py
```diff
@@ -234,6 +234,113 @@ def test_elemwise_binary(func, shape):
...
+    np.nan,
+    np.inf,
+    -np.inf
+])
+@pytest.mark.filterwarnings('ignore:invalid value')
+def test_nary_broadcasting_pathological(shapes, func, value):
+    def value_array(n):
+        ar = np.empty((n,), dtype=np.float_)
+        ar[:] = value
+        return ar
```
mrocklin: We might want just a few of the values to be pathological instead of all of them.

sparse/coo.py
hameerabbasi: No, we don't. I'm not omniscient, so I went ahead and added this check in case someone tried to trigger caching on `return_midx` (which we don't cache, it's never repeated); or tried to match indices for `len(args) != 2` (I'm not sure if we'll need this in the future, but we might, and it's useful to err rather than have it return incorrect results).

sparse/tests/test_coo.py
hameerabbasi: I'd like to verify we don't create additional zeros for all our operations, so that seems like a rather useful addition.

hameerabbasi: Although I would prefer to use `np.count_nonzero`.

sparse/tests/test_coo.py
hameerabbasi: I'll modify the test to match that.

sparse/coo.py
hameerabbasi: Added!

sparse/coo.py
hameerabbasi: Added a small test that hits this.

sparse/coo.py
hameerabbasi: Added another small test for this.

sparse/tests/test_coo.py
hameerabbasi: Modified.

hameerabbasi:
I've incorporated more or less all of your suggestions about coverage, with one exception (see comments!)

mrocklin:
@shoyer do you have a chance to look at this?  "Nope" is a fine answer.

hameerabbasi:
I'm guessing @shoyer doesn't work weekends. :-) If there's no reply or a "Nope" by the end of Monday, we can decide what to do next.

shoyer:
It's a mixed bag on weekend, but this weekend my wife is away so I have time for open source :).

I'll take a look.

sparse/tests/test_coo.py
```diff
@@ -250,7 +250,128 @@ def test_elemwise_trinary(func, shape):
...
+
+    assert_eq(fs, func(x, y, z))
+
+
+@pytest.mark.parametrize('shapes, func', [
+    ([
+         (2,),
+         (3, 2),
+         (4, 3, 2),
+     ], lambda x, y, z: (x + y) * z),
```
shoyer: Consider doing a full cross-product of shapes and functions here.

sparse/tests/test_coo.py
```diff
@@ -250,7 +250,128 @@ def test_elemwise_trinary(func, shape):
...
+         (4,),
+         (4, 4),
+         (4, 4, 4),
+     ], lambda x, y, z: x + y - z),
+    ([
+         (4,),
+         (4, 4),
+         (4, 4, 4),
+     ], lambda x, y, z: x - y + z),
+])
```
shoyer: It would be good to add checks for a few more variations on the broadcasting logic to exercise the matching logic:
- Dimensions of size 1, e.g., `(3, 1)` + `(3, 4)` -> `(3, 4)`
- Output shapes that don't match one of the inputs, e.g., `(3, 1)` + `(1, 4)` -> `(3, 4)`.
- Outputs that require matching across three inputs, e.g., `(1, 1, 2)` + `(1, 3, 1)` + `(4, 1, 1)` -> `(4, 3, 2)`.

sparse/tests/test_coo.py
shoyer: On a related note, maybe it would be useful to add a test that verifies "sparse" broadcasting is actually done in a sparse way?

I think you could do this by mocking the underlying functions (e.g., `np.mul`) and then verifying that the calls match expectations.

sparse/tests/test_coo.py
hameerabbasi: The first two were already covered in `test_broadcasting`. I renamed that to `test_binary_broadcasting` and moved it closer to these.

The third, I also added.

sparse/tests/test_coo.py
hameerabbasi: Done!

hameerabbasi:
I can't seem to be able to respond to your "sparse" broadcasting comment, so I'm responding here.

I monkey-patched one of our own functions and verified the behavior is correct there. I also verified `operator.add` takes the 'dense' path.

mrocklin:
> I can't seem to be able to respond to your "sparse" broadcasting comment, so I'm responding here.

Yeah, I was trying to do that as well.  I haven't seen that before

Is checking for the right number of non-zeros in the output not sufficient?  Do we have code paths that would re-sparsify a dense intermediate result?

hameerabbasi:
We're actually talking about the "optimized" code path for things like `operator.mul` where it only calculates full matches and not partial matches.

mrocklin:
Ah, happy to retract my comment

hameerabbasi:
No, let's leave it up for other people who can't follow the terminology.

hameerabbasi:
Are there any further comments or is this good to merge?

mrocklin:
Merged!

PR #102, opened 2018-02-19
Add XArray compatibility features

hameerabbasi:
Once the right broadcasting setup was in place; it was trivial to implement the three-argument version of `where` and NaN-skipping aggregations.

cc @mrocklin Feedback welcome.
cc @shoyer Feedback *really* welcome, as you know the ins and outs of XArray. Of course, "I can't" is okay; as always. :-)

### DIFF
--- a/docs/generated/sparse.COO.nanmax.rst
+++ b/docs/generated/sparse.COO.nanmax.rst
@@ -1,0 +1,6 @@
+COO.nanmax
+==========
+
+.. currentmodule:: sparse
+
+.. automethod:: COO.nanmax
--- a/docs/generated/sparse.COO.nanmax.rst
+++ b/docs/generated/sparse.COO.nanmax.rst
@@ -1,6 +1,0 @@
-COO.nanmax
-==========
-
-.. currentmodule:: sparse
-
-.. automethod:: COO.nanmax
--- a/docs/generated/sparse.COO.nanmin.rst
+++ b/docs/generated/sparse.COO.nanmin.rst
@@ -1,6 +1,0 @@
-COO.nanmin
-==========
-
-.. currentmodule:: sparse
-
-.. automethod:: COO.nanmin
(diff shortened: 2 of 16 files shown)


hameerabbasi:
Unfortunately, there doesn't seem to be a way to override `np.nansum`. I added a print command in there, and:

```python
>>> import numpy as np
>>> import sparse
>>> x = np.asarray([5, 6, np.nan])
>>> s.nansum()
nanreduce
11.0
>>> np.nansum(s)
11.0
```

hameerabbasi:
Hmm. It seems that (for `nanmin` and `nanmax`), sometimes we get `-inf` as a value (on a full axis that contained JUST `nan` and nothing else), whereas NumPy actually returns `nan`. I think the actual value of the reduction in this case is debatable (I, for one, think empty min should be `inf`, etc, as a hobbyist mathematician).

In any case, I'm willing to leave this as an xfail for now. Any input welcome.

hameerabbasi:
I added tests and matched the Numpy API. Looks review ready to me. 💃

hameerabbasi:
Do we want to support the edge case of NaNs in object arrays? It's proving difficult.

mrocklin:
Make a test with pytest.mark.xfail  and move on?

shoyer:
I don't think there are many uses for object dtype sparse arrays. Mostly we
use those for strings, but I imagine sparse arrays are most useful for
numbers.

hameerabbasi:
It seems we don't support object arrays in many cases at the moment. I've opened #104 to track this but am inclined to ignore this.

hameerabbasi:
I fixed the `nan` issue (by using `fmin`/`fmax`); tested for it; and matched the Numpy warning (and tested for it). This time around I feel this really is ready for review.

mrocklin:
This looks good to me :)

hameerabbasi:
Merged!

PR #116, opened 2018-03-02
Reorganize coo.py

### LINKED ISSUE #112: Reorganize coo.py?
`coo.py` is getting large and unweildly. I sometimes spend a lot of time scrolling through it or finding stuff in it. I was thinking of breaking it up into smaller chunks for better maintainability.

I propose the following:

- A `coo` directory.
- `coo/coo.py` (for the `COO` class)
- Move large implementations out of the `COO` class into methods, one for `elemwise`, one for `reduction`, one for `__getitem__`, and one for other methods.
- Similarly for `DOK`.

I love any proposed changes to this structure.

hameerabbasi:
Fixes #112


hameerabbasi:
cc @mrocklin Feedback welcome. This is blocking some other work, so quick is good.

sparse/compatibility.py
```diff
@@ -1,21 +1,20 @@
 # flake8: noqa F401
+import sys
 
-try:
-    from itertools import izip_longest as zip_longest
-except ImportError:
-    from itertools import zip_longest
+cur_version = sys.version_info
```
mrocklin: I recomment avoiding the alias here and just using `sys.version_info` everywhere.  Otherwise when someone gets dropped into this file looking for a bug and they see `sys.version_info` rather than `cur_version` they will know immediately what it means rather than have to track down the reference.

sparse/coo/__init__.py
```diff
@@ -0,0 +1,4 @@
+from ._core import COO
+from ._elemwise import elemwise
+from ._common import tensordot, dot, concatenate, stack, triu, tril, where, \
+    nansum, nanprod, nanmin, nanmax, nanreduce
```
mrocklin: I recommend keeping these module names public.  I believe that private modules are somewhat rare.

mrocklin:
In general this seems fine to me.  I left a couple of small comments.

hameerabbasi:
Hmm. Seems like there are some issues between Travis and GitHub today... On the other PR, Travis builds didn't report after finishing, and on this one, it isn't starting.

PR #118, opened 2018-03-02
Fix negative axes reductions.

### LINKED ISSUE #117: Reduce operations raise error for negative axis values
Reduce operations like `.sum()`, `.prod()` etc. raise exceptions when you pass a negative `axis` value:

```python
import sparse
sparse.random((40, 50)).sum(0)  # works
sparse.random((40, 50)).sum(1)  # works
sparse.random((40, 50)).sum(-1)  # fails, should be equivalent to .sum(1)
```

The traceback I get:

```
  File "<stdin>", line 1, in <module>
  File "proj_dir/sparse/sparse/coo.py", line 1038, in prod
    return self.reduce(np.multiply, axis=axis, keepdims=keepdims, dtype=dtype)
  File "proj_dir/sparse/sparse/coo.py", line 771, in reduce
    a = self.transpose(neg_axis + axis)
  File "proj_dir/sparse/sparse/coo.py", line 1099, in transpose
    raise ValueError("repeated axis in transpose")
ValueError: repeated axis in transpose
```

hameerabbasi:
Fixes #117

### DIFF
--- a/sparse/coo.py
+++ b/sparse/coo.py
@@ -712,13 +712,15 @@ class COO(SparseArray, NDArrayOperatorsMixin):
         if not isinstance(axis, tuple):
             axis = (axis,)
 
+        axis = tuple(a if a >= 0 else a + self.ndim for a in axis)
+
         if set(axis) == set(range(self.ndim)):
             result = method.reduce(self.data, **kwargs)
             if self.nnz != self.size:
                 result = method(result, _zero_of_dtype(self.dtype)[()], **kwargs)
         else:
             axis = tuple(axis)
-            neg_axis = tuple(ax for ax in range(self.ndim) if ax not in axis)
+            neg_axis = tuple(ax for ax in range(self.ndim) if ax not in set(axis))
 
             a = self.transpose(neg_axis + axis)
             a = a.reshape((np.prod([self.shape[d] for d in neg_axis]),
--- a/sparse/tests/test_coo.py
+++ b/sparse/tests/test_coo.py
@@ -19,7 +19,7 @@ from sparse.utils import assert_eq, is_lexsorted, random_value_array
     ('prod', {}, {}),
     ('min', {}, {}),
 ])
-@pytest.mark.parametrize('axis', [None, 0, 1, 2, (0, 2)])
+@pytest.mark.parametrize('axis', [None, 0, 1, 2, (0, 2), -3, (1, -1)])
 @pytest.mark.parametrize('keepdims', [True, False])
 def test_reductions(reduction, axis, keepdims, kwargs, eqkwargs):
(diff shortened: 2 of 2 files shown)


sparse/coo.py
```diff
@@ -712,13 +712,15 @@ def reduce(self, method, axis=(0,), keepdims=False, **kwargs):
...
+        axis = tuple(a if a >= 0 else a + self.ndim for a in axis)
+
         if set(axis) == set(range(self.ndim)):
             result = method.reduce(self.data, **kwargs)
             if self.nnz != self.size:
                 result = method(result, _zero_of_dtype(self.dtype)[()], **kwargs)
         else:
             axis = tuple(axis)
-            neg_axis = tuple(ax for ax in range(self.ndim) if ax not in axis)
+            neg_axis = tuple(ax for ax in range(self.ndim) if ax not in set(axis))
```
mrocklin: Why this change?  I'm not entirely sure of Python semantics here but I suspect that you're creating a new set every time.

sparse/coo.py
```diff
@@ -712,13 +712,15 @@ def reduce(self, method, axis=(0,), keepdims=False, **kwargs):
         if not isinstance(axis, tuple):
             axis = (axis,)
 
+        axis = tuple(a if a >= 0 else a + self.ndim for a in axis)
```
mrocklin: Are there other uses of `axis=` where this should be done but is not yet currently done?

sparse/coo.py
hameerabbasi: Makes it `O(ndim)` rather than `O(ndim^2)`. I ran a quick test to see if this was indeed the case:

```python
>>> class A:
...     def get_set(self):
...         print("Called")
...         return set(range(5))
...     
>>> tuple(i for i in A().get_set())
Called
(0, 1, 2, 3, 4)
```

sparse/coo.py
hameerabbasi: With a short "find all" of `axis`, I'm inclined to say no. It does appear in `tensordot`, `concatenate` and `stack` but those already have this.

sparse/coo.py
mrocklin: This doesn't really matter either way.   Asymptotic arguments don't really carry significance when n is likely to be very small.


```python
In [1]: axis = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)

In [2]: ndim = 20

In [3]: %timeit tuple(ax for ax in range(ndim) if ax not in axis)
The slowest run took 6.44 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 4.03 µs per loop
```

Happy either way though.  I'm glad to learn that Python handles things in that order.

PR #123, opened 2018-03-03
Cython additions

hameerabbasi:
~~Do NOT merge this PR yet.~~

Update: Now ready to merge.

I wrote up a simple algorithm to do what `_match_arrays` does in Cython (it takes up most of the time in `elemwise`). Unfortunately, I can't seem to be able to make it faster than the original. At best, it approaches the original. `--march=native` and `-O3` don't make a difference.

Take this as a conversation starter or a "suggestions?" PR to build upon.


sparse/coo/umath_cython.pyx
```diff
@@ -0,0 +1,57 @@
...
+    a_ind = a_ind.astype(np.min_scalar_type(a.shape[0]))
+    b_ind = b_ind.astype(np.min_scalar_type(b.shape[0]))
+
+    return a_ind, b_ind
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef cywhere_wrapped(int64_t[:] a, int64_t[:] b):
+    cdef vector[int64_t] a_ind = vector[int64_t]()
+    cdef vector[int64_t] b_ind = vector[int64_t]()
```
mrocklin: If you know these sizes ahead of time I recommend pre-allocating these as static numpy arrays (or something similar).

mrocklin: Or even if you're only able to bound the size reasonably.

hameerabbasi: Unfortunately, I don't. It could be zero, it could be `na * nb` in the worst case. And having a "size" pass that just counts the size, then a second pass, makes it go from `6.55` to `~8` seconds.

mrocklin: Is there a particular benchmark that you're using?

hameerabbasi: Here it is:

```python
In[2]: import numpy as np
In[3]: import sparse
In[4]: x = sparse.random((1000, 1000, 1000), density=0.01)
In[5]: y = sparse.random((1000, 1000, 1000), density=0.01)
In[6]: %timeit x + y
```

hameerabbasi: Just so you know, this function takes in two sorted 1-D `ndarray`s, and performs the equivalent of an SQL outer join on them.

mrocklin: That's useful context, thanks!

I tried rewriting this in Numba using two passes and got similar results. https://gist.github.com/35f0d359f98b370c5a7b2cd862333577

Throwing in a couple print statements:

```python
In [1]: import sparse

In [2]: x = sparse.random((1000, 1000, 1000), density=0.01)

In [3]: y = sparse.random((1000, 1000, 1000), density=0.01)

In [4]: x + y
duration: 1.1437668800354004
n 100564 m 993185491
duration: 1.1581335067749023
duration: 1.039297342300415
n 100564 m 993185491
duration: 1.030043125152588
duration: 1.0377395153045654
n 100564 m 993774438
duration: 1.040597677230835
Out[4]: <COO: shape=(1000, 1000, 1000), dtype=float64, nnz=19899436, sorted=False, duplicates=False>
```

It's unfortunate that we seem to be passing through all billion possibilities.  Each pass takes around 1s, which naively matches what I would expect from a sequential operation on a GHz processor.  The fact that there are three calls with the same length is interesting, implying that we might be able to drop the first pass in the second two cases (although this still may not compete against the dynamic approach).  Perhaps there is some indexing structure that we can save in order to avoid the second pass as well?  I guess I'm curious why we call this function three times, and if there is work that can be saved between those subsequent calls (or if they're even necessary).

I also haven't looked at all at the algorithm to see if it's ideal.

hameerabbasi: > It's unfortunate that we seem to be passing through all billion possibilities.

We aren't. [See my SO question on the topic](https://stackoverflow.com/questions/47831287/get-index-of-matching-elements-in-array-taking-into-account-repetitions) to see what the algorithm does. TL;DR, it moves forward in each array of nonzeros, only adding those indices into the original arrays that match. It's linear in the output size, so it's as efficient as can be made.

For `operator.add`: One match for "common" elements, and two of the others to filter out "already matched" so the "overlap" isn't operated on with zero. As far as I can tell, this was necessary for broadcasting.

mrocklin: In my case the nnz was around 2e7 while the number of passes through the doubly nested loop was 1e9.  So the code that I wrote (which I think is similar to the code that you wrote) is not linear in the output size.

hameerabbasi:
This is now actually faster. I realized that I was resetting the index for `b` even when the next item _wasn't_ a match. This caused `ib` to jump backwards by large amounts sometimes, which was causing the slowdowns. To be fair, this was in the original code from SO.

Benchmarks:
Old:
```
In[7]: %timeit x + y
6.47 s ± 38.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
In[8]: %timeit x * y
1.91 s ± 21 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

New:
```
In[7]: %timeit x + y
4.14 s ± 37 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
In[8]: %timeit x * y
1.15 s ± 15.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

sparse/coo/umath_cython.pyx
```diff
@@ -0,0 +1,58 @@
...
+    cdef size_t na = a.shape[0], nb = b.shape[0]
+    cdef size_t ia, ib = 0
+    cdef uint64_t j
+    cdef uint64_t match = 0
+
+    with nogil:
+        for ia in range(na):
+            j = a[ia]
+            if j == b[match]:
+                ib = match
```
hameerabbasi: This is the condition that was missing previously.

hameerabbasi:
With a short cursory look, it seems the majority of time is now being taken up by `argsort` (44/75 ms in a 5000x5000 `operator.add` with density 0.01)

hameerabbasi:
I'm thinking of merging this since it's now faster by a significant margin. We would need to update installation docs to say you need a C++ compiler.

How would we manage distribution? A second `setup.py`?

mrocklin:
Adding Cython is, I believe, a significant step for this project.  I will
probably want to discuss this further but may not have time to do so this
weekend.  If possible I'd like to wait before merging.

mrocklin:
Certainly, a requirement of adding Cython would be clear documentation on
how to help developers build things locally for development, clear
instructions on how to build packages for PyPI, etc..

hameerabbasi:
Closing until #126 is properly resolved. Whatever is decided on, I'll rewrite this in that language.

PR #127, opened 2018-03-17
Numba matching speedup

hameerabbasi:
Benchmarks

```
In[2]: import numpy as np
In[3]: import sparse
In[4]: x = sparse.random((1000, 1000, 1000), density=0.01)
In[5]: y = sparse.random((1000, 1000, 1000), density=0.01)
In[6]: x.sum_duplicates()
In[7]: y.sum_duplicates()
In[8]: a = x.linear_loc()
In[9]: b = y.linear_loc()
In[10]: %timeit sparse.coo.umath._match_arrays(a, b)
```

Old:
```
852 ms ± 13.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

New:
```
112 ms ± 2.21 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```


rgommers:
How about making the benchmark code in this PR description into a proper benchmark? Copying the `asv` benchmark setup from scipy should be easy, and writing benchmarks is very simple.

About the `version.py` rename - my experience is that you want to keep that a private submodule.

hameerabbasi:
Thanks for the input, @rgommers, it's very welcome. I've set up a `asv`. I'll hopefully keep adding benchmarks as time goes on.

rgommers:
This looks good - not only an 8x speedup, but I find the code a lot more readable.

hameerabbasi:
Waiting on @mrocklin's input before I merge - He usually has excellent insights into different aspects of everything.

sparse/coo/src/umath.py
```diff
@@ -0,0 +1,44 @@
...
+            if j == b[ib]:
+                a_ind.append(ia)
+                b_ind.append(ib)
+
+                if b[match] < b[ib]:
+                    match = ib
+
+            ib += 1
+
+    return np.array(a_ind, dtype=np.uintp), np.array(b_ind, dtype=np.uintp)
```
mrocklin: Should we reduce the dtype here to be the minimum that can fit these values?

sparse/coo/umath.py
```diff
@@ -3,8 +3,9 @@
 import numpy as np
 import scipy.sparse
 
-from ..utils import isscalar, PositinalArgumentPartial, _zero_of_dtype
+from .src.umath import pywhere
```
mrocklin: Why define and import this in another module?  Can we define it here instead?  It would be nice to reduce the number of indirections that someone new to this codebase would have to make in order to find functionality.

mrocklin: Ah, it would be especially nice to remove the `src` submodule.  While these are quite common in other languages I rarely find them in Python code in the pydata ecosystem.  In general the community seems to have preferred a somewhat flat hierarchy when possible.

sparse/coo/src/umath.py
hameerabbasi: There are a couple of reasons against that but I'm divided myself:

- We would have to give up nopython mode and go into mixed mode.
- There would be so many distinct signatures (all possible input dtypes, and different output dtypes) that the JIT-ing itself would take a bit of time.

For it are obviously:

- Better performance (maybe?) I haven't tested.
- Lower memory consumption.

I'll let you make the call.

sparse/coo/umath.py
hameerabbasi: I've separated it for these reasons:

- coverage.py, CodeCov et.al. can't produce coverage for Numba JIT-ed functions. Keeping it separate lets us ignore it in coverage.
- Keeping Numba code separate (for a number of reasons I can't coherently put down in words)

hameerabbasi:
A couple of replies. Yes, rebase-and-merge seems like a good idea.

sparse/coo/src/umath.py
hameerabbasi: It seems I was right. The minimum dtype thing massively increases the time taken. 5.67 s for the benchmark in the comments. Writing a wrapper is still an option though.

mrocklin: OK, happy to skip this for now

sparse/coo/umath.py
mrocklin: I would prefer to keep things flat if possible.  I think that the submodule and file hierarchy are more important than coverage metrics, so coverage considerations shouldn't determine our hierarchy here.  You might consider using a [`#pragma: no cover`](http://coverage.readthedocs.io/en/coverage-4.3.4/excluding.html) comment instead.

I think that one of the advantages of Numba is that we *can* include it in our normal code and don't need to treat it as something special.  Removing this separation between low-level efficient code and normal code is, I think, important.

mrocklin: ... I think that keeping these things together is especially important for new developers or users of this code as they try to navigate things.   I agree that this isn't much of an issue for us as we're writing things, I do think that splitting things up and requiring navigation of a codebase affects readability for others though.

hameerabbasi: Sounds like a good idea. My main concern was code coverage, the pragma seems like a good way to handle that.

hameerabbasi:
cc @mrocklin If you could give this a final look before we merge, that'd be great. :-)

mrocklin:
+1 from me.  Feel free to rebase and merge.  

One other question I have (which could be answered at another time if you'd like to go ahead with this now)  is where the extra missed lines are coming from.  Looking at the codecov CI I suspect that it may be coming from the benchmark files?  Is this correct?  It doesn't seem like it should given what you've written down in the codecov.yaml file, but I'm not as familiar with this, so I'm not sure.

hameerabbasi:
The diff is here: https://codecov.io/gh/pydata/sparse/pull/127/diff#diff-YmVuY2htYXJrcy9iZW5jaG1hcmtfY29vLnB5

It seems to me that CodeCov uses the YAML from the base, it seems consistent with this and other behavior I've seen so far. This is unlike Travis, which uses the config in the PR.

There will be missed lines here but the next one should be fine, if my hunch is correct.

I validated the YAML accross their web API.

PR #128, opened 2018-03-20
Numba indexing speedups

### LINKED ISSUE #60: Indexing is O(n) when it can be made O(log n)
If we call `COO.sum_duplicates()` beforehand, it may be possible, with some trickery, to make indexing `O(log n)`. I'm still looking into how this can be done, and how to generalize it to multiple indices, but I know it's possible.

One concern I have is that I do not want it to become `O(n log n)` for multiple indices, so this needs to be thought out carefully.

hameerabbasi:
Significant speedups to indexing. Benchmark:
```python
import sparse
x = sparse.random((1000, 1000, 1000), density=0.01)
%timeit x[5]
```

master:
```
46.9 ms ± 397 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

This branch:
```
174 µs ± 34.5 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Note that this speedup is partially because of Numba but also because our old indexing was completely suboptimal in big-O complexity.

The partial support that we had for advanced indexing is currently broken, but I plan to work on full-blown advanced indexing soon.

As far as I can tell, however, it, too, requires `list[list]]` on Numba's end. xref numba/numba#2560

Fixes #60


hameerabbasi:
I will note, however, that this indexing is geared towards integer or tuple of integer. Large slices WILL make this slow as molasses. Maybe a fallback if the slices are too large?

hameerabbasi:
Latest benchmarks:
```
In[2]: import sparse
In[3]: x = sparse.random((1000, 1000, 1000), density=0.01)
In[4]: %timeit x[:500]
78.2 ms ± 613 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
In[5]: %timeit x[:500, :500]
223 ms ± 6.73 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
In[6]: %timeit x[:500, :500, :500]
230 ms ± 6.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
In[7]: %timeit x[5]
157 µs ± 423 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Faster in all cases than old. Falls back to simple matching based on a "guesstimate" of which one will be faster.

hameerabbasi:
cc @mrocklin Ready for review. :-)

mrocklin:
Cool.  I'm looking forward to it.   I took a brief glance at it and noticed that it was fairly involved, including many new routines with complex logic.  This will probably take some time to review properly and so I may not get to this today.  

If you're inclined I suspect that you could make reviewing (and thus reading by future devs) easier by including small and simple doctests within each utility function that show what it does with a minimal example.

hameerabbasi:
I had already partially done that, but I made it more complete.

sparse/coo/core.py
```diff
@@ -504,15 +507,17 @@ def __getitem__(self, index):
 
         last_ellipsis = len(index) > 0 and index[-1] is Ellipsis
         index = normalize_index(index, self.shape)
-        if len(index) != 0 and all(not isinstance(ind, Iterable) and ind == slice(None) for ind in index):
+
+        if len(index) != 0 and all(not isinstance(ind, Iterable) and ind == slice(0, dim, 1)
+                                   for ind, dim in zip_longest(index, self.shape)):
```
mrocklin: Is this change fixing an error that previously existed?  If so then is this tested somewhere?

sparse/coo/core.py
```diff
@@ -504,15 +507,17 @@ def __getitem__(self, index):
...
             return self
-        mask = np.ones(self.nnz, dtype=np.bool)
-        for i, ind in enumerate([i for i in index if i is not None]):
-            if not isinstance(ind, Iterable) and ind == slice(None):
-                continue
-            mask &= _mask(self.coords[i], ind, self.shape[i])
+        mask = _mask(self.coords, index, self.shape)
+
+        if isinstance(mask, slice):
+            n = len(range(mask.start, mask.stop, mask.step))
```
mrocklin: This may be somewhat inefficient in Python 2

sparse/coo/core.py
```diff
@@ -1572,28 +1560,303 @@ def _keepdims(original, new, axis):
...
+    start, stop, step = idx[0], idx[1], idx[2]
+
+    if step > 0:
+        return (stop - start + step - 1) // step
+    else:
+        return (start - stop - step - 1) // (-step)
+
+
+@numba.jit(nopython=True)
+def _get_mask_pairs_inner(starts_old, stops_old, c, idx):  # pragma: no cover
```
mrocklin: It looks like there are a number of functions like this that are only used once.  I'm somewhat inclined to just inline these directly into where they are called rather than construct a new function.  I personally find it difficult to follow a new codebase if I have to jump between several functions.  Giant long functions are also unpleasant of course, but I find them easier than those same lines of code split out to many places in the same file.

mrocklin: This seems to happen a number of times here, and I suspect it is partially what is partially what is causing me to be unable to review this PR efficiently.  I wonder if the explanation could be handled as well with comments.

sparse/coo/core.py
hameerabbasi: Not really. For the purposes of this project, I decided that `slice` objects without any `None` in them were better than those with `None` in the `start`, `stop` or step.

This (and the relevant changes in `slicing.py`) are to accommodate those changes.

sparse/coo/core.py
hameerabbasi: In Python 2, we are using `xrange` (imported from `compatibility.py` as `range`). It doesn't construct a list here, if that's what you're worried about.

sparse/coo/core.py
hameerabbasi: I did it that way at first, but apparently Numba has some unfixed bugs with type inference if I inline these into their parent functions. If I keep the logic unchanged and inline these functions, Numba compilation fails as it fails to infer the type of some of the lists.

hameerabbasi:
A couple of replies.

hameerabbasi:
I added a bit of comments to the main function.

sparse/coo/core.py
hameerabbasi: If it helps, bottom-up reviewing may be better in this case than top-down.

hameerabbasi:
Rebased onto infrastructure fix commits.

hameerabbasi:
@mrocklin I'm done working on this. Do you have time to review in the near future?

sparse/coo/core.py
hameerabbasi: I've merged at least two functions. I tried to remove another one, but that unfortunately tanks the performance.

sparse/coo/core.py
mrocklin: For the slice check why

    not isinstance(ind, Iterable) and ind == slice(0, dim, 1)

Rather than 

    ind == slice(0, dim, 1)

Separately, I'm curious about the use of `zip_longest`.

sparse/coo/core.py
```diff
@@ -1572,28 +1560,294 @@ def _keepdims(original, new, axis):
...
         else:
-            start = idx.start if idx.start is not None else (shape - 1)
-            stop = idx.stop if idx.stop is not None else -1
-            return (coords <= start) & (coords > stop) & \
-                   (coords % step == start % step)
-    elif isinstance(idx, Iterable):
-        mask = np.zeros(len(coords), dtype=np.bool)
-        for item in idx:
-            mask |= _mask(coords, item, shape)
+            ind_ar[i] = [idx, idx + 1, 1]
```
mrocklin: Just checking, but this will never get other numpy arrays of indices or boolean values?

sparse/coo/core.py
```diff
@@ -1572,28 +1560,294 @@ def _keepdims(original, new, axis):
...
-    if isinstance(idx, numbers.Integral):
-        return coords == idx
-    elif isinstance(idx, slice):
-        step = idx.step if idx.step is not None else 1
-        if step > 0:
-            start = idx.start if idx.start is not None else 0
-            stop = idx.stop if idx.stop is not None else shape
-            return (coords >= start) & (coords < stop) & \
-                   (coords % step == start % step)
+def _mask(coords, indices, shape):
```
mrocklin: Request small docstring here.  Reviewing this I don't know what you mean by a mask.  I can take a guess, but it's unlikely that someone who has to fix bugs in this code in the future will understand immediately.

mrocklin: Maybe it would make sense to move all of these functions to a separate file (slicing.py?) and add a module-level doscstring describing how they work together?

sparse/coo/core.py
```diff
@@ -1572,28 +1560,294 @@ def _keepdims(original, new, axis):
...
+    ----------
+    coords : np.ndarray
+        The coordinates of the array.
+    indices : np.ndarray
+        The indices in the form of slices.
+
+    Returns
+    -------
+    mask : np.ndarray
+        The starts and stops in the mask.
```
mrocklin: and steps?

sparse/coo/core.py
```diff
@@ -1572,28 +1560,294 @@ def _keepdims(original, new, axis):
...
+    is_slice : bool
+        Whether or not the array represents a continuous slice.
```
mrocklin: Continuous slide meaning step == 1 or just "a slice" of any sort?

sparse/coo/core.py
```diff
@@ -1572,28 +1560,294 @@ def _keepdims(original, new, axis):
...
+    """
+    # Set the initial mask to be the entire range of coordinates.
+    starts = [0]
+    stops = [coords.shape[1]]
+    n_matches = coords.shape[1]
+
+    i = 0
+    while i < len(indices):
+        # Guesstimate whether working with pairs is more efficient or
```
mrocklin: What do you mean pairs here?

sparse/tests/test_coo.py
```diff
@@ -815,10 +809,24 @@ def test_gt():
     (slice(2, 0, -1), slice(None, None), -1),
     (slice(-2, None, None),),
     (slice(-1, None, None), slice(-2, None, None)),
```
mrocklin: Does your new code add any new functionality?  For example are there types of slices that we now support that we didn't before?

sparse/coo/core.py
hameerabbasi: Right now advanced indexing isn't supported but it will be in the future. That is, technically, advanced indexing.

sparse/coo/core.py
hameerabbasi: This sounds like a good idea, but `slicing.py` isn't suitable as it's common to `COO` and `DOK`.

sparse/coo/core.py
hameerabbasi: Steps turn out to always be 1 in this algorithm.

sparse/coo/core.py
hameerabbasi: `step == 1`.

sparse/coo/core.py
hameerabbasi: Start-stop pairs.

sparse/tests/test_coo.py
hameerabbasi: No, it doesn't at the moment.

sparse/coo/core.py
mrocklin: A different name then?

sparse/coo/core.py
mrocklin: Is the organization of code as you've made it here easy to extend to advanced cases?

hameerabbasi: I'm fairly certain how it will be done, but I haven't mapped it out in code yet. However, it requires `list[list[int]]` on Numba's end.

hameerabbasi:
I think your idea of moving this to a separate module and adding a module-level docstring is probably best. I'll request you to hold off reviewing further until I make that change (it'll make things easier). I'll hopefully do this tomorrow at some point.

The algorithm itself probably isn't that difficult, but I can see the need for terminology being clarified.

mrocklin:
Thanks.  In general I'm excited about the performance gains here.  Another option would just be to merge and move ahead without review.  Ideally in the future we'll have more people who have time to both write and review code.

hameerabbasi:
I've clarified everything, hopefully. :-) Feel free to review and/or merge.

sparse/coo/indexing.py
```diff
@@ -0,0 +1,444 @@
...
+        break
+    if i != 0:
+        indices = indices[:-i]
+    return indices
+
+
+@numba.jit(nopython=True)
+def _compute_mask(coords, indices):  # pragma: no cover
+    """
+    Gets the mask for the coords given the indices in slice format.
```
mrocklin: We should define slice format somewhere.  Either here or in _mask above (and then refer to that function here).  I think you mean the (start, stop, step) triples.

sparse/coo/indexing.py
```diff
@@ -0,0 +1,444 @@
...
+
+    Works with either start-stop ranges of matching indices into coords
+    called "pairs" (start-stop pairs) or filters the mask directly, based
+    on which is faster.
+
+    Exploits the structure in sorted coords, which is that for a constant
+    value of coords[i - 1], coords[i] is sorted. It uses this sortedness
```
mrocklin: What does it mean for `coords[i]` to be sorted?

sparse/coo/indexing.py
```diff
@@ -0,0 +1,444 @@
...
+    starts = [0]
+    stops = [coords.shape[1]]
+    n_matches = coords.shape[1]
+
+    i = 0
+    while i < len(indices):
+        # Guesstimate whether working with pairs is more efficient or
+        # working with the mask directly
+        n_current_slices = _get_slice_len(indices[i]) * len(starts) + 2
+        if n_current_slices * np.log(n_current_slices / max(len(starts), 1)) > n_matches:
```
mrocklin: What is the motivation for these values?  Should they be documented somewhere?

sparse/coo/indexing.py
```diff
@@ -0,0 +1,444 @@
...
+    # Set the initial mask to be the entire range of coordinates.
+
+    starts = [0]
+    stops = [coords.shape[1]]
+    n_matches = coords.shape[1]
+
+    i = 0
+    while i < len(indices):
+        # Guesstimate whether working with pairs is more efficient or
+        # working with the mask directly
```
mrocklin: I'm curious, what is the performance cost to just using one strategy or the other?  If it is not too great then we might consider simplifying to one.

mrocklin:
It seems like a lot of the logic here is based around performance considerations.  Do we want to add benchmarks for these various cases?

sparse/coo/indexing.py
hameerabbasi: It's a binary search vs linear checking consideration. One is bigger sometimes than the other.

sparse/coo/indexing.py
hameerabbasi: Unfortunately both are needed. In some cases, one is better, in some cases the other, and the difference is usually large.

For example, simple integer slicing mask construction is close to `O(log nnz)` in one and `O(nnz)` in the other.

And for slices such as `x[:500, :500, :500]`, the first approach boils down to `O(500^3)` worst case. However, if we work with pairs for the first two and switch to a mask later, it's a lot faster.

I made sure that the tests hit both cases, so we should be good. :-)

sparse/coo/indexing.py
hameerabbasi: It means, concretely, that for all `i` that won't err, and all `value`s, `coords[i, coords[i-1] == value]` is sorted. (due to the linear sorting order). I'll edit this into the docstring.

sparse/coo/indexing.py
hameerabbasi: That seems like a good idea.

hameerabbasi:
> It seems like a lot of the logic here is based around performance considerations. Do we want to add benchmarks for these various cases?

Seems like an excellent idea!

sparse/coo/indexing.py
hameerabbasi: It's already defined in the parameters section.

hameerabbasi:
I noticed there were a few cases we supported but didn't actually test for. That isn't additional in this PR, but it was nice to add tests for those cases as well.

mrocklin:
> I'm fairly certain how it will be done, but I haven't mapped it out in code yet. However, it requires list[list[int]] on Numba's end.

How confident are we that this will be implemented soon and, if implemented, that it will be fast enough?  If neither of those are true then do we have a backup plan?

hameerabbasi:
> How confident are we that this will be implemented soon and, if implemented, that it will be fast enough? If neither of those are true then do we have a backup plan?

numba/numba#2560 is the issue we're looking for. It has a milestone of 0.38 (RC), which is due out April 11. Unless it's pushed to the next release, we should be okay.

There may be a way with flat data structures as well, but it's slightly more complicated. May be more performant, though. I'm shooting for next weekend. It requires some Numba trickery in any case (I think that trickery should be possible in `nopython` mode and if so, it'll be fast, roughly equivalent to `n` integer indexes where `n` is the size of the advanced index).

If nothing works out we can simply loop over the relevant indices in Python.

Do you need it urgently in your work?

mrocklin:
No, I just want to make sure that we don't over-engineer into a solution that can not be extended.  I suspect that accepting lists and arrays of boolean and integer values will be important for XArray support, which is a nice use case to target.

hameerabbasi:
Good thinking! But don't worry, I'm fairly sure that with integer indexing and some transposing and reshaping, we'll be good to go.

benchmarks/benchmark_coo.py
```diff
@@ -33,3 +38,24 @@ def time_add(self):
...
+        self.x[5]  # Numba compilation
+
+    def time_index_scalar(self):
+        self.x[5]
+
+    def time_index_slice(self):
+        self.x[:50]
+
+    def time_index_slice2(self):
+        self.x[:50, :50]
```
mrocklin: Are different steps likely to make a difference here?

sparse/coo/indexing.py
```diff
@@ -0,0 +1,450 @@
...
+
+    Parameters
+    ----------
+    starts_old, stops_old : list[int]
+        The input starts and stops
+    Returns
+    -------
+    starts, stops : list[int]
+        The reduced starts and stops.
+    Examples
```
mrocklin: Style nit, add line breaks before section titles.

mrocklin:
I still haven't gone through all of the logic here, but I've probably gone as far as I'm likely to.  In general things seem fine to me.  +1

Thanks for slogging through this @hameerabbasi .  I hope you're enjoying yourself :)

benchmarks/benchmark_coo.py
hameerabbasi: What matters most here is the length of the slice and the number of nonzeros within it.

sparse/coo/indexing.py
hameerabbasi: I'm not too used to RST, I believe that's required for clean rendering as well.

PR #139, opened 2018-04-22
Switch from Travis CI to CircleCI

### LINKED ISSUE #120: Move to CircleCI from Travis
[Travis contains a `pip` cache](https://docs.travis-ci.com/user/caching/). This would really help speed up builds as Miniconda doesn't have to be installed, and Packages just have to be cached.

The downsides are managing the cache effectively.

Is this worth considering?

gforsyth:
Jobs set to run in parallel and cache their respective conda environments.

Resolves #120 

Someone with admin rights will have to enable to circle hooks, and then I can rebase to trigger the job to make sure everything is working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,107 @@
+version: 2
+
+jobs:
+  build_36:
+    machine: true
+    environment:
+      PYTHON: "3.6"
+      ENV_NAME: "py36-sparse-test"
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - miniconda-v1-{{ checksum "ci/environment-3.6.yml" }}
+      - run:
+          name: install miniconda
+          command: |
+              if [ ! -d "/home/circleci/miniconda" ]; then
+                wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+                bash miniconda.sh -b -p $HOME/miniconda
+                export PATH="$HOME/miniconda/bin:$PATH"
+                conda config --set always_yes yes --set changeps1 no
+              fi
+              sudo chown -R $USER.$USER $HOME
+      - run:
+          name: configure conda
+          command: |
+              export PATH="$HOME/miniconda/bin:$PATH"
+              if [ ! -d "/home/circleci/miniconda/envs/py36-sparse-test" ]; then
+                conda update -q conda
+                conda env create -f ci/environment-${PYTHON}.yml --name=${ENV_NAME}
+                source activate ${ENV_NAME}
+              fi
+              conda env list
+              conda list ${ENV_NAME}
+      - save_cache:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+            source activate ${ENV_NAME}
+            pip install -e .[tests]
+      - run:
+          command: |
+            /home/circleci/miniconda/envs/py36-sparse-test/bin/pytest --flake8
+      - run:
+          command: |
+            /home/circleci/miniconda/envs/py36-sparse-test/bin/codecov
+
+  build_27:
```
hameerabbasi: There seems to be a lot of repetition between the two builds. Is there some way to bring that down?

ci/environment-2.7.yml
```diff
@@ -0,0 +1,14 @@
...
+dependencies:
+  - python=2.7
+  - pytest
+  - numpy
+  - scipy
+  - pytest-cov
+  - nomkl
+  - numba
+  - pip:
+      - pytest-flake8
```
hameerabbasi: Can we move this to conda-forge?

ci/environment-3.6.yml
```diff
@@ -0,0 +1,14 @@
...
+dependencies:
+  - python=3.6
+  - pytest
+  - numpy
+  - scipy
+  - pytest-cov
+  - nomkl
+  - numba
+  - pip:
+      - pytest-flake8
```
hameerabbasi: Same as above.

gforsyth:
Added a badge to the README and switched the environment files to use conda-forge instead of defaults.  There might be some way to avoid the build repetition, but I'm not familiar enough with circle 2 to know where (and to not bork up the cache restore).

hameerabbasi:
Hmm. Could you make a minor change? CircleCI was set to not build forked PRs.

hameerabbasi:
I turned off Travis builds completely, and made CircleCI required instead. Can you re-trigger with some change?

.circleci/config.yml
```diff
@@ -0,0 +1,103 @@
...
+          paths:
+            - "/home/circleci/miniconda"
+      - run:
+          command: |
+            export PATH="$HOME/miniconda/bin:$PATH"
+            source activate ${ENV_NAME}
+            pip install -e .[tests]
+      - run:
+          command: |
+            /home/circleci/miniconda/envs/py36-sparse-test/bin/pytest --flake8
```
hameerabbasi: Can you cd into the project directory and run `pytest`? It'll automatically pick up `pytest.ini` that way. You also won't need `--flake8`, it's already in `pytest.ini`.

This also makes the coverage not generate.

gforsyth:
Ok, should be all set now.  Had to properly set path for the `codecov` call because apparently `codecov` is calling the `coverage` entrypoint with `subprocess` (or something similarly strange)

hameerabbasi:
Thanks, @gforsyth! This is in.

PR #141, opened 2018-04-23
Make COO canonical in the constructor.

hameerabbasi:
Based on recommendations in https://github.com/scipy/scipy/issues/8162#issuecomment-344469036 I believe that we should make `COO` always be canonical.

This has several advantages:

- Nobody really wants the non-canonical version.
- We don't have to keep track of when or if it's ever canonical.
- The `x.sum_duplicates()` calls just disappear.
- The string representation becomes simpler. The `sorted=?, duplicates=?` was just confusing to someone who didn't know how `COO` really worked.
- The object is now truly immutable.

### DIFF
--- a/benchmarks/benchmark_coo.py
+++ b/benchmarks/benchmark_coo.py
@@ -9,9 +9,6 @@ class ElemwiseSuite(object):
         self.x = sparse.random((100, 100, 100), density=0.01)
         self.y = sparse.random((100, 100, 100), density=0.01)
 
-        self.x.sum_duplicates()
-        self.y.sum_duplicates()
-
         self.x + self.y  # Numba compilation
 
     def time_add(self):
@@ -30,9 +27,6 @@ class ElemwiseBroadcastingSuite(object):
         self.x = sparse.random((100, 1, 100), density=0.01)
         self.y = sparse.random((100, 100), density=0.01)
 
-        self.x.sum_duplicates()
-        self.y.sum_duplicates()
-
     def time_add(self):
         self.x + self.y
 
@@ -44,8 +38,6 @@ class IndexingSuite(object):
     def setup(self):
         np.random.seed(0)
         self.x = sparse.random((100, 100, 100), density=0.01)
-        self.x.sum_duplicates()
-
         self.x[5]  # Numba compilation
 
     def time_index_scalar(self):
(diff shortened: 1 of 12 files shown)


hameerabbasi:
@nils-werner This should address your concerns in #58. You're welcome to review as well.

mrocklin:
I don't have enough experience with performance applications to know if this is a good or a bad idea performance-wise in the common case.  I recommend that we ping the original author of the comment, @perimosocordiae , for his thoughts.  It might also be good to ask for more information and possibly for a wider viewpoint from some of the author maintainers of the library.  Perhaps this is a good use of the scipy-dev mailing list?  cc @stefanv

mrocklin:
I still recommend asking more broadly within the scipy community.  They may
have reasons for the sum_duplicates behavior.  Do we know why this is?  Are
there cases where this is important?  I don't personally have enough
hands-on experience with the uses of COO to make this kind of decision.

perimosocordiae:
I think this is a good idea. At least with scipy's sparse matrices, so many operations require canonical (or at least duplicate-summed) matrices that the odds of constructing a COO matrix with dups and never indirectly calling `sum_duplicates` are very low. This approach front-loads that work, which makes performance a lot simpler to reason about. It also makes maintenance easier, like @hameerabbasi commented.

The downside is that you're explicitly disallowing some uses of the object, like representing a multi-graph. I'd argue that those users are better served by an object designed for that purpose, though, and since this package is very new you're unlikely to break anyone's existing code.

Pinging the scipy-dev list for objections is still a good idea, though.

hameerabbasi:
I posted it to the Scipy-Dev mailing list. [Link to thread.](https://mail.python.org/pipermail/scipy-dev/2018-April/022797.html)

hameerabbasi:
Allowing another day for comments and then merging, in the absence of objections.

mrocklin:
Sounds good.  Thanks for doing this @hameerabbasi !

shoyer:
Just to be clear, this "canonicalizes" COO arrays, right? So you an still provide non-canonical inputs to the COO constructor, and this will convert them into canonical form?

If so, this makes complete sense to me.

hameerabbasi:
Yes, that's what it does. In addition, you can pass two flags (as before), `sorted` and `has_duplicates` which signify that canonicalization should be skipped.

PR #146, opened 2018-05-03
Inplace ops

### LINKED ISSUE #80: In-place operations
I was wondering about operators such as `operator.iadd`, etc. There are a few ways we can go about this:

1. Support them by mutating the object in-place and invalidating the cache, i.e. performing the operation and then making `self` a copy of the returned object.
2. Support them only when the sparsity structure is the same, and modify `data` in-place.
3. Don't support them at all.

If we want to maintain compatibility with Numpy code (I hope to make `COO` a mostly drop-in replacement for `ndarray` with a few exceptions at some point), I would go with 1, with a warning in the docs that in-place isn't really "in-place".

If we want to do our own thing... Then we have options 2 and 3.

hameerabbasi:
Closes #80 

Leverages the existing `__array_ufunc__` implementation without needing too much extra code.

### DIFF
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -676,8 +676,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):
         >>> s.sum()
         25
         """
-        assert out is None
-        return self.reduce(np.add, axis=axis, keepdims=keepdims, dtype=dtype)
+        return np.add.reduce(self, out=out, axis=axis, keepdims=keepdims, dtype=dtype)
 
     def max(self, axis=None, keepdims=False, out=None):
         """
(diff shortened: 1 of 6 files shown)


mrocklin:
This looks pretty slick :)

Might want to test something like the following as well:

```python
np.sin(x, out=x)
```

sparse/coo/core.py
```diff
@@ -1039,16 +1039,26 @@ def __rmatmul__(self, other):
...
-            return elemwise(ufunc, *inputs, **kwargs)
+            result = elemwise(ufunc, *inputs, **kwargs)
         elif method == 'reduce':
-            return COO._reduce(ufunc, *inputs, **kwargs)
+            result = COO._reduce(ufunc, *inputs, **kwargs)
         else:
             return NotImplemented
 
+        if out is not None:
+            out = out[0]
```
mrocklin: I've seen people do things like the following to ensure length

```python
(out,) = out
```

hameerabbasi:
Docs updated, warning added to docs that "in-place ops aren't really in-place".

Supported `out` kwarg for all ops now, including `round` and `COO.sum`, etc.

hameerabbasi:
cc @mrocklin I think this is ready for a final review. :-)

hameerabbasi:
Removed some unnecessary code for `astype` -- It doesn't have an `out` argument. It was what was causing coverage to fail.

mrocklin:
Merging.  Sorry for the delay in review.  Thanks @hameerabbasi

PR #147, opened 2018-05-03
Unfiy unary and n-ary elemwise

hameerabbasi:
This PR merges the unnecessary "unary ufunc" case and the "n-ary" ufunc case under a single umbrella, so the unary case is no longer a special case.

### DIFF
--- a/sparse/coo/umath.py
+++ b/sparse/coo/umath.py
@@ -57,8 +57,7 @@ def elemwise(func, *args, **kwargs):
         elif isscalar(arg) or (isinstance(arg, np.ndarray)
                                and not arg.shape):
             # Faster and more reliable to pass ()-shaped ndarrays as scalars.
-            if isinstance(arg, np.ndarray):
-                args[i] = arg[()]
+            args[i] = np.asarray(arg)[()]
 
             pos.append(i)
             posargs.append(args[i])
@@ -74,9 +73,6 @@ def elemwise(func, *args, **kwargs):
 
     if len(args) == 0:
         return func(**kwargs)
-
-    if len(args) == 1:
-        return _elemwise_unary(func, args[0], **kwargs)
 
     return _elemwise_n_ary(func, *args, **kwargs)
 
@@ -150,7 +146,7 @@ def _elemwise_n_ary(func, *args, **kwargs):
 
     args = list(args)
 
-    args_zeros = tuple(_zero_of_dtype(arg.dtype)[()] for arg in args)
+    args_zeros = tuple(_zero_of_dtype(np.dtype(arg)) for arg in args)
 
     func_value = func(*args_zeros, **kwargs)
     func_zero = _zero_of_dtype(func_value.dtype)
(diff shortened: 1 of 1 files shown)


mrocklin:
I'm afraid that I no longer have enough detailed knowledge of the individual internal functions to be able to quickly review this.  It's nice to see the removal of complexity though and the tests pass, so it's good to see!

sparse/coo/umath.py
```diff
@@ -57,8 +57,7 @@ def elemwise(func, *args, **kwargs):
         elif isscalar(arg) or (isinstance(arg, np.ndarray)
                                and not arg.shape):
             # Faster and more reliable to pass ()-shaped ndarrays as scalars.
-            if isinstance(arg, np.ndarray):
-                args[i] = arg[()]
+            args[i] = np.asarray(arg)[()]
```
hameerabbasi: Converting all arguments to Numpy-based `dtype`s so accessing their `dtype` isn't an issue.

sparse/coo/umath.py
```diff
@@ -150,7 +146,7 @@ def _elemwise_n_ary(func, *args, **kwargs):
 
     args = list(args)
 
-    args_zeros = tuple(_zero_of_dtype(arg.dtype)[()] for arg in args)
+    args_zeros = tuple(_zero_of_dtype(np.dtype(arg)) for arg in args)
```
hameerabbasi: Remove the indexing to get a 0D array instead of a scalar, some functions work only on arrays.

sparse/coo/umath.py
```diff
@@ -173,7 +169,7 @@ def _elemwise_n_ary(func, *args, **kwargs):
     result_shape = _get_nary_broadcast_shape(*[arg.shape for arg in args])
 
     # Concatenate matches and mismatches
-    data = np.concatenate(data_list) if len(data_list) else np.empty((0,), dtype=args[0].dtype)
+    data = np.concatenate(data_list) if len(data_list) else np.empty((0,), dtype=func_value.dtype)
```
hameerabbasi: Useful for comparisons, they can have any input `dtype`. But the output will always be the function value.

sparse/coo/umath.py
```diff
@@ -294,7 +290,7 @@ def _unmatch_coo(func, args, mask, cache, **kwargs):
     matched_arrays = _match_coo(*matched_args, cache=cache)
 
     pos = tuple(i for i, m in enumerate(mask) if not m)
-    posargs = [_zero_of_dtype(arg.dtype)[()] for arg, m in zip(args, mask) if not m]
+    posargs = [_zero_of_dtype(arg.dtype) for arg, m in zip(args, mask) if not m]
```
hameerabbasi: Same here, converting to array.

hameerabbasi:
It's mostly just removing the function, some changes were needed in the n-ary case to handle some corner cases (small lurking bugs). It should be all good now.

PR #148, opened 2018-05-03
Add COO.nonzero method.

### LINKED ISSUE #145: Grabbing nonzero indices
Is the preferred way to access the coordinates/indices to use something like

```
i, j, k = arr.coords  # for a 3d array
```

Is there a reason why `coo` arrays don't implement `nonzero` as a method? I think this would make `np.argwhere(arr)` work as well without converting to a dense array.

Huge thanks for developing this package -- it will be very useful to me.

hameerabbasi:
Closes #145

### DIFF
--- a/docs/generated/sparse.COO.nonzero.rst
+++ b/docs/generated/sparse.COO.nonzero.rst
@@ -1,0 +1,6 @@
+COO.nonzero
+===========
+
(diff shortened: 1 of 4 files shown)


mrocklin:
Nice to see this come out so easily

PR #150, opened 2018-05-05
add full function to simplify COO creation

ahwillia:
This package may end up being very important to me, so I wanted to dip my toes in and try adding a simple extension in case I want to add something more substantial later.

The basic idea is to make array creation easier when all the data values are the same (I think this is incredibly common, e.g. if you are representing sparse graph structure)

```python
sparse.full(np.array([[0,1,2],[0,1,2]]), 1).todense()

array([[1, 0, 0],
       [0, 1, 0],
       [0, 0, 1]])
```

A few things I wasn't sure on...
* how to document `**kwargs`
* Should I be accepting non-array objects for `coords`? In particular should I add something to the effect of `coords = np.asarray(coords)` at the top of my function

Hope that this is welcome and within scope. Thanks!

### DIFF
--- a/sparse/tests/test_coo.py
+++ b/sparse/tests/test_coo.py
@@ -264,6 +264,22 @@ def test_elemwise(func):
     assert_eq(func(x), fs)
 
 
+@pytest.mark.parametrize('func', [np.expm1, np.log1p, np.sin, np.tan,
+                                  np.sinh, np.tanh, np.floor, np.ceil,
+                                  np.sqrt, np.conj,
+                                  np.round, np.rint, np.conjugate,
+                                  np.conj, lambda x, out: x.round(decimals=2, out=out)])
+def test_elemwise_inplace(func):
+    s = sparse.random((2, 3, 4), density=0.5)
+    x = s.todense()
+
+    func(s, out=s)
+    func(x, out=x)
+    assert isinstance(s, COO)
+
+    assert_eq(x, s)
+
+
 @pytest.mark.parametrize('func', [
     operator.mul, operator.add, operator.sub, operator.gt,
     operator.lt, operator.ne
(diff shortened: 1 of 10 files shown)


sparse/tests/test_coo.py
```diff
@@ -1395,3 +1395,15 @@ def test_argwhere():
...
 
     assert_eq(np.argwhere(s), np.argwhere(x), compare_dtype=False)
+
+
+def test_full():
+    coords = np.tile(np.arange(5), (2, 1))
+
+    a = sparse.full(coords, 1)
+    e = np.diag(np.ones(5, dtype=int))
+    assert_eq(e, a, compare_dtype=True)
```
hameerabbasi: `compare_dtype` is on by default, no need to add it here.

sparse/tests/test_coo.py
```diff
@@ -1395,3 +1395,15 @@ def test_argwhere():
...
+
+    a = sparse.full(coords, 1.0)
+    e = np.diag(np.ones(5))
+    assert_eq(e, a, compare_dtype=True)
```
hameerabbasi: See comment to above.

mrocklin:
I wonder if it would make sense to have the COO constructor do this automatically

```python
x = COO(coords=..., data=1)
```

To my knowledge this doesn't conflict with any other potential input.

hameerabbasi:
Yes, we can. The question is whether we should add this to the constuctor, which is already quite bloated (both in terms of using and coding).

mrocklin:
I agree that it's somewhat bloated, however to me this doesn't seem to
increase that bloat substantially.

ahwillia:
Thanks for the feedback.

> I would recommend mirroring the signature of np.full rather than adding **kwargs, and then documenting those keywords as usual.

I am actually passing `kwargs` to the `COO` constructor **not** to `np.full`. My reasoning here is that the constructor signature may change? This might be an argument for instead adding this functionality to the constructor as @mrocklin suggested.

As someone who just started playing around with the package I actually find the COO constructor a bit confusing. I would be in favor of keeping `sparse.full(...)` because it matches a common function in numpy and I think familiar users will know how to use it immediately.

I don't mind also adding this to the `COO` constructor as well -- having multiple ways to do something isn't necessarily a bad thing. On the other hand, I kind of wish that the `COO` constructor didn't accept a dense numpy array and instead threw an error pointing to `COO.from_numpy(x)`. I managed to confuse myself quite a bit while initially playing around with this package.

Would it make sense to hold off on this PR and open an issue/discussion on the `COO` constructor behavior?

hameerabbasi:
Sure. I'm +0.5 on changing constructor behaviour. @ahwillia would you mind opening that issue?

ahwillia:
Closing this in favor of #152

Issue #143, opened 2018-04-23
Arbitrary fill value


hameerabbasi:
As requested in #1.

hameerabbasi:
@shoyer How important is this to XArray? While I find that implementing fill values is trivial algorithmically, a few lines go into every single function in the whole library, and may lead to a lot of corner cases and bugs. I'm debating whether it's worth it to implement this in practice.

shoyer:
I do think that sparse arrays in xarray be a very interesting alternative to using `MultiIndex` in pandas, especially when coupled with tools like dask for scaling to large datasets.

For this to work in any reasonable way, support for NaN/NA as a fill-value will be much pretty essential.
Most xarray operations automatically insert or skip NaN values (in data alignment/aggregations), so NaN is much more natural and useful fill-value than zero. Pandas's [sparse data-structures](https://pandas.pydata.org/pandas-docs/stable/sparse.html) support flexible fill-values defaulting to NaN for this same reason.

For computation oriented use-cases, I'm sure sparse would get a bit of use with xarray even if fill-values were always zero. But for analytics (which I think is the most exciting use-case), fill-values of NaN would be pretty much essential.

PR #152, opened 2018-05-09
Refactor and clean format conversions.

### LINKED ISSUE #151: COO constructor behavior
The COO constructor currently accepts many inputs. While this flexibility is nice, it isn't fully documented yet and we might want to modify it further. Here are a couple of points to start the discussion

-----------------------

1. Should COO accept scalar input for data (new functionality proposed in #150)?
```python
COO(coords, data=5)   # creates sparse array with 5s at all nonzero entries
```
-----------------------

2. Should `COO` convert dense numpy arrays or should we force `COO.from_numpy(...)` to be called separately?

```python
x = np.random.randint(10, size=(10,1000))
x[x < 5] = 0
s = COO(x)   # is x supposed to be interpreted as coords or as data?
s = COO.from_numpy(x)  # seems a bit more explicit to me
```

If we added a `dtype` option to the constructor I think this would be even more ambiguous.

```python
# cast `x` to bool array or create sparse array of True at indices specified by x?
s = COO(x, dtype=bool)
```

I suppose my confusion here stems from the assumption that calling `s = COO(coords, dtype=...)` without specifying `data` would initialize `self.data` with `np.empty(dtype=dtype)`

-----------------------

3. Other input forms `{(i, j, k): x, (i, j, k): y, ...}` or `[((i, j, k), value), (i, j, k), value), ...]` are also accepted but don't seem to be documented? Should these be accepted to the constructor or should we provide class method for it (e.g. `COO.from_dict(...)`)

hameerabbasi:
Closes #151

### DIFF
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -23,9 +23,9 @@ class COO(SparseArray, NDArrayOperatorsMixin):
     ----------
     coords : numpy.ndarray (COO.ndim, COO.nnz)
         An array holding the index locations of every value
-        Should have shape (number of dimensions, number of non-zeros)
+        Should have shape (number of dimensions, number of non-zeros).
     data : numpy.ndarray (COO.nnz,)
-        An array of Values
+        An array of Values. If not given, defers to :obj:`as_coo`.
     shape : tuple[int] (COO.ndim,)
         The shape of the array.
     has_duplicates : bool, optional
@@ -52,6 +52,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):
     See Also
     --------
     DOK : A mostly write-only sparse array.
+    as_coo : Convert any given format to :obj:`COO`.
 
     Examples
     --------
@@ -180,48 +181,9 @@ class COO(SparseArray, NDArrayOperatorsMixin):
         if cache:
             self.enable_caching()
         if data is None:
-            from ..dok import DOK
-
-            if isinstance(coords, COO):
-                self._make_shallow_copy_of(coords)
-                return
-
-            if isinstance(coords, DOK):
-                shape = coords.shape
-                coords = coords.data
-
-            # {(i, j, k): x, (i, j, k): y, ...}
-            if isinstance(coords, dict):
-                coords = list(coords.items())
-                has_duplicates = False
-
-            if isinstance(coords, np.ndarray):
-                result = COO.from_numpy(coords)
-                self._make_shallow_copy_of(result)
-                return
-
-            if isinstance(coords, scipy.sparse.spmatrix):
-                result = COO.from_scipy_sparse(coords)
-                self._make_shallow_copy_of(result)
-                return
-
-            # []
-            if not coords:
-                data = []
-                coords = []
-
-            # [((i, j, k), value), (i, j, k), value), ...]
-            elif isinstance(coords[0][0], Iterable):
-                if coords:
-                    assert len(coords[0]) == 2
-                data = [x[1] for x in coords]
-                coords = [x[0] for x in coords]
-                coords = np.asarray(coords).T
-
-            # (data, (row, col, slab, ...))
-            else:
-                data = coords[0]
-                coords = np.stack(coords[1], axis=0)
+            arr = as_coo(coords, shape=shape)
+            self._make_shallow_copy_of(arr)
+            return
 
         self.data = np.asarray(data)
         self.coords = np.asarray(coords)
@@ -278,7 +240,6 @@ class COO(SparseArray, NDArrayOperatorsMixin):
         True
         """
         self._cache = defaultdict(lambda: deque(maxlen=3))
-        return self
 
     @classmethod
     def from_numpy(cls, x):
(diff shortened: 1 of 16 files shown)


hameerabbasi:
cc @ahwillia I encourage you to take a look, if you can.

mrocklin:
My apologies for the delay.  I'm at a conference this week and may not have
a ton of time.  I'll put this high-ish on my todo list, but it's a slow
moving list at the moment.  My apologies for being a bottleneck.

hameerabbasi:
Ah, take your time, in that case.

mrocklin:
@jcrist *might* be interested in starting reviewing some code in this
project.

hameerabbasi:
It's always awesome to have more people involved! Developers, users, reviewers, or any other kind of role. 😃

ahwillia:
These changes generally look good to me. Now that broadcasting works in the `COO` constructor it seems quite intuitive to me. Does this supercede #150 ? It doesn't seem like there are any conflicts, but maybe `sparse.full(...)` is now unnecessary / redundant? I'm happy to close that PR if so.

If we were to keep `sparse.full(...)` would it make sense to add a `format` keyword? Right now the function is only able to produce a `COO` array.

hameerabbasi:
I hadn't added data broadcasting yet when you last commented, but I just did (along with a docs update). I'll leave the decision of whether or not to add `sparse.full` up to you. IMHO, it's unnecessary, and it doesn't match up exactly with `np.full`. In this library, we try to make things match exactly where possible.

In that light, `sparse.full` should be an array with all-same values. This will be possible after #143 is implemented.

ahwillia:
> In that light, sparse.full should be an array with all-same values.

I agree with this 👍

sparse/coo/core.py
```diff
@@ -372,14 +350,86 @@ def from_scipy_sparse(cls, x):
...
+        >>> s = COO.from_iter(x)
+        >>> s.todense()
+        array([[1, 0],
+               [0, 1]])
+        """
+        if isinstance(x, dict):
+            x = list(x.items())
+
+        if len(x) != 2 and not all(len(item) == 2 for item in x):
+            raise ValueError('Invalid iterable to convert to COO.')
```
mrocklin: Is it possible that they user will give an `Iterator` rather than an `Iterable`?  If so then this check will consume our data.  We might want to defensively convert to a tuple or list.

sparse/coo/core.py
```diff
@@ -1465,6 +1515,76 @@ def nonzero(self):
...
+    if isinstance(x, np.ndarray):
+        return COO.from_numpy(x)
+
+    if isinstance(x, scipy.sparse.spmatrix):
+        return COO.from_scipy_sparse(x)
+
+    if isinstance(x, Iterable):
+        return COO.from_iter(x, shape=shape)
+
+    raise NotImplementedError('Format not supported for conversion.')
```
mrocklin: It might be helpful to list here what was provided from the user and a set of valid values.

sparse/sparse_array.py
```diff
@@ -20,7 +20,7 @@ class SparseArray(object):
 
     def __init__(self, shape):
         if not isinstance(shape, Iterable):
-            shape = (int(shape),)
+            shape = (shape,)
```
mrocklin: Why this change?

sparse/coo/core.py
hameerabbasi: I'm not too familiar with the differences between the two. Will this check break or is it just a performance consideration?

sparse/sparse_array.py
hameerabbasi: If we provided a simple unsupported class that wasn't `Iterable` as a shape, this would break but give an unhelpful error message about not being able to cast to `int`. This way, the error message is more informative.

I tend to bunch in small bugfixes along with big PRs in order to get them in quickly, I don't really know how good of a habit this is.

sparse/coo/core.py
mrocklin: Iterators are ephemeral and get consumed on first use.  They are common, for example, when reading a sequence of data from a file.  Here is an example that creates an iterator (seq) from an iterable (L) does a check, and then discovers that some values have been removed:

```
In [1]: L = [1, 2, 3]

In [2]: seq = iter(L)

In [3]: if all(x > 100 for x in seq):
   ...:     print("hello")
   ...:     

In [4]: list(seq)
Out[4]: [2, 3]

In [5]: from collections import Iterator, Iterable

In [7]: isinstance(seq, Iterator)
Out[7]: True

In [8]: isinstance(L, Iterable)
Out[8]: True

In [9]: L
Out[9]: [1, 2, 3]
```

sparse/sparse_array.py
mrocklin: It is common for numpy integers to sneak in here.  Sometimes these cause issues later on.  It would be good to ensure that all values of `shape` are eventually coerced to be normal Python integers.

hameerabbasi: I anticipated this already. This is already done, a few lines below this one. :-)

It was being done here an extra time before the check. I just changed that so it happens after the check, and gives a meaningful error message.

https://github.com/pydata/sparse/blob/b9fc91cb68b37757f37881843b2e420a3c07f9be/sparse/sparse_array.py#L21-L29

sparse/coo/core.py
hameerabbasi: Ah, I understand now. I don't think making this a supported input is worth the extra effort.

mrocklin: I suspect that users will use iterators here whether or not we support it.  Currently we silently fail.  We should either fail loudly

```python
if isinstance(x, Iterator):
    raise TypeError("...")
```

Or we should coerce to a list

```python
if isinstance(x, Iterator):
    x = list(x)
```

hameerabbasi: Well, if we have to add code, I guess supporting it is better, so I did that.

sparse/coo/core.py
hameerabbasi: Done!

hameerabbasi:
Any further changes required here?

mrocklin:
No objection from me

PR #157, opened 2018-05-23
Miscellaneous fixes

hameerabbasi:
Some fixes (explained in the comments)

cc @mrocklin this should be small, but I'd like to clear the license thing with you first, if possible.


LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
```
hameerabbasi: License still BSD-3, but the copyright was changed from Anaconda Inc. to Sparse developers in order to acknowledge my own and others' contributions.

hameerabbasi: Also changed to not be RST so GitHub can detect it.

README.rst
```diff
@@ -5,7 +5,7 @@ Sparse Multidimensional Arrays
 
 This library provides multi-dimensional sparse arrays.
 
-* `Documentation <https://sparse.pydata.org/en/latest>`_
+* `Documentation <https://sparse.pydata.org/>`_
```
hameerabbasi: Didn't load in HTTPS, now it will.

docs/conf.py
```diff
@@ -62,7 +62,7 @@
 
 # General information about the project.
 project = 'sparse'
-copyright = '2017, Sparse Developers'
+copyright = '2018, Sparse developers'
```
hameerabbasi: Capitalization changed to better match SciPy and NumPy, and year changed to 2018.

setup.py
```diff
@@ -28,9 +28,8 @@
 }
 
 all_requires = []
-tests_requires = []
 
-for k, v in extras_require.items():
+for v in extras_require.values():
```
hameerabbasi: Didn't need the key, was a legacy from the `tests` extras.

setup.py
```diff
@@ -40,7 +39,7 @@
     version=versioneer.get_version(),
     cmdclass=versioneer.get_cmdclass(),
     description='Sparse n-dimensional arrays',
-    url='http://github.com/pydata/sparse/',
+    url='https://github.com/pydata/sparse/',
```
hameerabbasi: HTTPS, again.

sparse/tests/test_coo.py
```diff
@@ -240,12 +240,11 @@ def test_dot_nocoercion():
...
     la = a.tolist()
     lb = b.tolist()
-    la, lb  # silencing flake8
 
     if hasattr(operator, 'matmul'):
         # Operations with naive collection (list)
-        assert_eq(eval("la @ b"), eval("la @ sb"))
-        assert_eq(eval("a @ lb"), eval("sa @ lb"))
+        assert_eq(operator.matmul(la, b), operator.matmul(la, sb))
+        assert_eq(operator.matmul(a, lb), operator.matmul(sa, lb))
```
hameerabbasi: Converted `eval` to function call.

LICENSE
mrocklin: We probably need permission from @pzwang here.

Peter, this is a small project I did in my off time during employment at Anaconda Inc.  I hope that it becomes relevant at some point to the company's work, but these days it is mostly managed by others in the community, like @hameerabbasi .  It would be nice for the license to reflect that change.

pzwang: Thanks for the background.  This change looks good to me - and thanks to @hameerabbasi and others for working on the project!!

PR #159, opened 2018-06-12
Compile numba code with nogil=True

mrocklin:
This helps when parallelizing around sparse

### DIFF
--- a/sparse/coo/indexing.py
+++ b/sparse/coo/indexing.py
@@ -173,7 +173,7 @@ def _prune_indices(indices, shape, prune_none=True):
     return indices
 
 
-@numba.jit(nopython=True)
+@numba.jit(nopython=True, nogil=True)
 def _compute_mask(coords, indices):  # pragma: no cover
     """
     Gets the mask for the coords given the indices in slice format.
@@ -269,7 +269,7 @@ def _compute_mask(coords, indices):  # pragma: no cover
     return np.array(mask, dtype=np.intp), False
 
 
-@numba.jit(nopython=True)
+@numba.jit(nopython=True, nogil=True)
 def _get_mask_pairs(starts_old, stops_old, c, idx):  # pragma: no cover
     """
     Gets the pairs for a following dimension given the pairs for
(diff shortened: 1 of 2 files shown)


mrocklin:
Generic example:

```python
In [1]: import numba

In [2]: import numpy as np

In [3]: x = np.random.random(100000000)

In [4]: @numba.jit(nopython=True)
   ...: def f(x):
   ...:     total = 0
   ...:     for i in range(len(x)):
   ...:         total += x[i]
   ...:     return total
   ...: 

In [5]: f(x)  # run once to compile
Out[5]: 50001234.14227918

In [6]: %load_ext ptime

In [7]: %ptime -n 4 f(x)
Total serial time:   0.52 s
Total parallel time: 0.50 s
For a 1.05X speedup across 4 threads

In [8]: @numba.jit(nopython=True, nogil=True)
   ...: def f(x):
   ...:     total = 0
   ...:     for i in range(len(x)):
   ...:         total += x[i]
   ...:     return total
   ...: 

In [9]: f(x)  # run once to compile
Out[9]: 50001234.14227918

In [10]: %ptime -n 4 f(x)
Total serial time:   0.52 s
Total parallel time: 0.14 s
For a 3.63X speedup across 4 threads
```

hameerabbasi:
Is there any risk of incorrect results when doing this, assuming no race conditions?

mrocklin:
That's probably more a question for you than it is for me.  Do any of these functions rely on global state or mutate their inputs?  If they were called concurrently in multiple threads would that be a problem?

mrocklin:
Usually the answer is "no", but we may be doing inplace modifications for performance sake,.

mrocklin:
Previously these functions were run with a giant lock around them so that no other Python function would run at the same time.  Now we are removing that lock.

hameerabbasi:
> Do any of these functions rely on global state or mutate their inputs?

No, they don't do that. :-) Thanks for this, I wondered why the `dask.array` version was as slow as it was.

mrocklin:
Thanks for the review @hameerabbasi .  Merging!

PR #160, opened 2018-06-24
Implement roll (circular shift of elements)

ahwillia:
This is my attempt to implement the equivalent of [`numpy.roll`](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.roll.html) which I think would be really useful to have in this package.

I think this is nearly done, but it could be improved. In particular, the function is supposed to produce a copy of the array but I wasn't sure the preferred way of doing that so I called the `COO` constructor directly. Also, my function assumes a COO array as input, which is maybe not so flexible as other sparse array types are developed?

Thanks again for providing this awesome package. Looking forward to your comments!

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -1,6 +1,7 @@
 from functools import reduce
 import operator
 import warnings
+import collections
 
 import numpy as np
 import scipy.sparse
@@ -609,3 +610,69 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
     """
     arr = _replace_nan(x, method.identity if identity is None else identity)
     return arr.reduce(method, axis, keepdims, **kwargs)
+
+
+def roll(a, shift, axis=None):
+    """
+    Shifts elements of an array along specified axis. Elements that roll beyond
+    the last position are circulated and re-introduced at the first.
+
+    Parameters
+    ----------
+    x : COO
+        Input array
+    shift : int or tuple of ints
+        Number of index positions that elements are shifted. If a tuple is
+        provided, then axis must be a tuple of the same size, and each of the
+        given axes is shifted by the corresponding number. If an int while axis
+        is a tuple of ints, then broadcasting is used so the same shift is
+        applied to all axes.
+    axis : int or tuple of ints, optional
+        Axis or tuple specifying multiple axes. By default, the
+        array is flattened before shifting, after which the original shape is
+        restored.
+
+    Returns
+    -------
+    res : ndarray
+        Output array, with the same shape as a.
+    """
+    from .core import COO, as_coo
+    a = as_coo(a)
+
+    # roll flattened array
+    if axis is None:
+        return roll(a.reshape((-1,)), shift, 0).reshape(a.shape)
+
+    # roll across specified axis
+    else:
+        # parse axis input, wrap in tuple
+        axis = normalize_axis(axis, a.ndim)
+        if not isinstance(axis, tuple):
+            axis = (axis,)
+
+        # make shift iterable
+        if not isinstance(shift, collections.Iterable):
+            shift = (shift,)
+
+        elif np.ndim(shift) > 1:
+            raise ValueError(
+                "'shift' and 'axis' must be integers or 1D sequences.")
+
+        # handle broadcasting
+        if len(shift) == 1:
+            shift = np.full(len(axis), shift)
+
+        # check if dimensions are consistent
+        if len(axis) != len(shift):
+            raise ValueError(
+                "If 'shift' is a 1D sequence, "
+                "'axis' must have equal length.")
+
+        # shift elements
+        coords, data = np.copy(a.coords), np.copy(a.data)
+        for sh, ax in zip(shift, axis):
+            coords[ax] += sh
+            coords[ax] %= a.shape[ax]
+
+        return COO(coords, data=data, shape=a.shape, has_duplicates=False)
--- a/sparse/tests/test_coo.py
+++ b/sparse/tests/test_coo.py
@@ -1527,3 +1527,69 @@ def test_invalid_iterable_error():
     with pytest.raises(ValueError):
         x = [((2.3, 4.5), 3.2)]
         COO.from_iter(x)
+
+
+class TestRoll(object):
+
+    # test on 1d array #
+    @pytest.mark.parametrize('shift', [0, 2, -2, 20, -20])
+    def test_1d(self, shift):
+        xs = sparse.random((100,), density=0.5)
+        x = xs.todense()
+        assert_eq(np.roll(x, shift), sparse.roll(xs, shift))
+        assert_eq(np.roll(x, shift), sparse.roll(x, shift))
+
+    # test on 2d array #
+    @pytest.mark.parametrize(
+        'shift', [0, 2, -2, 20, -20])
+    @pytest.mark.parametrize(
+        'ax', [None, 0, 1, (0, 1)])
+    def test_2d(self, shift, ax):
+        xs = sparse.random((10, 10), density=0.5)
+        x = xs.todense()
+        assert_eq(np.roll(x, shift, axis=ax), sparse.roll(xs, shift, axis=ax))
+        assert_eq(np.roll(x, shift, axis=ax), sparse.roll(x, shift, axis=ax))
+
+    # test on rolling multiple axes at once #
+    @pytest.mark.parametrize(
+        'shift', [(0, 0), (1, -1), (-1, 1), (10, -10)])
+    @pytest.mark.parametrize(
+        'ax', [(0, 1), (0, 2), (1, 2), (-1, 1)])
+    def test_multiaxis(self, shift, ax):
+        xs = sparse.random((9, 9, 9), density=0.5)
+        x = xs.todense()
+        assert_eq(np.roll(x, shift, axis=ax), sparse.roll(xs, shift, axis=ax))
+        assert_eq(np.roll(x, shift, axis=ax), sparse.roll(x, shift, axis=ax))
+
+    # test original is unchanged #
+    @pytest.mark.parametrize(
+        'shift', [0, 2, -2, 20, -20])
+    @pytest.mark.parametrize(
+        'ax', [None, 0, 1, (0, 1)])
+    def test_original_is_copied(self, shift, ax):
+        xs = sparse.random((10, 10), density=0.5)
+        xc = COO(np.copy(xs.coords), np.copy(xs.data), shape=xs.shape)
+        sparse.roll(xs, shift, axis=ax)
+        assert_eq(xs, xc)
+
+    # test on empty array #
+    def test_empty(self):
+        x = np.array([])
+        assert_eq(np.roll(x, 1), sparse.roll(sparse.as_coo(x), 1))
+
+    # test error handling #
+    @pytest.mark.parametrize(
+        'args', [
+            # iterable shift, but axis not iterable
+            ((1, 1), 0),
+            # ndim(axis) != 1
+            (1, [[0, 1]]),
+            # ndim(shift) != 1
+            ([[0, 1]], [0, 1]),
+            ([[0, 1], [0, 1]], [0, 1])
+        ]
+    )
+    def test_valerr(self, args):
+        x = sparse.random((2, 2, 2), density=1)
+        with pytest.raises(ValueError):
+            sparse.roll(x, *args)
--- a/docs/generated/sparse.roll.rst
+++ b/docs/generated/sparse.roll.rst
@@ -1,0 +1,6 @@
+roll
+====
+
+.. currentmodule:: sparse
+
+.. autofunction:: roll
--- a/docs/generated/sparse.rst
+++ b/docs/generated/sparse.rst
@@ -45,6 +45,8 @@ API
 
     random
 
+    roll
+
     save_npz
 
     stack
--- a/sparse/coo/__init__.py
+++ b/sparse/coo/__init__.py
@@ -1,4 +1,4 @@
 from .core import COO, as_coo
 from .umath import elemwise
 from .common import tensordot, dot, concatenate, stack, triu, tril, where, \
-    nansum, nanprod, nanmin, nanmax, nanreduce
+    nansum, nanprod, nanmin, nanmax, nanreduce, roll


sparse/coo/common.py
```diff
@@ -609,3 +609,58 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+        return COO(a.coords, data=a.data, shape=a.shape)
+
+    # roll flattened array
+    elif axis is None:
+        return roll(a.reshape((-1,)), shift, 0).reshape(a.shape)
+
+    # roll across specified axis
+    else:
+        axis = normalize_axis_tuple(axis, a.ndim, allow_duplicate=True)
+        broadcasted = np.core.multiarray.broadcast(shift, axis)
```
hameerabbasi: Is this just broadcasting the shape? If so, we should use our own utility function. See https://github.com/pydata/sparse/blob/b51d74924d62ff6537b15ce4e1dd4e56080a3b6f/sparse/coo/umath.py#L358-L388

In general, I don't think it's good to rely on an undocumented NumPy function, which could change and break things at any time.

sparse/tests/test_coo.py
```diff
@@ -1527,3 +1527,35 @@ def test_invalid_iterable_error():
...
         x = [((2.3, 4.5), 3.2)]
         COO.from_iter(x)
+
+
+def test_roll_coo():
+    # 1d
+    x = np.array([0, 1, 0, 2, 0, 3, 0, 0, 4, 0, 0, 0, 5, 0, 6])
+    xs = sparse.as_coo(x)
+    for sh in (0, 2, -2, 20, -20):
+        assert_eq(np.roll(x, sh), sparse.roll(xs, sh).todense())
```
hameerabbasi: The `.todense()` can be removed.

sparse/tests/test_coo.py
```diff
@@ -1527,3 +1527,35 @@ def test_invalid_iterable_error():
...
+
+    # 2d
+    x2 = np.reshape(x, (3, 5))
+    x2s = sparse.as_coo(x2)
+    for ax in (None, 0, 1):
+        for sh in (0, 2, -2, 10, -10):
+            assert_eq(np.roll(x2, sh, axis=ax),
+                      sparse.roll(x2s, sh, axis=ax).todense())
```
hameerabbasi: The `.todense()` can be removed.

sparse/tests/test_coo.py
```diff
@@ -1527,3 +1527,35 @@ def test_invalid_iterable_error():
...
+
+    # 2d - roll multiple axes at once
+    shifts = [(0, 1), (1, 0), (-1, 1), (1, -1)]
+    for sh in shifts:
+        assert_eq(np.roll(x2, sh, axis=(0, 1)),
+                  sparse.roll(x2s, sh, axis=(0, 1)).todense())
```
hameerabbasi: The `.todense()` can be removed.

sparse/tests/test_coo.py
```diff
@@ -1527,3 +1527,35 @@ def test_invalid_iterable_error():
...
+
+    # 2d - roll multiple axes with duplicates
+    for ax in ((0, 0, 0), (1, 1, 1), (0, 1, 1), (0, 0, 1)):
+        for sh in [(0, 1, 0), (0, 1, 1), (-1, 1, -1), (1, -1, 1)]:
+            assert_eq(np.roll(x2, sh, axis=(0, 1, 1)),
+                      sparse.roll(x2s, sh, axis=(0, 1, 1)).todense())
```
hameerabbasi: The `.todense()` can be removed.

sparse/tests/test_coo.py
```diff
@@ -1527,3 +1527,35 @@ def test_invalid_iterable_error():
...
+
+    # empty
+    x = np.array([])
+    assert_eq(np.roll(x, 1), sparse.roll(sparse.as_coo(x), 1).todense())
```
hameerabbasi: The `.todense()` can be removed.

sparse/tests/test_coo.py
```diff
@@ -1527,3 +1527,35 @@ def test_invalid_iterable_error():
     with pytest.raises(ValueError):
         x = [((2.3, 4.5), 3.2)]
         COO.from_iter(x)
+
+
+def test_roll_coo():
```
hameerabbasi: Optional: Maybe follow the pattern of using `sparse.random` to generate something of a given shape and then roll it using both NumPy and sparse implementations.

hameerabbasi: Mandatory: Parameterize your tests. There are a lot of examples in the tests already, and [docs here](https://docs.pytest.org/en/latest/parametrize.html).

sparse/coo/common.py
```diff
@@ -609,3 +609,58 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+
+        if broadcasted.ndim > 1:
+            raise ValueError(
```
hameerabbasi: Needs a test for this branch. [See this SO question](https://stackoverflow.com/questions/23337471/how-to-properly-assert-that-an-exception-gets-raised-in-pytest).

sparse/coo/common.py
```diff
@@ -609,3 +609,58 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+                "'shift' and 'axis' should be scalars or 1D sequences")
+
+        # make a copy (TODO: should there be sparse.copy function?)
+        result = COO(a.coords, data=a.data, shape=a.shape)
+
+        for sh, ax in broadcasted:
+            result.coords[ax] += sh
+            result.coords[ax] %= result.shape[ax]
```
mrocklin: Can I ask for a test that verifies that the original remains unchanged after the roll operation?  

I suspect that `result.coords is a.coords` here.

sparse/coo/common.py
ahwillia: Hmm what about [`normalize_axis_tuple`](https://github.com/numpy/numpy/blob/59bd9ee273f26f335f1f77e0b85ede8d209a7e07/numpy/core/numeric.py#L1469) is there a utility function for that as well?

sparse/coo/common.py
ahwillia: Yes this is something I wasn't 100% sure on. What is the best/preferred way to deepcopy a `COO` array at the moment?

sparse/tests/test_coo.py
ahwillia: Interesting.... This causes the test to fail at `assert is_canonical(...)` what is this checking for? Do I need to resort the coordinates or something inside `sparse.roll(...)` so that I maintain lexographic order of the indices?

sparse/coo/common.py
ahwillia: Nevermind I found `normalize_axis` in utils.

sparse/tests/test_coo.py
hameerabbasi: `COO` arrays are meant to be immutable. This means you shouldn't construct and then change their `coords`/`data`.

What you should do instead is, make `coords`/`data` and pass that to the constructor.

`is_canonical` checks if the `coords` have any duplicates, and are sorted. This is needed by many operations internally. In your case, they're not sorted, because you're changing them directly in the object, instead of passing them to the constructor. I suggest you use `COO(..., has_duplicates=False)`

sparse/coo/common.py
hameerabbasi: There isn't a built-in method, because we kind of assume immutability, and if things are immutable, you don't need to worry about deep copies.

However, you can just do `coords = np.array(a.coords)` and similarly for `data` and then proceed. You're also modifying the object after construction (which is a big no-no), so this method is best anyway.

sparse/coo/common.py
hameerabbasi: You'll need to add `allow_repeat` into that one, I believe.

ahwillia:
I think all comments are addressed. Thanks so much for the feedback.

A few remaining questions:

* How should this function handle a non-COO input? Should I add a call to `as_coo(...)` at the top?
* Should I do more work to check inputs (e.g. if user provides a 2D sequence for `axis` or `shift`, should I throw a more descriptive error message?)

sparse/coo/common.py
```diff
@@ -609,3 +609,72 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+        Output array, with the same shape as a.
+    """
+    from .core import COO
+
+    # roll flattened array
+    if axis is None:
+        return roll(a.reshape((-1,)), shift, 0).reshape(a.shape)
+
+    # rolling does nothing for empty array
+    elif a.nnz == 0:
```
hameerabbasi: Optional: Is this branch really required? I'd do the same for all special cases that aren't required. Will the above code already handle it?

ahwillia:
Squashed everything to a single commit. Ready for final review.

sparse/coo/common.py
ahwillia: Ah yep let me get rid of that branch quickly

sparse/coo/common.py
```diff
@@ -609,3 +609,71 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+        elif not np.iterable(axis):
+            axis = (axis,)
+
+        # check if dimensions are consistent
+        if len(axis) != len(shift):
+            raise ValueError(
+                "If 'shift' is a 1D sequence, "
+                "'axis' must have equal length.")
+
+        if np.asarray(axis).ndim > 1 or np.asarray(shift).ndim > 1:
```
hameerabbasi: Use `np.ndim(...)` instead of `np.asarray(...).ndim`

sparse/coo/common.py
```diff
@@ -609,3 +609,71 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+            raise ValueError(
```
hameerabbasi: Mandatory: Add test for this `ValueError`, perhaps as a parametrized test in `test_valerr`.

sparse/coo/common.py
```diff
@@ -609,3 +609,71 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+        # make shift iterable
+        if not np.iterable(shift):
+            shift = (shift,)
+
+        # handle broadcasting
+        if np.iterable(axis) and len(shift) == 1:
+            shift = np.full(len(axis), shift)
+
+        # ensure axis is iterable
+        elif not np.iterable(axis):
```
hameerabbasi: Replace `np.iterable(axis)` with `isinstance(axis, tuple)` (`normalize_axis` does that for you).

sparse/coo/common.py
```diff
@@ -609,3 +609,71 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+    else:
+        # parse axis input
+        axis = normalize_axis(axis, a.ndim)
+
+        # make shift iterable
+        if not np.iterable(shift):
+            shift = (shift,)
+
+        # handle broadcasting
+        if np.iterable(axis) and len(shift) == 1:
```
hameerabbasi: Replace `np.iterable(axis)` with `isinstance(axis, tuple)` (`normalize_axis` does that for you).

sparse/coo/common.py
```diff
@@ -609,3 +609,71 @@ def nanreduce(x, method, identity=None, axis=None, keepdims=False, **kwargs):
...
+    if axis is None:
+        return roll(a.reshape((-1,)), shift, 0).reshape(a.shape)
+
+    # roll across specified axis
+    else:
+        # parse axis input
+        axis = normalize_axis(axis, a.ndim)
+
+        # make shift iterable
+        if not np.iterable(shift):
```
hameerabbasi: Replace `np.iterable(shift)` with `isinstance(shift, collections.Iterable)`

sparse/coo/common.py
ahwillia: I think that should be in there already at line 1585.

sparse/coo/common.py
ahwillia: nice. thanks 👍

sparse/coo/common.py
ahwillia: done.

sparse/coo/common.py
hameerabbasi: I see. There are a few things wrong with that. It's not parameterized. And when the top line raises an error, that one is detected and the last line is ignored.

sparse/coo/common.py
ahwillia: done.

sparse/coo/common.py
ahwillia: done.

sparse/coo/common.py
```diff
@@ -637,6 +637,7 @@ def roll(a, shift, axis=None):
         Output array, with the same shape as a.
     """
     from .core import COO, as_coo
+    import collections
```
hameerabbasi: Can you do this at the top of the file?

sparse/coo/common.py
hameerabbasi: You can see here that it isn't covered [here](https://codecov.io/gh/pydata/sparse/compare/eeae480e952328b30bf7e1b85d18318a35332bd8...85bd2cf675492d352ccc89bfbbce2c263f49e532/diff).

sparse/coo/common.py
ahwillia: yep. done in latest commit.

hameerabbasi: And remove it from here?

sparse/coo/common.py
ahwillia: Got it. Thanks for the patient explanation. Should be fixed now.

sparse/coo/common.py
ahwillia: https://github.com/ahwillia/sparse/commit/a6339db9498202fcee4d7561c96783bee26f6a3a

hameerabbasi: Ah, my mistake. Did you push it to the repo, are you sure it isn't local?

ahwillia: image

hameerabbasi: I get the same. I'm assuming GitHub is migrating stuff, and since it uses microservices, one thing is breaking at a time. Guess we'll just have to wait until things catch up and the tests run.

hameerabbasi: Could you possibly make a minor change and re-push to the repo? Possibly squash all the commits?

hameerabbasi:
Thanks for the patience and the addition, @ahwillia! This is in!

PR #165, opened 2018-07-15
Initial implementation for arbitrary fill values.

### LINKED ISSUE #143: Arbitrary fill value

hameerabbasi:
Closes #143 

Initial implementation for arbitrary fill values.

### DIFF
--- a/docs/operations.rst
+++ b/docs/operations.rst
@@ -227,3 +227,7 @@ Other Operations
 :obj:`dot`, :obj:`tensordot`, :obj:`concatenate`
 and :obj:`stack`, :obj:`transpose <COO.transpose>` and :obj:`reshape <COO.reshape>`.
 You can view the full list on the :doc:`API reference page <generated/sparse>`.
+
+.. note:: Some operations require zero fill-values (such as :obj:`nonzero <COO.nonzero>`)
+   and others (such as :obj:`concatenate`) require that all inputs have consistent fill-values.
+   For details, check the API reference.
--- a/docs/operations.rst
+++ b/docs/operations.rst
@@ -26,7 +26,6 @@ yet implemented for sparse arrays.
 
 .. code-block:: python
 
-   x + 1      # operations that produce dense results not supported
    np.svd(x)  # sparse svd not implemented
 
 
@@ -50,19 +49,9 @@ For example, the following will add two arrays:
 
 Auto-Densification
 ~~~~~~~~~~~~~~~~~~
-Operations that would result in dense matrices, such as binary
-operations with :doc:`Numpy arrays <reference/generated/numpy.ndarray>` objects or certain operations with
-scalars are not allowed and will raise a :obj:`ValueError`. For example,
-all of the following will raise a :obj:`ValueError`. Here, :code:`x` and
-:code:`y` are :obj:`COO` objects.
-
-.. code-block:: python
-
-   x == y
-   x + 5
-   x == 0
-   x != 5
-   x / y
+Operations that would result in dense matrices, such as
+operations with :doc:`Numpy arrays <reference/generated/numpy.ndarray>`
+objects a :obj:`ValueError`.
 
 However, all of the following are valid operations.
 
@@ -75,6 +64,11 @@ However, all of the following are valid operations.
    5 * x
    x / 7.3
    x != 0
+
+We also support operations with a nonzero fill value. These are operations
+that map zero values to nonzero values, such as :code:`x + 1` or :code:`operator.not_`.
+In these cases, they will produce an output with a fill value of :code:`1` or :code:`True`,
+assuming the original array has a fill value of :code:`0` or :code:`False` respectively.
 
 If densification is needed, it must be explicit. In other words, you must call
 :obj:`COO.todense` on the :obj:`COO` object. If both operands are :obj:`COO`,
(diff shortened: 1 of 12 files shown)


sparse/coo/common.py
```diff
@@ -56,6 +56,7 @@ def linear_loc(coords, shape):
     return out
 
 
+@check_fill_value(2)
```
shoyer: I find it more readable to use keyword argument with literal values, e..g, `@check_zero_fill_value(nargs=2)`

sparse/utils.py
```diff
@@ -240,3 +242,44 @@ def normalize_axis(axis, ndim):
         return tuple(normalize_axis(a, ndim) for a in axis)
 
     raise ValueError("axis %s not understood" % axis)
+
+
+def equivalent(x, y):
+    return (x == y) | ((x != x) & (y != y))
```
shoyer: note: if you're calling this on arrays, you could skip the non-NA checks for dtypes that can't hold NaN/NaT.

sparse/utils.py
```diff
@@ -240,3 +242,44 @@ def normalize_axis(axis, ndim):
...
+
+
+def check_fill_value(nargs):
+    def generator(func):
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
+            for arg in args[:nargs]:
```
shoyer: Watch out: This has one of the same issues we encountered in NEP 18: it means the public API for these functions will be changed, to only accepted positional arguments.

I would suggest writing helper functions to call *inside* `func` inside, e.g., `check_zero_fill_value(a, b)`.

sparse/utils.py
```diff
@@ -240,3 +242,44 @@ def normalize_axis(axis, ndim):
...
+                if hasattr(arg, 'fill_value') and \
```
shoyer: Nit: per PEP 8, prefer using extra parentheses `()` to explicit continuation with `\`.

sparse/utils.py
```diff
@@ -240,3 +242,44 @@ def normalize_axis(axis, ndim):
...
+        return wrapped
+
+    return generator
+
+
+def consistent_fill_value(func):
+    from .sparse_array import SparseArray
+
+    @functools.wraps(func)
+    def wrapped(arrays, *args, **kwargs):
```
shoyer: As above, consider using a function instead of a decorator here.

In this case, the decorator would work OK (as long as the first argument is always called `arrays`) but decorators are more magical than simple function calls.

docs/quickstart.rst
```diff
@@ -51,22 +51,22 @@ This includes arithmetic, :doc:`numpy.ufunc <reference/ufuncs>` operations, or f
 .. code-block:: python
 
    >>> np.sin(s) + s.T * 1
-   <COO: shape=(100, 100, 100), dtype=float64, nnz=189601>
+   <COO: shape=(100, 100, 100), dtype=float64, nnz=189601, fill_value=0.0>
 
-However, operations which convert the sparse array into a dense one will raise exceptions
-For example, the following raises a :obj:`ValueError`.
+However, operations which convert the sparse array into a dense one will usually change the fill
+value instead of raising an error.
```
shoyer: The new sentence here is a little confusing to me. These operations how change the fill value instead of converting a sparse array to a dense array, so they don't "convert the sparse array into a dense one" at all now.

docs/operations.rst
```diff
@@ -265,7 +216,6 @@ All of the following will raise an :obj:`IndexError`, like in Numpy 1.13 and lat
    z[3, 6]
    z[1, 4, 8]
    z[-6]
-   z[[True, True, False, True], 3, 4]
 
 .. note:: Numpy advanced indexing is currently not supported.
```
shoyer: below here: maybe note that `stack` and `concatenate` require matching fill values, and that some operations (e.g., `tensordot`) require a fill value of zero?

hameerabbasi: Done!

docs/quickstart.rst
hameerabbasi: Changed the sentence structure a bit.

hameerabbasi:
@shoyer I think this is now ready to merge.
@mrocklin If you have time, if it's going to take too long, let me know.
@ahwillia If you're willing to review. Reviewing is great for learning the codebase, if you're interested.

sparse/utils.py
```diff
@@ -258,13 +287,53 @@ def equivalent(x, y):
...
+    ValueError: This operation requires zero fill values.
+    >>> check_zero_fill_value(s1, s2)
+    Traceback (most recent call last):
+        ...
+    ValueError: This operation requires zero fill values.
+    """
     for arg in args:
         if (hasattr(arg, 'fill_value') and
                 not equivalent(arg.fill_value, _zero_of_dtype(arg.dtype))):
             raise ValueError('This operation requires zero fill values.')
```
shoyer: It's a minor point, but it's nice to include offending values in all error messages, e.g., `arg.fill_value` in this case

sparse/utils.py
```diff
@@ -276,4 +345,4 @@ def check_consistent_fill_value(arrays):
     fv = arrays[0].fill_value
 
     if not all(equivalent(fv, s.fill_value) for s in arrays):
-        raise ValueError('Consistent fill-values required.')
+        raise ValueError('This operation requires consistent fill-values.')
```
shoyer: same concern as above about including fill values in the error message

sparse/utils.py
hameerabbasi: I'm not sure what best practice would be here. We wouldn't know what exact argument would produce this fill value, so showing it may be not be useful. What do you think?

shoyer: Maybe:
```python
'This operation requires zero fill values, but argument {} has fill value {}'.format(i, arg.fill_value)
```
where `i` comes from iterating over `args` with `enumerate()`.

hameerabbasi: Done.

mrocklin:
> @mrocklin If you have time, if it's going to take too long, let me know.

I appreciate getting pinged, but I probably won't be reviewing this one.  That's also fine.  I probably shouldn't be active on every pull request here.  It looks like @shoyer seems pretty happy, which is a good sign.  I trust his attention to detail :)

hameerabbasi:
Well I believe there should be at least one reviewer. I'm thinking of ways to rope new contributors in... Reviewers or code-wise. 😄

PR #166, opened 2018-07-18
WIP: ENH: support other computational backends

stsievert:
This library relies heavily on the NumPy backend. This PR makes it easier to generalize to other computational backends, but doesn't change any behavior.  This is still a WIP and hear other people's opinions.

This PR really only modifies

* the storage for COO
* the `todense` function of COO

It does this by passing a backend to COO.

TODO:

- [ ] see if `np.broadcast_to` should live in the backend
- [ ] test with other backends
- [ ] document

[Tensorly]:http://tensorly.org/stable/index.html

### DIFF
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -187,7 +187,11 @@ class COO(SparseArray, NDArrayOperatorsMixin):
     __array_priority__ = 12
 
     def __init__(self, coords, data=None, shape=None, has_duplicates=True,
-                 sorted=False, cache=False):
+                 sorted=False, cache=False, backend=None):
+        import numpy
+        self._T = numpy if backend is None else backend
+        T = self._T
+
         self._cache = None
         if cache:
             self.enable_caching()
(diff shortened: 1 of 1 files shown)


stsievert:
The motivation behind the PR is to support sparse tensors in [Tensorly]. The main difficulty is that Tensorly supports many different computational backends, including Numpy, CuPy, PyTorch, Tensorflow and MXNet. The backends all conform to a simple API which can be found in [tensorly.backends].

[Tensorly]:http://tensorly.org/stable/index.html
[tensorly.backends]:https://github.com/tensorly/tensorly/tree/6649dfcccf2165dffdf8fcb0172cf41cd6419d64/tensorly/backend

I'd like this library to be the backbone for Tensorly sparse tensors. This is the easiest approach for that, and it requires very little duplicate code and seems to be a large change (mostly because it's only the storage format).

@hameerabbasi can you give this a look over? I'd like some feedback.

[tensorly.backend]:https://github.com/tensorly/tensorly/tree/6649dfcccf2165dffdf8fcb0172cf41cd6419d64/tensorly/backend

sparse/coo/core.py
```diff
@@ -187,7 +187,11 @@ class COO(SparseArray, NDArrayOperatorsMixin):
     __array_priority__ = 12
 
     def __init__(self, coords, data=None, shape=None, has_duplicates=True,
-                 sorted=False, cache=False):
+                 sorted=False, cache=False, backend=None):
+        import numpy
+        self._T = numpy if backend is None else backend
```
hameerabbasi: A more descriptive name would be nice. Also, can we directly use `np`?

hameerabbasi:
@stsievert Have you looked into the [`__array_function__` protocol](http://www.numpy.org/neps/nep-0018-array-function-protocol.html)? It might be relevant for choosing back-ends.

mrocklin:
> I'd like this library to be the backbone for Tensorly sparse tensors. This is the easiest approach for that

Why is this the case?  It's not yet clear to me how making this library backed by other ndarray implementations achieves the sparse tensorly objective.

Allowing pydata/sparse to be backed by other ndarray implementations does come with a maintenance and complexity burden for pydata/sparse.  We should make sure that this is necessary first.

hameerabbasi:
Indeed, I think it should be the other way around IIUC... Tensorly should have a backend for pydata/sparse.

stsievert:
> Tensorly should have a backend for pydata/sparse.

I agree. That will allow quicker development too, and it's cleaner.

PR #167, opened 2018-07-19
Move back to Travis-CI

hameerabbasi:


mrocklin:
I'm curious, what was the motivation here?  Most people I've spoken to recently have had a nicer time with Circle-CI

mrocklin:
In particular, I've found that the backlog on travis-ci can grow somewhat large, causing delays

hameerabbasi:
Well, mainly the lack of a Travis-like "matrix" interface. I've gone through every possibility, and the only way to do something like that is to write separate CI code for every combination. I've asked around in the CircleCI Forum and other similar places as well.

rgommers:
that's my experience too, TravisCI is still nicer if you have multiple builds.

PR #169, opened 2018-07-20
MAINT: make small floating point values zero

stsievert:


hameerabbasi:
This should be done externally to here, before feeding these values into `from_numpy`. We want to maintain full equivalence to a NumPy array.

hameerabbasi:
For example, you could have values that have a small magnitude in the entire array (by design). ALL of them are `O(10 ** -20)`, but some are exactly zero. This would cause issues with that.

hameerabbasi:
Is there actually a way you can't do this outside pydata/sparse, wherever `from_numpy is called`? This really seems like something that shouldn't be in here. I would be interested to know use cases where this couldn't be done outside here.

If however, there were prominent operations with sparse arrays that produced "almost zeros", I might be for it. But in this case it seems it's fairly easy to do this outside pydata/sparse, and introducing it here doesn't add any significant value.

stsievert:
This is super easy to do in pydata/sparse. It seemed fundamental enough to merge it upstream.

mrocklin:
I'm curious, what does scipy.sparse do here?  Do they have anything similar?

hameerabbasi:
Can you give me an example of where it's actually needed? The use-case doesn't seem clear to me. If something is producing `ndarray`s, it's unlilkely to be an output of this library (other than via `dot` or `tensordot`), and it shouldn't be this library's job to deal with it.

IF there is a concrete use-case where some output from this library produces `ndarray`s that are "meant to be zero but not quite zero", then I'd be willing to move forward with this.

sparse/coo/core.py
```diff
@@ -278,6 +280,9 @@ def from_numpy(cls, x):
         """
         x = np.asanyarray(x)
         if x.shape:
+            if approx and x.dtype and x.dtype.kind == 'f':
+                i = np.abs(x) < np.finfo(x.dtype).resolution * 10
+                x[i] = 0
```
hameerabbasi: We're mutating the input array here. Not good.

sparse/coo/core.py
```diff
@@ -278,6 +280,9 @@ def from_numpy(cls, x):
         """
         x = np.asanyarray(x)
         if x.shape:
+            if approx and x.dtype and x.dtype.kind == 'f':
+                i = np.abs(x) < np.finfo(x.dtype).resolution * 10
```
hameerabbasi: I'd compare the value with the arithmetic mean of the entire array multiplied by `np.finfo(x.dtype).resolution * 10`. We don't necessarily know the range of the inputs here.

sparse/tests/test_coo.py
```diff
@@ -471,6 +471,14 @@ def test_trinary_broadcasting_pathological(shapes, func, value, fraction):
     assert_eq(fs, func(*dense_args), equal_nan=True)
 
 
+def test_almost_zero_float():
+    x = np.array([1e-20, 1e-15, 1e-14], dtype=np.float16)
```
hameerabbasi: As above, it'd be nice to add tests of different ranges of inputs. Inputs smaller than the "rest of the array" should be cut off.

hameerabbasi:
> I'm curious, what does scipy.sparse do here? Do they have anything similar?

I couldn't find anything in the API, and a quick test in Jupyter Lab says no.

```python
>>> import scipy.sparse as sps
>>> import numpy as np
>>> x = np.array([1.0, 1.0e-100])
>>> s = sps.csc_matrix(x)
>>> s.nnz
0
```

stsievert:
I'm seeing the same thing with scipy.sparse. I'm not seeing any keyword args in `scipy.sparse.*_matrix` that mirror this behavior.

I need to look into this more to see if my use case. I'll close this in the meantime.

Issue #164, opened 2018-07-03
Store linearized indices instead of full coordinates


hameerabbasi:
I was thinking of replacing the full coordinates (`coords`) with just a 1-D array of linearised indices.

Pros:

- More memory efficient (this gives back a lot of benefits of #158)
- Some O(N) operations become O(1) such as `reshape`, `flatnonzero`.
- COO becomes a special case of the [GCRS/GCCS](https://ieeexplore.ieee.org/document/7237032/) scheme, which means we can drop CSD and go with GCRS/GCCS, but with the difference that any collection of axes (not just rows/columns) can be compressed.
- CSR/CSC are represented exactly as in SciPy/other libraries

Cons:

- Adds a small (my intuition says almost negligible) performance penalty in many operations.
- Some O(1) operations become O(N) such as `nonzero`, the one-argument version of `where`.

cc @mrocklin, @stefanv

ahwillia:
Would it be tenable to support both indexing schemes and give users the
option to choose?

hameerabbasi:
> Would it be tenable to support both indexing schemes and give users the option to choose?

`COO` arrays will be indexed exactly as before. This is about storage, not indexing.

ahwillia:
Yes but if there are performance tradeoffs then it might be nice to support both. Though maybe those tradeoffs are negligible?

hameerabbasi:
Yes, the performance loss in most cases is negligible according to my
intuition. But keeping both incurs technical debt, or at least maintenance
debt.

Sent from Astro <https://www.helloastro.com> for iOS


On 3. Jul 2018 at 17:51, Alex Williams <notifications@github.com> wrote:


Yes but if there are performance tradeoffs then it might be nice to support
both. Though maybe those tradeoffs are negligible?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/pydata/sparse/issues/164#issuecomment-402204960>, or mute
the thread

perimosocordiae:
FWIW, this is the approach I used for sparray: https://github.com/perimosocordiae/sparray/blob/master/sparray/flat.py#L20

I found it to be quite convenient, though it does mean you may need more integer bits to represent the flat indices.

hameerabbasi:
@perimosocordiae Currently, we use `np.intp` anyway, but due to the way things are designed, the maximum supported size is `2 ** 64 - 1` anyway, and I predict it might decrease to `2 ** 63 - 1`.

[...]

PR #171, opened 2018-07-24
MAINT: better error messages for initialization

stsievert:
During initialization, I kept on running into an assert. This PR resolves that and provides a better error message.

Why is `coords` in the shape it is? I would expect `coords.T` to be passed to `__init__`. That is, instead of `coords.shape == (ndim, num_points)`, I would expect `coords.shape == (num_points, ndim)`.

### DIFF
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -217,8 +217,19 @@ class COO(SparseArray, NDArrayOperatorsMixin):
 
         super(COO, self).__init__(shape, fill_value=fill_value)
         self.coords = self.coords.astype(np.intp)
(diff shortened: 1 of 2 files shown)


hameerabbasi:
Thanks, @stsievert! This looks great. Yeah, `coords` will probably become 1-D and hold just the "unwrapped" or "linearized 1-D index" soon (#164), so your question may not matter much. As to why, I'd guess it's because that's how it is in `scipy.sparse.coo_matrix`. It has `row` and `col` or something similar IIRC, often passed in as `(row, col)`, and they're 1-D arrays, so here we also have the "last dimension" representing the actual coordinate and the first representing the dimension.

PR #181, opened 2018-09-11
tweaked reduce method to return COO with correct fill_value

argearriojas:
This is the error that this pull request fixes
```
>>> A = sparse.random((10, 10), density=0.1)
>>> A
<COO: shape=(10, 10), dtype=float64, nnz=10, fill_value=0.0>
>>> B = 1-A
>>> B
<COO: shape=(10, 10), dtype=float64, nnz=10, fill_value=1.0>
>>> B.prod(axis=0)
<COO: shape=(10,), dtype=float64, nnz=8, fill_value=0.0>
```
The `fill_value` should remain `fill_value=1.0` for the object to remain correct.

### DIFF
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -691,7 +691,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):
             result = result[mask]
 
             a = COO(coords, result, shape=(a.shape[0],),
-                    has_duplicates=False, sorted=True)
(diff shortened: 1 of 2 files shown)


argearriojas:
Hello! I included a new test. I hope I dropped it on the right place at that test file.

BTW, Nice project!

hameerabbasi:
+1, will merge once CI passes.

PR #183, opened 2018-09-17
Add eye, zeros, zeros_like, ones, ones_like

jcrist:
Add implementations of `eye`, `zeros`, and `zeros_like`, mirroring the equivalent numpy api.

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -762,9 +762,10 @@ def eye(N, M=None, k=0, dtype=float):
         n_coords = m_coords = np.arange(data_length, dtype=np.intp)
 
     coords = np.stack([n_coords, m_coords])
-    data = np.ones(data_length, dtype=dtype)
+    data = np.array(1, dtype=dtype)
 
-    return COO(coords, data=data, shape=(N, M), has_duplicates=False)
+    return COO(coords, data=data, shape=(N, M), has_duplicates=False,
+               sorted=True)
 
 
 def zeros(shape, dtype=float):
@@ -798,7 +799,8 @@ def zeros(shape, dtype=float):
         shape = (shape,)
     data = np.empty(0, dtype=dtype)
     coords = np.empty((len(shape), 0), dtype=np.intp)
-    return COO(coords, data=data, shape=shape, has_duplicates=False)
+    return COO(coords, data=data, shape=shape, has_duplicates=False,
+               sorted=True)
 
 
 def zeros_like(a, dtype=None):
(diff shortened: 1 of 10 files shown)


mrocklin:
Looks like docstrings need to be spaced slightly differently.  Maybe a difference in Numpy version?

jcrist:
Grumble. Since `NORMALIZE_WHITESPACE` only treats whitespace variations as equivalent (and not equivalent with the empty string), there's no way to AFAICT to get get the doctests to work with multiple versions of numpy if they differ in output between

```
np.array([ 0.0, 0.0])  # and
np.array([0.0, 0.0])
```

Seems to be only an issue with float arrays, just skipped those tests.

rgommers:
The normal approach is to only run doctests for the latest version of numpy. With all the print changes around numpy 1.14, that's the only reasonable thing to do in CI.

sparse/coo/common.py
```diff
@@ -705,3 +705,122 @@ def roll(a, shift, axis=None):
...
+        data_length = max(min(data_length, N + k), 0)
+        m_coords = np.arange(data_length, dtype=np.intp)
+        n_coords = m_coords - k
+    else:
+        n_coords = m_coords = np.arange(data_length, dtype=np.intp)
+
+    coords = np.stack([n_coords, m_coords])
+    data = np.ones(data_length, dtype=dtype)
+
+    return COO(coords, data=data, shape=(N, M), has_duplicates=False)
```
hameerabbasi: Might be nice to add `sorted=True`.

sparse/coo/common.py
```diff
@@ -705,3 +705,122 @@ def roll(a, shift, axis=None):
...
+    array([[0, 0],
+           [0, 0]])
+    """
+    from .core import COO
+
+    if not isinstance(shape, tuple):
+        shape = (shape,)
+    data = np.empty(0, dtype=dtype)
+    coords = np.empty((len(shape), 0), dtype=np.intp)
+    return COO(coords, data=data, shape=shape, has_duplicates=False)
```
hameerabbasi: `sorted=True`

sparse/coo/common.py
```diff
@@ -705,3 +705,122 @@ def roll(a, shift, axis=None):
...
+    out : COO
+        Array of zeros with the same shape and type as `a`.
+
+    Examples
+    --------
+    >>> x = np.ones((2, 3), dtype='i8')
+    >>> zeros_like(x).todense()  # doctest: +NORMALIZE_WHITESPACE
+    array([[0, 0, 0],
+           [0, 0, 0]])
+    """
```
hameerabbasi: `sorted=True`.

sparse/coo/common.py
```diff
@@ -705,3 +705,122 @@ def roll(a, shift, axis=None):
...
+        m_coords = n_coords + k
+    elif k < 0:
+        data_length = max(min(data_length, N + k), 0)
+        m_coords = np.arange(data_length, dtype=np.intp)
+        n_coords = m_coords - k
+    else:
+        n_coords = m_coords = np.arange(data_length, dtype=np.intp)
+
+    coords = np.stack([n_coords, m_coords])
+    data = np.ones(data_length, dtype=dtype)
```
hameerabbasi: Data as a scalar should work here. Takes less memory due to scalar broadcasting being a view.

mrocklin:
That's what we do in dask/dask.  This is probably Hameer's call though.

jcrist:
Thanks for the review, I believe all comments have been addressed.

hameerabbasi:
Thabks for making these changes! They’re very welcome. Merging now.

PR #184, opened 2018-09-17
Fixup pickle, add copy

jcrist:
- Fixes pickle of `COO` to exclude the `_cache` attribute
- Adds a `COO.copy` method to mirror numpy.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* :feature:`184` Added :code:`COO.copy` method, and changed pickle of :object:`COO` to not include its cache.
 * :feature:`183` Added :code:`sparse.eye`, :code:`sparse.zeros`,
   :code:`sparse.zeros_like`, :code:`sparse.ones`, and :code:`sparse.ones_like`.
 * :release:`0.4.1 <2018-09-12>`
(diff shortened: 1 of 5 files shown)


jcrist:
I've been avoiding adding things to the changelog until approved because every merged PR results in merge conflicts for existing PRs. Other repos solve this by:

- Adding placeholder list elements to be edited later (git merge works better when editing a list element than adding a new line)
- Holding off on adding changelog entries until release
- Adding a separate tiny file in a directory for each new change, and merging these together on release
- Inserting items in the changelog rather than prepending, minimizing chances for conflicts over time.

Unfortunately none of these work with the sphinx releases extension you're using to generate a running changelog. I'm not sure if anything should be done to change this.

hameerabbasi:
Merging, thanks.

dhirschfeld:
It can be very painful for contributors to have to deal with CHANGELOG merge conflicts.

I've also seen it done where an issue `CHANGELOG - 0.4.2` is created and the changelog entries are accumulated there and only moved to the actual changelog file when the when the release is made.

mrocklin:
FWIW in Dask we stopped asking people to provide changelog entries on every
PR.  I tend to add these all afterwards when we go to release.  This seems
to be smoother overall.

hameerabbasi:
That seems like a good way to go, but you lose out on "what's coming next" for the uninitiated.

PR #185, opened 2018-09-18
Clip

jcrist:
Add `clip`, mirroring numpy api.


hameerabbasi:
Merging, thanks for all the awesome work, @jcrist!

Issue #163, opened 2018-06-28
Move to a more robust unit testing model


hameerabbasi:
I was thinking of moving the tests to a more robust unit test model. It'll take:

- The `dtype`, `shape` of each input array.
- The function name to test (both sparse and NumPy/SciPy) or callable.
- Where to place the arguments.
- Any additional arguments to the function.

It will:

- Generate the random arrays.
- Perform the operation on both arrays.
- Compare/assert.

This way, tests just reduce to "stub classes" instead of repetitions.

cc: @mrocklin

rgommers:
Sounds like that could be helpful. The one thing to check is designing it in such a way that the tracebacks are still as understandable (exactly which set of arguments is failing)

PR #190, opened 2018-09-19
Add COO.mean and sparse.nanmean

jcrist:
Adds `COO.mean` and `sparse.nanmean`, matching the numpy api

Also applies several cleanups to the reductions test suite:
- Test with larger arrays on numeric reductions. Some issues with current functions only showed up on larger arrays (before there was a decent chance only one element along an axis was nonzero).
- Test with more input dtypes
- XFail tests setting `dtype=float16`. These fail for some, but not all configurations. It's not clear how numpy interprets this for various reductions - neither casting prior to summing or casting after the sum give the same result as numpy depending on configuration. Since `COO.sum` was already failing on larger arrays, I xfailed these tests for now.
- Cleanup superfluous warnings, reducing pytest output
- A few assorted cleanups.

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -466,9 +466,7 @@ def nanmean(x, axis=None, keepdims=False, dtype=None, out=None):
         axis = tuple(range(x.ndim))
     elif not isinstance(axis, tuple):
         axis = (axis,)
-    den = 1
-    for ax in axis:
-        den *= x.shape[ax]
+    den = reduce(operator.mul, (x.shape[i] for i in axis), 1)
     den -= nancount
 
     if np.any(den == 0):
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -426,6 +426,58 @@ def nansum(x, axis=None, keepdims=False, dtype=None, out=None):
     return nanreduce(x, np.add, axis=axis, keepdims=keepdims, dtype=dtype)
 
 
+def nanmean(x, axis=None, keepdims=False, dtype=None, out=None):
+    """
+    Performs a ``NaN`` skipping mean operation along the given axes. Uses all axes by default.
+
+    Parameters
+    ----------
+    x : SparseArray
+        The array to perform the reduction on.
+    axis : Union[int, Iterable[int]], optional
+        The axes along which to compute the mean. Uses all axes by default.
+    keepdims : bool, optional
+        Whether or not to keep the dimensions of the original array.
+    dtype: numpy.dtype
+        The data type of the output array.
+
+    Returns
+    -------
+    COO
+        The reduced output sparse array.
+
+    See Also
+    --------
+    :obj:`COO.mean` : Function without ``NaN`` skipping.
+    numpy.nanmean : Equivalent Numpy function.
+    """
+    assert out is None
+    x = asCOO(x, name='nanmean')
+
+    if not np.issubdtype(x.dtype, np.floating):
+        return x.mean(axis=axis, keepdims=keepdims, dtype=dtype)
+
+    mask = np.isnan(x)
+    x2 = where(mask, 0, x)
+
+    # Count the number non-nan elements along axis
+    nancount = mask.sum(axis=axis, dtype='i8', keepdims=keepdims)
+    if axis is None:
+        axis = tuple(range(x.ndim))
+    elif not isinstance(axis, tuple):
+        axis = (axis,)
+    den = 1
+    for ax in axis:
+        den *= x.shape[ax]
+    den -= nancount
+
+    num = np.sum(x2, axis=axis, dtype=dtype, keepdims=keepdims)
+
+    if num.ndim:
+        return np.true_divide(num, den, casting='unsafe')
+    return (num / den).astype(dtype)
+
+
 def nanmax(x, axis=None, keepdims=False, dtype=None, out=None):
     """
     Maximize along the given axes, skipping ``NaN`` values. Uses all axes by default.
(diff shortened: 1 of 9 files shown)


jcrist:
cc @hameerabbasi for review.

sparse/coo/common.py
```diff
@@ -426,6 +426,62 @@ def nansum(x, axis=None, keepdims=False, dtype=None, out=None):
...
+
+    # Count the number non-nan elements along axis
+    nancount = mask.sum(axis=axis, dtype='i8', keepdims=keepdims)
+    if axis is None:
+        axis = tuple(range(x.ndim))
+    elif not isinstance(axis, tuple):
+        axis = (axis,)
+    den = 1
+    for ax in axis:
+        den *= x.shape[ax]
```
hameerabbasi: This can probably be written more compactly with [`functools.reduce`](https://docs.python.org/3.7/library/functools.html#functools.reduce) and a generator.

sparse/coo/core.py
```diff
@@ -1076,6 +1076,93 @@ def prod(self, axis=None, keepdims=False, dtype=None, out=None):
...
+
+        >>> s.mean()
+        0.5
+        """
+        if axis is None:
+            axis = tuple(range(self.ndim))
+        elif not isinstance(axis, tuple):
+            axis = (axis,)
+        den = 1
+        for ax in axis:
```
hameerabbasi: This can probably be written more compactly with [`functools.reduce`](https://docs.python.org/3.7/library/functools.html#functools.reduce) and a generator.

sparse/tests/test_coo.py
```diff
@@ -11,52 +11,88 @@
...
 
-@pytest.mark.parametrize('reduction,kwargs,eqkwargs', [
-    ('any', {}, {}),
-    ('all', {}, {}),
+@pytest.mark.xfail(reason=('Setting output dtype=float16 produces results '
+                           'inconsistent with numpy'))
+@pytest.mark.filterwarnings('ignore:overflow')
+@pytest.mark.parametrize('reduction, kwargs', [
+    ('sum', {'dtype': np.float16}),
+    ('mean', {'dtype': np.float16}),
```
hameerabbasi: Optional: The reason we had `eqkwargs` for `float16` was large floating-point inaccuracies with `float16`. We're basically using [`np.allclose`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html) to verify equivalence, and the `kwargs` in `assert_eq` are passed along to that.

It might be nice to rewrite this at some point, but that's how it is now. There is a plan to re-write this, basically something that takes two callables and some shapes and compares the NumPy and sparse functions (see #163).

Anyway, back to the point: If the `rtol` and `atol` parameters are increased, I'm pretty sure this will pass as well.

jcrist: > Anyway, back to the point: If the rtol and atol parameters are increased, I'm pretty sure this will pass as well.

This may be true, but the errors have to be pretty large. With `atol=10, rtol=10`, you still get errors for `sum`. This is an algorithmic difference, not just floating point error.

hameerabbasi: > With `atol=10, rtol=10`, you still get errors for `sum`.

In that case, feel free to leave this as-is for now (and if you want, raise an issue documenting this).

sparse/coo/common.py
```diff
@@ -426,6 +426,62 @@ def nansum(x, axis=None, keepdims=False, dtype=None, out=None):
...
+    COO
+        The reduced output sparse array.
+
+    See Also
+    --------
+    :obj:`COO.mean` : Function without ``NaN`` skipping.
+    numpy.nanmean : Equivalent Numpy function.
+    """
+    assert out is None
+    x = asCOO(x, name='nanmean')
```
hameerabbasi: I'm surprised this works -- `asCOO` has no `name` parameter. I must have forgotten to check for superfluous `kwargs` somewhere.

jcrist: Yes it does, `asCOO` is different from `as_coo` in core.py. See the top of this file.

sparse/coo/common.py
```diff
@@ -426,6 +426,60 @@ def nansum(x, axis=None, keepdims=False, dtype=None, out=None):
...
+
+    if not np.issubdtype(x.dtype, np.floating):
+        return x.mean(axis=axis, keepdims=keepdims, dtype=dtype)
```
hameerabbasi: We need a test for this branch, apparently.

sparse/tests/test_coo.py
```diff
@@ -85,6 +125,7 @@ def test_ufunc_reductions_kwargs(reduction, kwargs):
 @pytest.mark.parametrize('keepdims', [False])
 @pytest.mark.parametrize('fraction', [0.25, 0.5, 0.75, 1.0])
 @pytest.mark.filterwarnings('ignore:All-NaN')
+@pytest.mark.filterwarnings('ignore:Mean of empty slice')
 def test_nan_reductions(reduction, axis, keepdims, fraction):
     s = sparse.random((2, 3, 4), data_rvs=random_value_array(np.nan, fraction),
```
hameerabbasi: It seems you forgot to change to the fixture here.

jcrist: I don't think so? `nanmean` throws that warning instead of `All-NaN`, we need both filters to cover all the warnings from this test. Unless I'm misunderstanding your comment here.

hameerabbasi: Ah, my bad. The fixture doesn't do NaNs. Ignore. It would be nice to and a non-floating `dtype` here to hit the line we missed -- But not is okay too.

hameerabbasi: I meant that the fixture you created wasn't used in the NaN-reductions, but like I said, you can safely ignore me.

jcrist:
Thanks @hameerabbasi!

PR #193, opened 2018-09-19
Add order parameter to `COO.reshape`

jcrist:
Adds support for `np.reshape` to dispatch to `COO.reshape` properly.

### DIFF
--- a/sparse/tests/test_coo.py
+++ b/sparse/tests/test_coo.py
@@ -229,10 +229,6 @@ def test_reshape_function():
     assert isinstance(s2, COO)
     assert_eq(s2, x.reshape(shape))
 
-    # order parameter not actually supported
-    with pytest.raises(NotImplementedError):
-        np.reshape(s, shape, order='F')
-
 
 def test_to_scipy_sparse():
     s = sparse.random((3, 5), density=0.5)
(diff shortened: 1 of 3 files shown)


sparse/tests/test_coo.py
```diff
@@ -220,6 +220,20 @@ def test_reshape_same():
...
+    x = s.todense()
+    shape = (3, 5)
+
+    s2 = np.reshape(s, shape)
+    assert isinstance(s2, COO)
+    assert_eq(s2, x.reshape(shape))
+
+    # order parameter not actually supported
+    with pytest.raises(NotImplementedError):
+        np.reshape(s, shape, order='F')
```
hameerabbasi: We might want to remove these two lines from the test, if we add this functionality later it'll crap out.

hameerabbasi: An exception from coverage is already in place.

jcrist: Done.

I personally prefer to check these errors - in the past I've seen code that was supposed to raise a `DeprecationWarning` or `NotImplementedError` fail to actually raise due to a bug. I don't think of the tests as the-one-uneditable-source-of-truth, I think someone adding this functionality wouldn't be surprised to see a test fail due to a `NotImplementedError` not being raised anymore, and be fine changing the test.

Anyway, happy to cede to you here. I believe this should be good to go.

hameerabbasi: Fair enough. 👍

PR #194, opened 2018-09-20
Add sparse.kron

jcrist:
Added `sparse.kron` implementation.

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -212,6 +212,65 @@ def _dot(a, b):
     return aa.dot(b)
 
 
+def kron(a, b):
+    """Kronecker product of 2 sparse arrays.
+
+    Parameters
+    ----------
+    a, b : SparseArray, scipy.sparse.spmatrix, or np.ndarray
+        The arrays over which to compute the Kronecker product.
+
+    Returns
+    -------
+    res : COO
+        The kronecker product
+
+    Notes
+    -----
+    Currently only arrays with ndim <= 2 are supported.
+
+    Examples
+    --------
+    >>> a = eye(3, dtype='i8')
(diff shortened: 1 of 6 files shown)


sparse/coo/common.py
```diff
@@ -212,6 +212,65 @@ def _dot(a, b):
     return aa.dot(b)
 
 
+def kron(a, b):
```
jcrist: What is the approach in sparse when coercing inputs? If neither input is a sparse array, do you return a dense result, or coerce to sparse?

hameerabbasi: We generally raise a `ValueError` if the result is guaranteed to be dense.

hameerabbasi: The `dot` behaviour is weird, it should work on mixed or sparse only, but not on dense-dense. To answer your question: In general, we only accept dense inputs if mixed sparse-dense operations are more efficient than a dense-only one. If the dense-only operation is the best one available, we err.

For `dot`, there is an efficient implementation of sparse-dense and dense-sparse multiplication, as in this case. For element-wise operations, this is sometimes the case (think `operator.mul`), but not always (think `operator.add`).

This case is similar: A mixed operation could benefit from the sparse structure, but not a dense-only one.

jcrist:
> If this is really urgent, go ahead and merge it after these changes. I'm working on a version of kron using that will work on arbitrary dimensionality, and that one will supersede this one.

Glad to hear it! I figured there'd be a more efficient/elegant way to express this in ndimensions, but my immediate need was for a 2 dimensions only and using the scipy sparse implementation was easy. I look forward to seeing your implementation.

I believe all concerns have been addressed. Thanks for the review, merging.

PR #195, opened 2018-09-20
Generalise kron to an arbitrary number of dimensions.

hameerabbasi:
Generalizes `kron` to an arbitrary number of dimensions.

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -228,7 +228,7 @@ def kron(a, b):
     Raises
     ------
     ValueError
-        If all arguments are dense.
+        If all arguments are dense or arguments have nonzero fill-values.
 
     Examples
     --------
@@ -242,6 +242,7 @@ def kron(a, b):
     """
     from .core import COO
     from .umath import _cartesian_product
+    check_zero_fill_value(a, b)
 
     a_sparse = isinstance(a, (SparseArray, scipy.sparse.spmatrix))
     b_sparse = isinstance(b, (SparseArray, scipy.sparse.spmatrix))
@@ -271,7 +272,7 @@ def kron(a, b):
     o_data = a.data[a_idx] * b.data[b_idx]
     o_shape = tuple(i * j for i, j in zip(a.shape, b.shape))
 
-    return COO(o_coords, o_data, shape=o_shape, has_duplicates=False, fill_value=a.fill_value*b.fill_value)
+    return COO(o_coords, o_data, shape=o_shape, has_duplicates=False)
 
 
 def concatenate(arrays, axis=0):
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -228,7 +228,7 @@ def kron(a, b):
     Raises
     ------
     ValueError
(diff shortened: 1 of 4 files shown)


hameerabbasi:
cc @jcrist for review.

sparse/tests/test_coo.py
```diff
@@ -334,10 +334,10 @@ def test_kron(a_ndim, b_ndim):
         assert_eq(sparse.kron(a, b), sol)
 
 
-@pytest.mark.parametrize('ndim', [0, 1, 2])
```
hameerabbasi: This was, again, a case better handled in the dense regime, so I removed it.

sparse/tests/test_coo.py
```diff
@@ -314,11 +314,11 @@ def test_dot_nocoercion():
...
 
-@pytest.mark.parametrize('a_ndim', [1, 2])
-@pytest.mark.parametrize('b_ndim', [1, 2])
+@pytest.mark.parametrize('a_ndim', [1, 2, 3])
+@pytest.mark.parametrize('b_ndim', [1, 2, 3])
 def test_kron(a_ndim, b_ndim):
-    a_shape = (10, 11)[:a_ndim]
-    b_shape = (12, 13)[:b_ndim]
+    a_shape = (3, 4, 5)[:a_ndim]
+    b_shape = (4, 5, 3)[:b_ndim]
```
jcrist: It'd be maybe good to test:
- Using dimensions that don't ever match up (`len(set(a_shape).intersection(b_shape)) == 0`) to ensure no accidental reliance on broadcasting aligning. Perhaps `(3, 4, 5)`, and `(6, 7, 8)`.
- Cases with a nonzero fill value

sparse/coo/common.py
```diff
@@ -244,44 +240,38 @@ def kron(a, b):
...
     """
-    from .core import COO, as_coo
-    check_zero_fill_value(a, b)
+    from .core import COO
+    from .umath import _cartesian_product
 
-    a_sparse = isinstance(a, SparseArray)
-    b_sparse = isinstance(b, SparseArray)
+    a_sparse = isinstance(a, (SparseArray, scipy.sparse.spmatrix))
+    b_sparse = isinstance(b, (SparseArray, scipy.sparse.spmatrix))
```
jcrist: We don't ever test with a `scipy.sparse.spmatrix`, is this something that should also be added to the tests?

sparse/tests/test_coo.py
jcrist: The first one might be being overly paranoid though, feel free to ignore.

hameerabbasi: Ah, yes. My algorithm only works with a zero fill-value.

sparse/coo/common.py
hameerabbasi: Added a test.

hameerabbasi: For future references, `asCOO` and the `COO` constructor are pretty heavily tested, so we probably don't need to test things like these everywhere.

PR #196, opened 2018-10-02
Add COO.imag, COO.real, COO.conj()

jcrist:
Add methods and properties to support complext sparse arrays, matching the numpy api.


sparse/coo/core.py
```diff
@@ -1288,6 +1288,51 @@ def T(self):
         """
         return self.transpose(tuple(range(self.ndim))[::-1])
 
+    @property
+    def real(self):
+        """The real part of the array.
+
+        Examples
```
hameerabbasi: A "See also" section would be nice, linking to `np.real` or `np.ndarray.real`.

sparse/coo/core.py
```diff
@@ -1288,6 +1288,51 @@ def T(self):
...
+        array([1., 0.])
+        >>> x.real.dtype
+        dtype('float64')
+        """
+        return self.__array_ufunc__(np.real, '__call__', self)
+
+    @property
+    def imag(self):
+        """The imaginary part of the array.
+
```
hameerabbasi: Ditto for see also.

sparse/coo/core.py
```diff
@@ -1288,6 +1288,51 @@ def T(self):
...
+        dtype('float64')
+        """
+        return self.__array_ufunc__(np.imag, '__call__', self)
+
+    def conj(self):
+        """Return the complex conjugate, element-wise.
+
+        The complex conjugate of a complex number is obtained by changing the
+        sign of its imaginary part.
+
```
hameerabbasi: Ditto for see also.

sparse/tests/test_coo.py
```diff
@@ -1959,3 +1959,15 @@ def test_full_like():
     x = np.zeros((5, 5), dtype='i8')
     assert_eq(sparse.full_like(x, 9.5), np.full_like(x, 9.5))
     assert_eq(sparse.full_like(x, 9.5, dtype='f8'), np.full_like(x, 9.5, dtype='f8'))
+
+
+@pytest.mark.parametrize('complex', [True, False])
```
hameerabbasi: Nit: I usually like further parametrizations so I know exactly what's failing. Maybe something like:

```python
@pytest.mark.parametrize('method', [
    lambda x: x.real,
    lambda x: x.imag,
    lambda x: x.conj()
])
```

PR #208, opened 2018-11-12
Fix overflow in reduce.

### LINKED ISSUE #207: Overflow on reshape
Hey. Sorry if this is trivial: the array shape appears to be stored as a tuple of int values, not long, and so the reshaping function (and presumably others) encounters problems for even relatively small arrays:

```
import sparse
sparse.COO([], shape = (10,1000000,1000000)).sum(0)
```

Yields an error, originating from

```
# TODO: this self.size enforces a 2**64 limit to array size
linear_loc = self.linear_loc()

coords = np.empty((len(shape), self.nnz), dtype=np.intp)
strides = 1
for i, d in enumerate(shape[::-1]):
	coords[-(i + 1), :] = (linear_loc // strides) % d
	strides *= d

```

where `strides` overflows. You can simply cast this as `strides = 1L` instead but it's hard to imagine there isn't other errors introduced by this. Other reshapes may only raise a warning, as the overflow gives a positive. A comment in the offending function reads `# TODO: this self.size enforces a 2**64 limit to array size`, but the size is currently limited to 2**32. In the above case

```2**32 < 1,000,000**2 < 2**64```

I should add this happens on Windows 10 with Python versions 2.7 and 3.7.

hameerabbasi:
Fixes #207

### DIFF
--- a/sparse/compatibility.py
+++ b/sparse/compatibility.py
@@ -14,5 +14,5 @@ else:
     from itertools import izip_longest as zip_longest
     from itertools import izip as zip
 
-    int = long
+    int = int
     range = xrange
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -7,7 +7,7 @@ import numpy as np
 import scipy.sparse
 
 from ..sparse_array import SparseArray
-from ..compatibility import range
+from ..compatibility import range, int
 from ..utils import isscalar, normalize_axis, check_zero_fill_value, check_consistent_fill_value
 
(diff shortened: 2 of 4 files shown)


hameerabbasi:
@mrocklin Reaching out to you for advice, any way to fix and/or ignore the Python 2 doctests?

mrocklin:
Hi @hameerabbasi .  My recommendation would be to just not test doctests in Python 2 and be done with it

PR #209, opened 2018-11-19
Fix sorted disclaimer.

ShadenSmith:


hameerabbasi:
Thanks, looks like that was a typo!

PR #210, opened 2018-11-19
Configure densification.

hameerabbasi:
Add an environment variable to allow configuration of densification.

### DIFF
--- a/sparse/__init__.py
+++ b/sparse/__init__.py
@@ -7,3 +7,9 @@ from .io import save_npz, load_npz
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+import os
+
+_AUTO_DENSIFICATION_ENABLED = bool(int(os.environ.get('SPARSE_AUTO_DENSIFY', '1')))
+
+del os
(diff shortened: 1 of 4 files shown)


sparse/__init__.py
```diff
@@ -7,3 +7,9 @@
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+import os
+
+_AUTO_DENSIFICATION_ENABLED = bool(int(os.environ.get('SPARSE_AUTO_DENSIFY', '1')))
+
```
asmeurer: I would also add

```py
def set_auto_densify(flag):
    global _AUTO_DENSIFICATION_ENABLED
    _AUTO_DENSIFICATION_ENABLED = flag
```

hameerabbasi: The problem is that that isn't thread safe. For example, if you do it with Dask or whatever else, it can blow stuff up.

An environment variable loaded at import time is guaranteed to be thread safe.

asmeurer: Can we add a warning to the function that it isn't thread safe. I think if it is only called right after import it should be fine too.

rgommers: > unfortunately there's no clear way to make it user facing but not library facing.

That's in the same category as `numpy.set_printoptions`, `np.seterr`, private class attributes, etc. - just don't mess with it as a library. It doesn't happen a lot, and if it does the first person who notices files a bug against that library.

PR #211, opened 2018-11-21
Add logo.

hameerabbasi:


mrocklin:
Whoa, very cool.  Who made the logo?

Also, slight style feedback, it looks like it's taking up a bit of screen space on the documentation.

image

hameerabbasi:
The design was my idea but the logo itself was made by Summer Froisland over at Quansight. :-)

I like it better this size (20em width), but I can make it smaller. What size do you suggest?

mrocklin:
I think that the challenge here is that it takes up enough vertical space so that initial readers will likely need to scroll to get to any meaningful content.  There are probably a couple options here:

1.  On the main content pane (with the white background) maybe the logo could be aligned to the right so that text wraps around it.
2.  Alternatively maybe you drop the logo from the main content pane because it's already visible on the left sidebar and there isn't much value in having it twice.  
3.  You might consider having another version of the logo that drops the top orange and bottom blue squares for use in vertically space challenged environments, leaving only the central sparse cube.  This could be something that you use all the time, or only in certain situations.

mrocklin:
Someone familiar with web design and layout might have more thoughts here.  This is not my area of expertise.

hameerabbasi:
@rgommers, I assumed you (or PyData) would be okay with this, I can delete the commit if that's not the case.

rgommers:
> @rgommers, I assumed you (or PyData) would be okay with this, I can delete the commit if that's not the case.

It seems fine to me. I'm not in any capacity formally responsible for NumFOCUS/PyData stuff though, so I'll forward the question - @lsilen, @aterrel please let us know if you see a problem with using the logo in the screenshot above for the `sparse` project (which is hosted under the PyData Github org, and affiliated with SciPy).

aterrel:
@rgommers I think it looks awesome and thanks for checking with us.

We don't have strict guidelines on the usages of the PyData trademark and logo, but as long as sparse remains a part of the PyData ecosystem it seems appropriate. So don't go rewriting in lua ;)

PR #214, opened 2018-12-06
added fast escape for tensordot() when nnz == 0

scopatz:
This way we don't convert to CSR or CSC if we don't need to.

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -149,6 +149,11 @@ def tensordot(a, b, axes=2):
     newshape_b = (N2, -1)
     oldb = [bs[axis] for axis in notin]
 
+    # if either a or b is all zeros, we can just return a zeros COO
+    if ((isinstance(a, (COO, scipy.sparse.spmatrix)) and a.nnz == 0) or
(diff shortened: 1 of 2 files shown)


hameerabbasi:
I'm concerned about what this would do in the case of infinities and NaNs.

scopatz:
@hameerabbasi - do you have a suggestion for the NaN / inf issue.  I see a couple of possibilities:

1. change the `or` in the condition to an `and` so that we only early return if both arrays are zero.
2. run `isfinite(array).all()` on the array that has non-zero entries and only finish if True.

scopatz:
(2) has the potential to add a lot of runtime.

scopatz:
Also, sparse doesn't have `isfinite()` yet.

hameerabbasi:
> Needs a test case with a nonzero fill value.

We currently error on nonzero fill-values. It's possible to do it though.

> do you have a suggestion for the NaN / inf issue

We can do (1) but it would be too specific to be of any use.

> Also, sparse doesn't have isfinite() yet.

NumPy `ufunc`s work fine on sparse arrays.

hameerabbasi:
Superseded by #215

PR #215, opened 2018-12-10
[WIP] Add a custom-coded dot product.

hameerabbasi:
Add a custom-coded dot product. Done:

- [x] `COO`-`COO`
- [x] `COO`-`ndarray`
- [x] `ndarray`-`COO`
- [x] Documentation

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -1134,6 +1134,27 @@ def ones_like(a, dtype=None):
 
 
 def _memoize_dtype(f):
+    """
+    Memoizes a function taking in NumPy dtypes.
+
+    Parameters
+    ----------
+    f : Callable
+
+    Returns
+    -------
+    wrapped : Callable
+
+    Examples
+    --------
+    >>> def func(dt1):
+    ...     return object()
+    >>> func = _memoize_dtype(func)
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for putting this in @hameerabbasi! I have been trying it out with https://github.com/scopatz/minimal-failing-sparse-dot and it seems to fail no matter if it is a `COO @ COO` or `ndarray @ COO` or `COO @ ndarray`. At least in Jupyter, the kernel dies.

hameerabbasi:
@scopatz Thanks for the message. From the traceback you showed, you're still using the code from `master`, not this branch, since this branch hasn't been merged yet. The code in the traceback has been removed from this branch.

You'll need to clone the `custom-dot` branch in my repo and run `pip install` on it.

scopatz:
Oh I haven't pushed an updated minimal example repo. I have just playing around with a local install of this branch.

scopatz:
I am on `0.5.0+16.gdcc437c` locally.

scopatz:
Hmm actually, that repo might be bad example because in the worst case it could end up with the dot product having a `nnz == 674916321618`, which is about 643 GB for float64.

scopatz:
Yeah, I agree.  That is a breaking case, so please ignore.

PR #217, opened 2018-12-11
Fix wrong shape of matmul result when the first input is 2-D

### LINKED ISSUE #202: COO.__matmul__ gives the wrong shape of resultant array
Please see the following, COO.__matmul__ is not consistent with np.ndarray.__matmul__
```python
In [1]: import numpy as np

In [2]: import sparse

In [3]: npx = np.random.random((1,2,3,4))

In [4]: npy = np.random.random((1,2,4,3))

In [5]: npz = npx @ npy

In [6]: npz.shape
Out[6]: (1, 2, 3, 3)

In [7]: spx = sparse.random((1,2,3,4), 0.8)

In [8]: spy = sparse.random((1,2,4,3), 0.8)

In [9]: spz = spx @ spy

In [10]: type(spz)
Out[10]: numpy.ndarray

In [11]: spz.shape
Out[11]: (1, 2, 3, 1, 2, 3)
```

gongliyu:
When the first input is 2-D, #204 does not give the correct shape. This PR will Fix #202 in this case.

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -185,9 +185,15 @@ def matmul(a, b):
             (type(a), type(b)))
 
     # When one of the input is less than 2-d, it is equivalent to dot
-    if a.ndim <= 2 or b.ndim <= 2:
+    if b.ndim <= 2:
         return dot(a, b)
 
+    if a.ndim <= 2:
+        res = dot(a, b)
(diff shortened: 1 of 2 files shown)


hameerabbasi:
Thanks, @gongliyu! Merging.

PR #220, opened 2018-12-13
Disallow auto-densification.

### LINKED ISSUE #218: Disallow np.array(COO)
Currently, we have a mix of when to allow dense inputs and when not to allow them. Currently, we have the `__array__` protocol which converts the sparse array into a dense one when using `np.array(COO)` or similar.

This would be fine if we knew that we were densifying, but in practice, we really don't. Think of something that calls `np.asarray` internally, this would densify, and in many cases fill up memory and raise a `MemoryError`.

In practice, what we want most of the time is to disallow `np.[as]array(COO)` and provide an _explicit_ escape hatch in the form of `COO.todense()`.

Recently, I added an environment variable to control this, so this discussion is mostly about changing the default behaviour.

Thoughts, @rgommers, @shoyer, @mrocklin, @perimosocordiae?

hameerabbasi:
Fixes #218

### DIFF
--- a/docs/generated/sparse.SparseArray.rst
+++ b/docs/generated/sparse.SparseArray.rst
@@ -19,4 +19,5 @@ SparseArray
       :toctree:
 
       SparseArray.asformat
+      SparseArray.todense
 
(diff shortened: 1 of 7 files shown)


scopatz:
LGTM! Thanks @hameerabbasi!

PR #233, opened 2019-02-25
Fix collections-related DeprecationWarnings on import

nils-werner:
Currently, when using Python 3.7, one sees the following warning when importing `sparse`:

    .../sparse/sparse/coo/core.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
      from collections import Iterable, Iterator, Sized

This warning will become an error in Python 3.8, the upcoming version. So it's time to fix it.

### DIFF
--- a/sparse/coo/common.py
+++ b/sparse/coo/common.py
@@ -1,7 +1,10 @@
 from functools import reduce, wraps
 import operator
 import warnings
-from collections import Iterable
+try:
+    from collections.abc import Iterable
+except AttributeError:
+    from collections import Iterable
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


hameerabbasi:
The issue is that this doesn't work on Python 2.7. If we're willing to drop Python 2.7 support then we can do this.

nils-werner:
Of course it works in Python 2.7, I just have to catch the right exception ;-)

hameerabbasi:
Merging, thanks @nils-werner!

PR #238, opened 2019-03-11
Support reduce fill_value in specific cases.

### LINKED ISSUE #237: Support sum with fill value
```python
import sparse
import numpy as np
s = sparse.COO(np.eye(5))
s.sum()
# 5
```

```python
(s + 1).sum()
```

### Traceback

<details>

```python-traceback
ValueError                                Traceback (most recent call last)
<ipython-input-5-561034dfff3e> in <module>
----> 1 (s + 1).sum()

~/miniconda/envs/dev/lib/python3.7/site-packages/sparse/coo/core.py in sum(self, axis, keepdims, dtype, out)
    795         25
    796         """
--> 797         return np.add.reduce(self, out=out, axis=axis, keepdims=keepdims, dtype=dtype)
    798
    799     def max(self, axis=None, keepdims=False, out=None):

~/miniconda/envs/dev/lib/python3.7/site-packages/sparse/coo/core.py in __array_ufunc__(self, ufunc, method, *inputs, **kwargs)
   1441             result = elemwise(ufunc, *inputs, **kwargs)
   1442         elif method == 'reduce':
-> 1443             result = COO._reduce(ufunc, *inputs, **kwargs)
   1444         else:
   1445             return NotImplemented

~/miniconda/envs/dev/lib/python3.7/site-packages/sparse/coo/core.py in _reduce(method, *args, **kwargs)
    621             self = COO.from_scipy_sparse(self)
    622
--> 623         return self.reduce(method, **kwargs)
    624
    625     def reduce(self, method, axis=(0,), keepdims=False, **kwargs):

~/miniconda/envs/dev/lib/python3.7/site-packages/sparse/coo/core.py in reduce(self, method, axis, keepdims, **kwargs)
    693         if not equivalent(zero_reduce_result, self.fill_value):
    694             raise ValueError("Performing this reduction operation would produce "
--> 695                              "a dense result: %s" % str(method))
    696
    697         if axis is None:

ValueError: Performing this reduction operation would produce a dense result: <ufunc 'add'>
```

</details>

It looks like we're unable to compute sum when there is a fill value.

hameerabbasi:
Fixes #237

### DIFF
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -17,7 +17,7 @@ from ..utils import normalize_axis, equivalent, check_zero_fill_value, _zero_of_
 
 
 _reduce_super_ufunc = {
-    np.sum: np.multiply,
+    np.add: np.multiply,
     np.multiply: np.power,
 }
 
@@ -730,7 +730,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):
                 result[missing_counts] = method(result[missing_counts],
                                                 self.fill_value, **kwargs)
             else:
-                result = method(result, reduce_super_ufunc(self.fill_value, a.shape[1] - counts))
+                result = method(result, reduce_super_ufunc(self.fill_value, a.shape[1] - counts)).astype(result.dtype)
                 result_fill_value = reduce_super_ufunc(self.fill_value, a.shape[1])
             coords = a.coords[0:1, inv_idx]
 
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -734,13 +734,8 @@ class COO(SparseArray, NDArrayOperatorsMixin):
                 result_fill_value = reduce_super_ufunc(self.fill_value, a.shape[1])
             coords = a.coords[0:1, inv_idx]
 
-            # Filter out zeros
-            mask = ~equivalent(result, self.fill_value)
-            coords = coords[:, mask]
-            result = result[mask]
-
             a = COO(coords, result, shape=(a.shape[0],),
-                    has_duplicates=False, sorted=True, fill_value=result_fill_value)
+                    has_duplicates=False, sorted=True, prune=True, fill_value=result_fill_value)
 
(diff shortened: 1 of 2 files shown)


hameerabbasi:
@mrocklin I think are some weird overflow or dtype issues going on here... I'm not smart enough to recognize what they are.

sparse/coo/core.py
```diff
@@ -712,9 +722,16 @@ def reduce(self, method, axis=(0,), keepdims=False, **kwargs):
...
-            result[missing_counts] = method(result[missing_counts],
-                                            self.fill_value, **kwargs)
+
+            result_fill_value = self.fill_value
+
+            if reduce_super_ufunc is None:
+                missing_counts = counts != a.shape[1]
+                result[missing_counts] = method(result[missing_counts],
+                                                self.fill_value, **kwargs)
+            else:
```
hameerabbasi: This the the problematic "new" branch.

sparse/coo/core.py
```diff
@@ -712,9 +722,16 @@ def reduce(self, method, axis=(0,), keepdims=False, **kwargs):
...
+                result = method(result, reduce_super_ufunc(self.fill_value, a.shape[1] - counts)).astype(result.dtype)
```
hameerabbasi: `counts.dtype` is almost definitely `int64` or `uint64`, plenty. Using `astype` to ensure not too big a type, as `result.dtype` is the dtype of the reduced array.

sparse/coo/core.py
```diff
@@ -712,9 +722,16 @@ def reduce(self, method, axis=(0,), keepdims=False, **kwargs):
...
+                result_fill_value = reduce_super_ufunc(self.fill_value, a.shape[1])
```
hameerabbasi: The fill value is automatically cast.

hameerabbasi:
The line not hit wasn't being hit originally either...

PR #239, opened 2019-03-11
[WIP] Add __array_function__ support.

pentschev:
I still need to work on tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ matrix:
       env: NUMPY_VERSION="<1.14.0"
     - python: 3.6
     - python: 3.7
+    - python: 3.7
       env:
         - NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
         - NUMPY_VERSION="==1.16.2"
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,9 @@ matrix:
       env: NUMPY_VERSION="<1.14.0"
     - python: 3.6
     - python: 3.7
-      env: NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
-      env: NUMPY_VERSION="==1.16.2"
+      env:
+        - NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
+        - NUMPY_VERSION="==1.16.2"
       dist: xenial
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ matrix:
       env: NUMPY_VERSION="<1.14.0"
     - python: 3.6
     - python: 3.7
+      env: NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
+      env: NUMPY_VERSION="==1.16.2"
       dist: xenial
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ matrix:
       env: NUMPY_VERSION="<1.14.0"
     - python: 3.6
     - python: 3.7
+      dist: xenial
     - python: 3.7
       env:
         - NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -1436,6 +1436,22 @@ class COO(SparseArray, NDArrayOperatorsMixin):
         except NotImplementedError:
             return NotImplemented
 
+    def __array_function__(self, func, types, args, kwargs):
+        import sparse as module
+        for submodule in func.__module__.split('.')[1:]:
+            try:
+                module = getattr(module, submodule)
+            except AttributeError:
+                return NotImplemented
+        if not hasattr(module, func.__name__):
+            return NotImplemented
+        sparse_func = getattr(module, func.__name__)
+        if sparse_func is func:
+            return NotImplemented
+        if not any(issubclass(t, (COO, np.ndarray)) for t in types):
+            return NotImplemented
+        return sparse_func(*args, **kwargs)
+
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         out = kwargs.pop('out', None)
         if out is not None and not all(isinstance(x, COO) for x in out):
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -864,6 +864,8 @@ class COO(SparseArray, NDArrayOperatorsMixin):
         """
         return np.maximum.reduce(self, out=out, axis=axis, keepdims=keepdims)
 
+    amax = max
+
     def any(self, axis=None, keepdims=False, out=None):
         """
         See if any values along array are ``True``. Uses all axes by default.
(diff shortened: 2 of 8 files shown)


sparse/coo/core.py
```diff
@@ -1436,6 +1436,22 @@ def __rmatmul__(self, other):
         except NotImplementedError:
             return NotImplemented
 
+    def __array_function__(self, func, types, args, kwargs):
```
hameerabbasi: I think it's worth moving this to `SparseArray`, and it'll be inherited automatically.

sparse/coo/core.py
```diff
@@ -1436,6 +1436,22 @@ def __rmatmul__(self, other):
...
+        import sparse as module
+        for submodule in func.__module__.split('.')[1:]:
+            try:
```
hameerabbasi: The `try`/`catch` can be moved outside the `for`, with even the line `sparse_func = getattr(module, func.__name__)` inside it.

sparse/coo/core.py
```diff
@@ -1436,6 +1436,22 @@ def __rmatmul__(self, other):
...
+                module = getattr(module, submodule)
+            except AttributeError:
+                return NotImplemented
```
hameerabbasi: It's worth doing `if len(func.__module__.split('.')) == 1: sparse_func = getattr(self, func.__name__)` as well.

hameerabbasi: This is because we don't have `sparse.sum`, `sparse.mean` etc... We simply have methods.

sparse/coo/core.py
pentschev: Yes, thanks for that, I totally missed the inheritance over there.

sparse/coo/core.py
pentschev: I'm not sure I understand. What would be inside the `for`?

sparse/coo/core.py
pentschev: > This is because we don't have `sparse.sum`, `sparse.mean` etc... We simply have methods.

And why is that? I think this will in fact create some problems with `__array_function__`. Calling for example `np.transpose()` on a `sparse.COO` would then fail, and I'm certain this happens in some in Dask, for example.

sparse/coo/core.py
```diff
@@ -1436,6 +1436,22 @@ def __rmatmul__(self, other):
         except NotImplementedError:
             return NotImplemented
 
+    def __array_function__(self, func, types, args, kwargs):
+        import sparse as module
+        for submodule in func.__module__.split('.')[1:]:
```
hameerabbasi:
```suggestion
        try:
            submodules = func.__module__.split('.')[1:]
            for submodule in submodules:
                module = getattr(module, submodule)
            sparse_func = getattr(module, func.__name__)
            return sparse_func(*args, **kwargs)
        except AttributeError:
            pass
        if not hasattr(type(self), func.__name__):
            return NotImplemented
            
        sparse_func = getattr(type(self), func.__name__)
        return sparse_func(*args, **kwargs)
```

sparse/coo/core.py
hameerabbasi: Take a look at my suggestion. ;)

sparse/coo/core.py
pentschev: Oh, I see now. Indeed, that's a great solution, committing!

hameerabbasi: Yes, GitHub has no way to do multiline suggestions AFAICT.

sparse/sparse_array.py
```diff
@@ -212,3 +212,19 @@ def __array__(self, **kwargs):
                                'To manually densify, use the todense method.')
 
         return np.asarray(self.todense(), **kwargs)
+
+    def __array_function__(self, func, types, args, kwargs):
```
hameerabbasi: You'll want to move the changes over here... My bad for suggesting them in the other place.

pentschev: Yup, already committed that locally. No worries, I had too many implementations, so that was my fault. :)

pentschev: Just working on some tests now, then I'll push everything at once.

hameerabbasi:
This pull request **introduces 1 alert** and **fixes 1** when merging 11715a40158832510905b293d72b039cf4a0bccc into 23a4858600d0754d4f8b35e47d44f0db75c9c111 - [view on LGTM.com](https://lgtm.com/projects/g/pydata/sparse/rev/pr-b182e9a3b86aa5bf5600b0972347f075ea2dce95)

**new alerts:**

* 1 for Syntax error

**fixed alerts:**

* 1 for \_\_eq\_\_ not overridden when adding attributes

---

*Comment posted by [LGTM.com](https://lgtm.com)*

.travis.yml
```diff
@@ -9,6 +9,8 @@ matrix:
       env: NUMPY_VERSION="<1.14.0"
     - python: 3.6
     - python: 3.7
+      env: NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
+      env: NUMPY_VERSION="==1.16.2"
```
pentschev: I guess this is the issue. I don't have much experience with Travis, should the correct syntax be like below?

```
env:
  NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
  NUMPY_VERSION="==1.16.2"
```

hameerabbasi: Something like this, I think:
```
env:
  - NUMPY_EXPERIMENTAL_ARRAY_FUNCTION="1"
  - NUMPY_VERSION="==1.16.2"
```

Reference: https://docs.travis-ci.com/user/environment-variables/

sparse/tests/test_array_function.py
```diff
@@ -0,0 +1,22 @@
+import sparse
+from sparse.utils import assert_eq
+import pytest
+np = pytest.importorskip('numpy', minversion='1.16')
```
hameerabbasi: Also add a skip here based on the environment variable. We don't want tests to fail if one didn't export the var. :slightly_smiling_face:

.travis.yml
hameerabbasi: Also add a new env for the `__array_function__` testing, leave the current python 3.7 one as-is.

pentschev: It's a bit confusing when to add the `-` or not. In Dask's `.travis.yml` there's also one before `env:`. I will try not to add one for now and follow exactly your suggestion.

hameerabbasi:
Okay: I see three issues:

1. `min` and `max` call out to `amin` and `amax`. Easily fixed by aliasing: `COO.amin = COO.min`, etc.
2. Properties should not be "called". One can simply do `if not isinstance(sparse_func, typing.Callable): return getattr(self, func.__name__) `.
3. `argwhere` isn't defined... This should be fixed by having a simple implementation.

hameerabbasi:
Ideally one would have the tiniest docstring: `See COO.min`.

hameerabbasi:
cc @shoyer @mrocklin: We ended up needing `np.NotImplementedButCoercible` too, apparently, for `amin`/`amax` and `argwhere`.

hameerabbasi:
You're also free to let me finish up this PR (you retain authorship, of course).

pentschev:
@hameerabbasi would you prefer to finish it? It's up to you, I can continue or you can go ahead and finish.

hameerabbasi:
Just waiting on CI to turn green before the merge.

hameerabbasi:
Thanks, @pentschev for taking this rather huge change on, especially as your first contribution!

pentschev:
@hameerabbasi nice that you fix everything, thanks for the help!

PR #251, opened 2019-04-19
Add resize method and change reshape

### LINKED ISSUE #241: Add resize function and method
Would be nice to have an analogous function and method to NumPy's `resize`.

ref: https://docs.scipy.org/doc/numpy/reference/generated/numpy.resize.html
ref: https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.resize.html

### LINKED ISSUE #250: API incompatibility for COO.reshape
Ndarray.reshape raises a value error when the two shapes don't have the same number of elements. COO.reshape currently allows reshaping to any new shape without raising any error. This is more in line with ndarray.resize. However, resize is an in-place operation whereas reshape is not. I can probably make a pull request for this pretty soon. Seems like a quick fix.

daletovar:
I added a validation check in the reshape method for compatibility with numpy. I also added an in-place resize method. I didn't do anything with the refcheck parameter. Just like numpy, the shape can be passed either as a tuple or a series of integers, hence the *args. 

Fixes #241 
Fixes #250

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ coverage.xml
 *,cover
 .pytest_cache/
 test_results/
+junit/
 
 # Airspeed velocity
 .asv/
--- a/docs/generated/sparse.COO.resize.rst
+++ b/docs/generated/sparse.COO.resize.rst
@@ -1,0 +1,6 @@
+COO.resize
+==========
+
+.. currentmodule:: sparse
+
+.. automethod:: COO.resize
--- a/docs/generated/sparse.COO.rst
+++ b/docs/generated/sparse.COO.rst
@@ -73,6 +73,7 @@ COO
       COO.copy
       COO.dot
       COO.reshape
+      COO.resize
       COO.transpose
       COO.nonzero
 
--- a/sparse/coo/core.py
+++ b/sparse/coo/core.py
@@ -1753,10 +1753,8 @@ class COO(SparseArray, NDArrayOperatorsMixin):  # lgtm [py/missing-equals]
         else:
             raise ValueError('Invalid input')
         
-        if any(d == -1 for d in shape):
-            extra = int(self.size /
-                        np.prod([d for d in shape if d != -1]))
-            shape = tuple([d if d != -1 else extra for d in shape])
+        if any(d < 0 for d in shape):
+            raise ValueError('negative dimensions not allowed')
         
         # TODO: this self.size enforces a 2**64 limit to array size
         linear_loc = self.linear_loc()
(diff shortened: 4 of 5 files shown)


hameerabbasi:
I hope you don't mind I edited your comment with magic words that indicate a fix.

Would you mind adding tests? Just a small one for reshape and resize each, you can find them in https://github.com/pydata/sparse/blob/948cee79dbfba9c69ea52f04806e0f9dc5e05032/sparse/tests/test_coo.py

If you do `pip install -e .[all]` (assuming you're in the repo dir), and then run `pytest`, it'll run all the tests for you.

daletovar:
I'm having a tough time figuring it out. I might not get a chance until tomorrow to try to add that.

hameerabbasi:
Of course! If you’re having trouble figuring it out, don’t hesitate to ask more questions. I’m here to help.

hameerabbasi:
@daletovar If you're okay with it can I finish up this PR?

daletovar:
@hameerabbasi I'm so sorry for getting to this so late. I have a numba-related issue that's causing tests to fail on my laptop. I can finish it up at work on Monday. But I wouldn't mind if you finished it up either. I certainly don't mean to drag this out.

hameerabbasi:
No pressure! You can finish it up if you like. :-)

daletovar:
Thanks!

hameerabbasi:
> I'm so sorry for getting to this so late. I have a numba-related issue that's causing tests to fail on my laptop. I can finish it up at work on Monday. But I wouldn't mind if you finished it up either. I certainly don't mean to drag this out.

Could you tell me your exact version and environment? I've seen this issue reported but have been unable to reproduce it myself...

daletovar:
I'm having a tough time running the tests. I added a resize test but because I can't seem to run them I don't have a good sense of what's happening. Thanks so much for your patience on all of this. Would you mind having a look and running it? ^^The issue with my laptop had to do with not being able to properly update llvmlite.

hameerabbasi:
Okay, there were a few issues here:

1. The `test_caching` test was wrong... That's legacy code. :) I fixed it up.
2. Your code wasn't properly formatted, it was tripping the linter.
3. There were garbage values in the dense array but not the sparse one, they had to be gotten rid of.
4. If the shape being resized to was smaller, we had to discard some values.

hameerabbasi:
This is merged, thanks @daletovar!

daletovar:
Thanks so much, @hameerabbasi ! I really appreciate it.

PR #267, opened 2019-08-11
__array_function__ suppresses AttributeError

crusaderky:
Fix bug where any of the top-level functions invoked by ``COO.__array_function__`` raises AttributeError for any reason, but the exception is improperly suppressed and ``__array_function__`` returns NotImplemented instead.

### DIFF
--- a/sparse/sparse_array.py
+++ b/sparse/sparse_array.py
@@ -221,14 +221,15 @@ class SparseArray:
             for submodule in submodules:
                 module = getattr(module, submodule)
             sparse_func = getattr(module, func.__name__)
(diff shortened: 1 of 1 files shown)


hameerabbasi:
Thanks. Is a test needed to check for this use-case?

crusaderky:
@hameerabbasi it's a matter of defensive coding than an actual real situation. I can't think of any way to deliberately trigger the failure - not without digging into the numpy implementation which would be pretty fragile.

hameerabbasi:
Thanks, @crusaderky.

PR #272, opened 2019-08-11
__array_function__ test coverage; numpy 1.17; result_type()

crusaderky:
- Support for numpy 1.17 (it's already what's used in CI today)
- Implement sparse.result_type + unit tests
- Refactor environment variables and other switches into sparse.settings
- Add unit tests for mixed numpy+sparse inputs to ``__array_function__``, with different permutations

### DIFF
--- a/docs/generated/sparse.result_type.rst
+++ b/docs/generated/sparse.result_type.rst
@@ -1,0 +1,6 @@
+result_type
+===========
+
+.. currentmodule:: sparse
+
+.. autofunction:: result_type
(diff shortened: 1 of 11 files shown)


hameerabbasi:
I added a nice test for NEP-18 here: dask/dask#4675. Would you care to copy that one, as it tests the protocol?

crusaderky:
@hameerabbasi done

hameerabbasi:
Can you change the implementation of result_type to the following:

`return np.result_type(*(x.dtype if isinstance(x, SparseArray) else x for x in args)`

And move it to common.py?

crusaderky:
All done, see if you like it

Issue #288, opened 2019-09-13
Add diagonal/diag operations


pettni:
Thanks developing this, great stuff!

In a project I ended up needing two types of diagonalization operations, one that extracts a diagonal (reduces dimension by 1), and one that diagonalizes an existing dimension (increasing dimension by 1). 

The first is like [np.diagonal](https://docs.scipy.org/doc/numpy/reference/generated/numpy.diagonal.html)

The second is like [tensorflow.diag](https://www.tensorflow.org/api_docs/python/tf/linalg/diag), but with an input to select which dimension to diagonalize.

Both functions are quite simple, so maybe they are not within the scope, but if there is interest in adding them I can create a pull request.

hameerabbasi:
They're definitely within scope. A PR would be great!

rgommers:
For the `tensorflow.diag` one, it raises the question for me if you want to be adding functions/methods that are similar to but incompatible with functions in NumPy?

hameerabbasi:
`np.diag` does both in NumPy, IIUC.

pettni:
It's a valid question. ```np.diag``` does both, but only for 1D->2D and 2D->1D. I think np.diag is meant to imitate the function in Matlab with the same name, but it doesn't seem practical to have a single function doing both in the Nd case

[np.diag source](https://github.com/numpy/numpy/blob/v1.17.0/numpy/lib/twodim_base.py#L216-L285)

hameerabbasi:
There’s `np.diagonal` as well, which does ND, but it only eliminates dimensions, doesn’t add them.

hameerabbasi:
Fixed by #289.

PR #289, opened 2019-09-17
Diagonal/diagonalize functions

pettni:
Additions discussed in #288

### DIFF
--- a/sparse/_coo/common.py
+++ b/sparse/_coo/common.py
@@ -1215,6 +1215,127 @@ def ones_like(a, dtype=None):
     return ones(a.shape, dtype=(a.dtype if dtype is None else dtype))
 
 
+def diagonal(a, offset=0, axis1=0, axis2=1):
+    """
+    Extract diagonal from a COO array. The equivalent of :obj:`numpy.diagonal`.
+
+    Parameters
+    ----------
+    a: COO
+        The array to perform the operation on.
+    offset: int, optional
+        Offset of the diagonal from the main diagonal. Defaults to main diagonal (0).
+    axis1: int, optional
+        First axis from which the diagonals should be taken.  
+        Defaults to first axis (0).
+    axis2 : int, optional
+        Second axis from which the diagonals should be taken.  
+        Defaults to second axis (1).
+
+    Examples
+    --------
+    >>> import sparse
+    >>> x = sparse.as_coo(np.arange(9).reshape(3,3))
+    >>> sparse.diagonal(x).todense()
+    array([0, 4, 8])
+    >>> sparse.diagonal(x,offset=1).todense()
+    array([1, 5])
+
+    >>> x = sparse.as_coo(np.arange(12).reshape((2,3,2)))
+    >>> x_diag = sparse.diagonal(x, axis1=0, axis2=2)
+    >>> x_diag.shape
+    (3, 2)
+    >>> x_diag.todense()
+    array([[ 0,  7],
+           [ 2,  9],
+           [ 4, 11]])
+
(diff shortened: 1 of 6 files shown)


sparse/_coo/core.py
```diff
@@ -2258,6 +2258,112 @@ def asformat(self, format, compressed_axes=None):
...
+        :obj:`numpy.diagonal`: NumPy equivalent function
+        """
+        if self.shape[axis1] != self.shape[axis2]:
+            raise ValueError('shape[axis1] != shape[axis2]')
+
+        diag_axes = [axis for axis in range(len(self.shape)) if axis != axis1 and axis != axis2] + [axis1]
+        diag_shape = [self.shape[axis] for axis in diag_axes]
+        diag_shape[-1] -= abs(offset)
+
+        diag_idx = [i for i in range(len(self.data)) if self.coords[axis1][i]+offset == self.coords[axis2][i]]
```
hameerabbasi: Is it possible to move this loop into Numba?

sparse/_coo/core.py
```diff
@@ -2258,6 +2258,112 @@ def asformat(self, format, compressed_axes=None):
 
         raise NotImplementedError("The given format is not supported.")
 
+    def diagonal(self, offset=0, axis1=0, axis2=1):
```
hameerabbasi: Let's convert this into functions rather than methods, as it is in NumPy.

sparse/_coo/core.py
```diff
@@ -2258,6 +2258,112 @@ def asformat(self, format, compressed_axes=None):
...
+        diag_shape[-1] -= abs(offset)
+
+        diag_idx = [i for i in range(len(self.data)) if self.coords[axis1][i]+offset == self.coords[axis2][i]]
+
+        diag_coords = [self.coords[axis][diag_idx] for axis in diag_axes]
+        diag_data  = self.data[diag_idx]
+
+        return COO(diag_coords, diag_data, diag_shape)
+
+    def diagonalize(self, axis=0):
```
hameerabbasi: Method -> Function here as well. Also need to mirror the NumPy API...

pettni: As discussed in #288, there is no numpy equivalent of ```diagonalize```. Thereis ```np.diag``` but that only does the same thing for 1D and actually does the opposite for 2D. It does not accept 3D or higher. I thought calling it ```diagonalize``` was the best solution to not confuse

hameerabbasi: That seems reasonable... But future code might not work if we have to change the signature to be in line with NumPy, so I think a warning is in order in the docs.

sparse/_coo/common.py
```diff
@@ -1215,6 +1215,127 @@ def ones_like(a, dtype=None):
...
+    diag_data = a.data[diag_idx]
+
+    return COO(diag_coords, diag_data, diag_shape)
+
+
+def diagonalize(a, axis=0):
+    """
+    Diagonalize a COO array. The new dimension is appended at the end.
+
+    **Warning**: :obj:`diagonalize` is not :obj:`numpy` compatible as there is no direct :obj:`numpy` equivalent. The API may change in the future.
```
hameerabbasi: This should be in a [`..warning::` directive](http://docutils.sourceforge.net/docs/ref/rst/directives.html#warning).

hameerabbasi:
You can run `black .` to fix the linting errors.

hameerabbasi:
Excellent, thanks, @pettni!

pettni:
Happy to contribute, thanks for the quick reviews!

PR #296, opened 2019-10-17
Avoid making a copy of `coords`

eric-wieser:
Coords is not modified anywhere in this class, so we probably do not need to copy it.

This also would allow for identity short-cuts when performing operations on sparse arrays with the same sparsity pattern.

### DIFF
--- a/sparse/_coo/core.py
+++ b/sparse/_coo/core.py
@@ -239,7 +239,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):  # lgtm [py/missing-equals]
                 shape = ()
 
         super().__init__(shape, fill_value=fill_value)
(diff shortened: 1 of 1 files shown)


hameerabbasi:
Thanks, @eric-wieser!

eric-wieser:
Worth noting that bad things could happen if anyone is doing something that mutates `coords`, such as:
```python
coords = np.zeros((3, 200), np.intp)
s = COO(coords=coords, ...)
coords[...] = 0  # !!!
```

hameerabbasi:
Yes, we don't follow that pattern anywhere in our code, and I'm fairly sure others don't either. Anyway, thanks for noting this.

PR #303, opened 2019-11-16
Add stack and concatenate

### LINKED ISSUE #301: GCXS indexing errors
@daletovar I was just playing around with the GCXS class on d9ed64d140d2d2c and ran into a couple indexing bugs. The first deals with incorrect results being returned when there is more than one compressed axis, the second deals with segfaults from similar cases.

In the first instance, incorrect values are being returned after indexing into uncompressed dimensions:

```python
from sparse._compressed.compressed import GCXS
import numpy as np

a = np.arange(27).reshape((3, 3, 3)).copy()
s = GCXS(a, compressed_axes=[0, 1])
s[:, :, 1].todense()
# array([[ 1, 13, 25],
#        [ 0,  0,  0],
#        [ 0,  0,  0]])
```

This does not happen if there is only one compressed axis:

```python
GCXS(a, compressed_axes=[1])[:, :, 1].todense()
# array([[ 1,  4,  7],
#        [10, 13, 16],
#        [19, 22, 25]])
```

If I expand to more dimensions, this keeps happening. 

```python
n = 4

a = np.arange(3 ** n).reshape([3] * n).copy()
s = GCXS(a, compressed_axes=[0, 1])
s[:, :, :, 1].nnz  # Should be 27
# 9
```

This leads to the second bug. Calling a conversion method on the result of that indexing statement causes a segfault.

```python
s[:, :, :, 1].todense() 
# Python(56425,0x10d72c5c0) malloc: can't allocate region
# *** mach_vm_map(size=11329358692995072) failed (error code=3)
# Python(56425,0x10d72c5c0) malloc: *** set a breakpoint in malloc_error_break to debug
# [1]    56425 segmentation fault  ipython
```

These occur on conversion to either dense or coo (haven't tried others). Similarly, I only run into issues with more than one compressed dimension. This occurs after indexing into any uncompressed dimension, but not the compressed ones. If only one dimension was compressed, indexing into the uncompressed dimensions works fine.

daletovar:
1) adds stack and concatenate for GCXS arrays.
2) fixes #301


daletovar:
Looks like I'll have the same commit issue as last time.

hameerabbasi:
Can you specify what commit issue? I can squash the commits but the tests seem to be failing pretty badly right now.

daletovar:
Oh I just meant we need to squash the commits. I think I found what's breaking.

hameerabbasi:
Don't worry about that too much, we can always just "Squash and Merge".

hameerabbasi:
One comment: We don't want to guarantee the type of the output array. If `GCXS` goes in and `COO` comes out, that's totally fine.

What we DO want is for more optimized versions to be available for each format, instead of round-tripping.

daletovar:
I totally agree. In light of that concern is there anything that you think I should change?

hameerabbasi:
1. The documentation should be changed.
2. Where you're converting the output back to GCXS, don't.

daletovar:
@hameerabbasi, it looks like everything's passing. Is there anything else you think I should do?

sparse/_common.py
```diff
@@ -0,0 +1,387 @@
...
+    --------
+    numpy.stack : NumPy equivalent function
+    """
+    from ._compressed import GCXS
+
+    gcxs_arrays = 0
+    for arr in arrays:
+        if isinstance(arr, GCXS):
+            gcxs_arrays += 1
+    if gcxs_arrays > len(arrays) / 2:
```
shoyer: I would suggest picking a firm rule about which sparse type has higher precedence rather than going with the majority. I think this would be easier to understand and would result in more predictable behavior.

hameerabbasi: I agree here. I would think, for now, dispatch to `COO` if there's any `COO` array present.

daletovar:
Thanks for reviewing, @shoyer. @hameerabbasi, is there anything else you'd like to see changed?

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    arrays : Iterable[SparseArray]
+        The input arrays to stack.
+    axis : int, optional
+        The axis along which to stack the input arrays.
+    compressed_axes : iterable, optional
+        The axes to compress if returning a GCXS array.
+
+    Returns
+    -------
+    COO or GCXS
```
hameerabbasi: This should be changed to `SparseArray`.

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    arrays : Iterable[SparseArray]
+        The input arrays to concatenate.
+    axis : int, optional
+        The axis along which to concatenate the input arrays. The default is zero.
+    compressed_axes : iterable, optional
+        The axes to compress if returning a GCXS array.
+
+    Returns
+    -------
+    COO or GCXS
```
hameerabbasi: `SparseArray`

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    dtype : data-type, optional
+        Data-type of the returned array.
+    format : str, optional
+        A format string.
+    compressed_axes : iterable, optional
+        The axes to compress if returning a GCXS array.
+
+    Returns
+    -------
+    I : [COO, GCXS, DOK] array of shape (N, M)
```
hameerabbasi: `SparseArray`

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+        data_length = max(min(data_length, N + k), 0)
+        m_coords = np.arange(data_length, dtype=np.intp)
+        n_coords = m_coords - k
+    else:
+        n_coords = m_coords = np.arange(data_length, dtype=np.intp)
+
+    coords = np.stack([n_coords, m_coords])
+    data = np.array(1, dtype=dtype)
+
+    return COO(
```
hameerabbasi: Is it possible to specialise this for `GCXS`? Doesn't need to be in this PR.

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    """
+    from sparse import COO
+
+    if dtype is None:
+        dtype = np.array(fill_value).dtype
+    if not isinstance(shape, tuple):
+        shape = (shape,)
+    data = np.empty(0, dtype=dtype)
+    coords = np.empty((len(shape), 0), dtype=np.intp)
+    return COO(
```
hameerabbasi: Same here, is specialization possible?

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    out : array with the specified format [COO, GCXS, DOK]
+        Array of `fill_value` with the same shape and type as `a`.
+
+    Examples
+    --------
+    >>> x = np.ones((2, 3), dtype='i8')
+    >>> full_like(x, 9.0).todense()  # doctest: +NORMALIZE_WHITESPACE
+    array([[9, 9, 9],
+           [9, 9, 9]])
+    """
```
hameerabbasi: Can we add a `if format is None: format=a.format` here?

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    return full(
```
hameerabbasi: Can we change this to `full(..., format=format)`

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+        coords,
+        data=data,
+        shape=shape,
+        fill_value=fill_value,
+        has_duplicates=False,
+        sorted=True,
+    ).asformat(format, compressed_axes=compressed_axes)
+
+
+def full_like(a, fill_value, dtype=None, format="coo", compressed_axes=None):
```
hameerabbasi: Can we make the default here be `format=None`? These changes are so that format can be inferred from `a` as well.

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    >>> zeros((2, 2), dtype=int).todense()  # doctest: +NORMALIZE_WHITESPACE
+    array([[0, 0],
+           [0, 0]])
+    """
+    return full(shape, 0, np.dtype(dtype)).asformat(
+        format, compressed_axes=compressed_axes
+    )
+
+
+def zeros_like(a, dtype=None, format="coo", compressed_axes=None):
```
hameerabbasi: Read comments for `full_like`.

sparse/_common.py
hameerabbasi: Same for `compressed_axes`.

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    >>> ones((2, 2), dtype=int).todense()  # doctest: +NORMALIZE_WHITESPACE
+    array([[1, 1],
+           [1, 1]])
+    """
+    return full(shape, 1, np.dtype(dtype)).asformat(
+        format, compressed_axes=compressed_axes
+    )
+
+
+def ones_like(a, dtype=None, format="coo", compressed_axes=None):
```
hameerabbasi: Read comments for `full_like`.

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    dtype : data-type, optional
+        Overrides the data type of the result.
+    format : str, optional
+        A format string.
+    compressed_axes : iterable, optional
+        The axes to compress if returning a GCXS array.
+
+    Returns
+    -------
+    out : [COO, GCXS, DOK]
```
hameerabbasi: `SparseArray`

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+        The desired data-type for the array, e.g., `numpy.int8`.  Default is
+        `numpy.float64`.
+    format : str, optional
+        A format string.
+    compressed_axes : iterable, optional
+        The axes to compress if returning a GCXS array.
+
+    Returns
+    -------
+    out : [COO, GCXS, DOK]
```
hameerabbasi: `SparseArray`

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+    dtype : data-type, optional
+        Overrides the data type of the result.
+    format : str, optional
+        A format string.
+    compressed_axes : iterable, optional
+        The axes to compress if returning a GCXS array.
+
+    Returns
+    -------
+    out : [COO, GCXS, DOK]
```
hameerabbasi: `SparseArray`

sparse/_common.py
```diff
@@ -0,0 +1,379 @@
...
+        The desired data-type for the array, e.g., `numpy.int8`.  Default is
+        `numpy.float64`.
+    format : str, optional
+        A format string.
+    compressed_axes : iterable, optional
+        The axes to compress if returning a GCXS array.
+
+    Returns
+    -------
+    out : array with the specified format [COO, GCXS, DOK]
```
hameerabbasi: `SparseArray`

daletovar:
Thanks so much, @hameerabbasi! I parameterized compressed_axes in the indexing tests so that should address #301.

PR #307, opened 2020-01-17
ENH: Add numba extension support for COO objects

eric-wieser:
Over in pygae/clifford, we're finding we want to be able to write optimized loops using `arr.coords` and `arr.data`.

This aims to be the minimum amount needed to make that work.

~~Marking as draft since this has no tests~~

cc @stuartarchibald, @luk-f-a

### DIFF
--- a/sparse/_coo/numba_extension.py
+++ b/sparse/_coo/numba_extension.py
@@ -19,7 +19,7 @@ from numba.targets.imputils import impl_ret_borrowed, lower_constant
 from numba.typing.typeof import typeof_impl
 from numba import cgutils, types
 from sparse._utils import _zero_of_dtype
-from contextlib import ExitStack
+from contextlib import ExitStack, contextmanager
 
 from . import COO
 
@@ -122,6 +122,30 @@ def lower_constant_COO(context, builder, typ, pyval):
     )
 
 
+@contextmanager
+def early_exit_if(builder, stack: ExitStack, cond):
+    """
+    Emit code similar to::
+
+        if (cond) {
+            <body>
+            return;
+        }
+        <everything after this call>
+
+    However, this "return" will just break out of the current `ExitStack`,
+    rather than out of the whole function
+    """
+    then, otherwise = stack.enter_context(builder.if_else(cond, likely=False))
+    with then:
+        yield
+    stack.enter_context(otherwise)
+
+
+def early_exit_if_null(builder, stack, obj):
+    return early_exit_if(builder, stack, cgutils.is_null(builder, obj))
+
+
 def _unbox_native_field(typ, obj, field_name: str, c):
     ret_ptr = cgutils.alloca_once(c.builder, c.context.get_value_type(typ))
     is_error_ptr = cgutils.alloca_once_value(c.builder, cgutils.false_bit)
@@ -129,19 +153,15 @@ def _unbox_native_field(typ, obj, field_name: str, c):
 
     with ExitStack() as stack:
         field_obj = c.pyapi.object_getattr_string(obj, field_name)
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, field_obj), likely=False))
-        with then:
+        with early_exit_if_null(c.builder, stack, field_obj):
             c.builder.store(cgutils.true_bit, is_error_ptr)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         field_native = c.unbox(typ, field_obj)
         c.pyapi.decref(field_obj)
-        then, otherwise = stack.enter_context(c.builder.if_else(field_native.is_error, likely=False))
-        with then:
+        with early_exit_if(c.builder, stack, field_native.is_error):
             c.builder.store(cgutils.true_bit, is_error_ptr)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         c.builder.store(cgutils.false_bit, is_error_ptr)
         c.builder.store(field_native.value, ret_ptr)
@@ -157,32 +177,24 @@ def unbox_COO(typ: COOType, obj: COO, c) -> NativeValue:
 
     with ExitStack() as stack:
         data = _unbox_native_field(typ.data_type, obj, "data", c)
-        then, otherwise = stack.enter_context(c.builder.if_else(data.is_error, likely=False))
-        with then:
+        with early_exit_if(c.builder, stack, data.is_error):
             c.builder.store(cgutils.true_bit, is_error_ptr)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         coords = _unbox_native_field(typ.coords_type, obj, "coords", c)
-        then, otherwise = stack.enter_context(c.builder.if_else(coords.is_error, likely=False))
-        with then:
+        with early_exit_if(c.builder, stack, coords.is_error):
             c.builder.store(cgutils.true_bit, is_error_ptr)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         shape = _unbox_native_field(typ.shape_type, obj, "shape", c)
-        then, otherwise = stack.enter_context(c.builder.if_else(shape.is_error, likely=False))
-        with then:
+        with early_exit_if(c.builder, stack, shape.is_error):
             c.builder.store(cgutils.true_bit, is_error_ptr)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         fill_value = _unbox_native_field(typ.fill_value_type, obj, "fill_value", c)
-        then, otherwise = stack.enter_context(c.builder.if_else(fill_value.is_error, likely=False))
-        with then:
+        with early_exit_if(c.builder, stack, fill_value.is_error):
             c.builder.store(cgutils.true_bit, is_error_ptr)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         coo = cgutils.create_struct_proxy(typ)(c.context, c.builder)
         coo.coords = coords.value
@@ -203,63 +215,49 @@ def box_COO(typ: COOType, val: "some LLVM thing", c) -> COO:
         coo = cgutils.create_struct_proxy(typ)(c.context, c.builder, value=val)
 
         data_obj = c.box(typ.data_type, coo.data)
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, data_obj), likely=False))
-        with then:
+        with early_exit_if_null(c.builder, stack, data_obj):
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
-        coords_obj = c.box(typ.coords_type, coo.coords)
 
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, coords_obj), likely=False))
-        with then:
+        coords_obj = c.box(typ.coords_type, coo.coords)
+        with early_exit_if_null(c.builder, stack, coords_obj):
             c.pyapi.decref(data_obj)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         shape_obj = c.box(typ.shape_type, coo.shape)
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, shape_obj), likely=False))
-        with then:
+        with early_exit_if_null(c.builder, stack, shape_obj):
             c.pyapi.decref(coords_obj)
             c.pyapi.decref(data_obj)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         fill_value_obj = c.box(typ.fill_value_type, coo.fill_value)
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, fill_value_obj), likely=False))
-        with then:
+        with early_exit_if_null(c.builder, stack, fill_value_obj):
             c.pyapi.decref(shape_obj)
             c.pyapi.decref(coords_obj)
             c.pyapi.decref(data_obj)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         class_obj = c.pyapi.unserialize(c.pyapi.serialize_object(COO))
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, class_obj), likely=False))
-        with then:
+        with early_exit_if_null(c.builder, stack, class_obj):
             c.pyapi.decref(shape_obj)
             c.pyapi.decref(coords_obj)
             c.pyapi.decref(data_obj)
             c.pyapi.decref(fill_value_obj)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         args = c.pyapi.tuple_pack([coords_obj, data_obj, shape_obj])
         c.pyapi.decref(shape_obj)
         c.pyapi.decref(coords_obj)
         c.pyapi.decref(data_obj)
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, args), likely=False))
-        with then:
+        with early_exit_if_null(c.builder, stack, args):
             c.pyapi.decref(fill_value_obj)
             c.pyapi.decref(class_obj)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
 
         kwargs = c.pyapi.dict_pack([("fill_value", fill_value_obj)])
         c.pyapi.decref(fill_value_obj)
-        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, kwargs), likely=False))
-        with then:
+        with early_exit_if_null(c.builder, stack, kwargs):
             c.pyapi.decref(class_obj)
             c.builder.store(fail_obj, ret_ptr)
-        stack.enter_context(otherwise)
         c.builder.store(c.pyapi.call(class_obj, args, kwargs), ret_ptr)
 
     return c.builder.load(ret_ptr)
--- a/sparse/_coo/numba_extension.py
+++ b/sparse/_coo/numba_extension.py
@@ -19,6 +19,7 @@ from numba.targets.imputils import impl_ret_borrowed, lower_constant
 from numba.typing.typeof import typeof_impl
 from numba import cgutils, types
 from sparse._utils import _zero_of_dtype
+from contextlib import ExitStack
 
 from . import COO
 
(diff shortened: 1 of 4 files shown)


sparse/_coo/numba_extension.py
```diff
@@ -0,0 +1,131 @@
...
+        models.StructModel.__init__(self, dmm, fe_type, members)
+
+
+@type_callable(COO)
+def type_COO(context):
+    def typer(coords, data):
+        return COOType(
+        	coords_dtype=numba.numpy_support.as_dtype(coords.dtype),
+        	data_dtype=numba.numpy_support.as_dtype(data.dtype),
+        )
```
eric-wieser: Needs to infer `ndim` too

sparse/_coo/numba_extension.py
```diff
@@ -0,0 +1,131 @@
...
+    def data_type(self):
+        return numba.from_dtype(self.data_dtype)[:]
+
+    @property
+    def coords_type(self):
+        return numba.from_dtype(self.coords_dtype)[:, :]
+
+    @property
+    def shape_type(self):
+       	return types.UniTuple(types.int64, self.ndim)
```
eric-wieser: Is there any way to avoid putting `ndim` in the type itself here, or is this a reasonable approach?

hameerabbasi:
I'm willing to put this in, provided we find a reviewer. I'm afraid my experience with Numba isn't enough to review this PR.

hameerabbasi:
Or there are thorough tests -- More than willing to learn as I go!

sparse/_coo/numba_extension.py
hameerabbasi: I think this is reasonable, with a quick search in Numba's codebase, I couldn't find any way to do this without specifying the length.

stuartarchibald: The length is part of the `Tuple` type. `types.UniTuple(types.int64, 1)` is a different type to `types.UniTuple(types.int64, 2)`.

eric-wieser:
> Or there are thorough tests

Added some tests that should hopefully cover everything here

hameerabbasi:
Could you perform a quick memory test and see if creating a humongous number of objects doesn't blow up memory? Doesn't need to be in the code.

Also, adding the `fill_value` would be nice.

eric-wieser:
> Also, adding the fill_value would be nice.

Perhaps selfish, but since my application leaves that as 0, I'd prefer to leave that to a follow-up. Hopefully with something like this patch in place, it will be easier for others to work out how to add properties.

sparse/_coo/numba_extension.py
```diff
@@ -0,0 +1,141 @@
...
+    shape = c.pyapi.object_getattr_string(obj, "shape")
+    coo = cgutils.create_struct_proxy(typ)(c.context, c.builder)
+    coo.coords = c.unbox(typ.coords_type, coords).value
+    coo.data = c.unbox(typ.data_type, data).value
+    coo.shape = c.unbox(typ.shape_type, shape).value
+    c.pyapi.decref(data)
+    c.pyapi.decref(coords)
+    c.pyapi.decref(shape)
+    is_error = cgutils.is_not_null(c.builder, c.pyapi.err_occurred())
+    return NativeValue(coo._getvalue(), is_error=is_error)
```
eric-wieser: @stuartarchibald: Do I need to call `is_not_null` after every single one of the python attribute accesses? What failure is this looking for here?

eric-wieser:
> Could you perform a quick memory test and see if creating a humongous number of objects doesn't blow up memory?

I don't have a good feel for what such a test would look like.

hameerabbasi:
> I don't have a good feel for what such a test would look like.

Just something that boxes and unboxes random `COO` objects in a loop.

hameerabbasi:
> Perhaps selfish, but since my application leaves that as 0, I'd prefer to leave that to a follow-up. Hopefully with something like this patch in place, it will be easier for others to work out how to add properties.

It is, but there's a correctness problem here. At least check if it's zero?

eric-wieser:
>It is, but there's a correctness problem here.

Ugh, that's unfortunate. Probably easier to add it than to work out how to insert an exception somewhere.

sparse/_coo/numba_extension.py
```diff
@@ -0,0 +1,165 @@
...
+    typ = sig.return_type
+    coords, data, shape = args
+    coo = cgutils.create_struct_proxy(typ)(context, builder)
+    coo.coords = coords
+    coo.data = data
+    coo.shape = shape
+    coo.fill_value = context.get_constant_generic(
+        builder, typ.fill_value_type, _zero_of_dtype(typ.data_dtype)
+    )
+    return impl_ret_borrowed(context, builder, sig.return_type, coo._getvalue())
```
eric-wieser: @sklam suggested that this might be possible to do using `@overload`, which would make the `fill_value` kwargs easy.

Would you be able to give an example?

sparse/_coo/numba_extension.py
```diff
@@ -0,0 +1,165 @@
...
+    data = c.pyapi.object_getattr_string(obj, "data")
+    coords = c.pyapi.object_getattr_string(obj, "coords")
+    shape = c.pyapi.object_getattr_string(obj, "shape")
+    fill_value = c.pyapi.object_getattr_string(obj, "fill_value")
+    coo = cgutils.create_struct_proxy(typ)(c.context, c.builder)
+    coo.coords = c.unbox(typ.coords_type, coords).value
+    coo.data = c.unbox(typ.data_type, data).value
+    coo.shape = c.unbox(typ.shape_type, shape).value
+    coo.fill_value = c.unbox(typ.fill_value_type, fill_value).value
+    c.pyapi.decref(data)
```
hameerabbasi: Is there an `xdecref`?

sparse/_coo/numba_extension.py
hameerabbasi: Maybe by `@overload`-ing the `COO` type itself?

sparse/_coo/numba_extension.py
eric-wieser: Probably. but `data` will be not null so it's not needed here.

sparse/_coo/numba_extension.py
eric-wieser: Right, but in an overload context I don't have access to `builder` and`context`, so can't call `cgutils.create_struct_proxy(typ)`

sparse/_coo/numba_extension.py
hameerabbasi: It could be if there's an error... Or does Numba handle that?

eric-wieser: I'm not sure, hence https://github.com/pydata/sparse/pull/307#discussion_r368073005

sparse/_coo/numba_extension.py
stuartarchibald: @eric-wieser not sure what the conversation with @sklam was, however, [`@intrinsic`](http://numba.pydata.org/numba-doc/latest/extending/high-level.html#implementing-intrinsics) is what you need if you want to use `@overload` and also have the `context`/`builder` accessible.

sparse/_coo/numba_extension.py
stuartarchibald: The `is_error` is to signal something failed in unboxing and that there's an exception pending. I think you might be able to just use `is_error=c.pyapi.c_api_error()`, to make a check to the Python C-API for the error state as per `PyErr_Occurred`.

eric-wieser: I think I perhaps haven't conveyed my point. In c, this would look like:
```
coords = unbox(c_coords);
if (coords == NULL) {
    return NULL;
}
shape = unbox(c_shape);
if (shape == NULL) {
    Py_DECREF(coords);
    return NULL;
}
```
How would I translate that here?

stuartarchibald: I think you'd need to replicate the first code block. Perhaps with things like (untested, I just typed this in):
```python
data = c.pyapi.object_getattr_string(obj, "data")
with cgutils.if_unlikely(c.builder,  cgutils.is_null(c.builder, data)):
  c.builder.ret(c.pyapi.get_null_object())
```
I don't recall from memory whether something also has to signal that an exception is pending. Perhaps try the above and see?

eric-wieser: That sounds roughly correct - if it works, then the interval example in yours docs should be updated to use it.

eric-wieser: With that example, what should I actually return from my python function body?

stuartarchibald: > With that example, what should I actually return from my python function body?

Sorry, I'm not sure to what you are referring? If that branch is taken a NULL is returned?

stuartarchibald: You can take a look at the LLVM IR that's being generated by breaking in this function and looking at `builder.module`.

eric-wieser: I mean what should the python function return, not what should the computation graph return

stuartarchibald: This: `    return NativeValue(coo._getvalue(), is_error=is_error)` seems correct for what you have, whether there's extra error detection things you want to put into the `is_error` bit I cannot ascertain from what's present. Essentially, a `NativeValue` needs returning and in this case the first arg to needs to be a load-instruction where the thing being loaded points to the native representation of the thing being unboxed.

hameerabbasi: Perhaps I can rephrase this another way: Provided every line that calls into Python's interpreter can potentially raise a Python exception, do we need to explicitly check for errors and perform the appropriate refcounting or will Numba skip the lines after the error and perform the correct refcounting for us?

sparse/_coo/numba_extension.py
```diff
@@ -151,37 +152,42 @@ def unbox_COO(typ: COOType, obj: COO, c) -> NativeValue:
     is_error_ptr = cgutils.alloca_once_value(c.builder, cgutils.false_bit)
     fail_obj = c.context.get_constant_null(typ)
 
-    data = _unbox_native_field(typ.data_type, obj, "data", c)
-    with c.builder.if_else(data.is_error, likely=False) as (then, otherwise):
+    with ExitStack() as stack:
```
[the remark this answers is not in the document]
eric-wieser: I don't think I've ever seen as compelling a use-case for it as this one

hameerabbasi:
You need to run `black` to fix the format checks. 😄  Otherwise this is good to go.

sparse/_coo/numba_extension.py
```diff
@@ -0,0 +1,272 @@
...
+        data_obj = c.box(typ.data_type, coo.data)
+        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, data_obj), likely=False))
+        with then:
+            c.builder.store(fail_obj, ret_ptr)
+        stack.enter_context(otherwise)
+        coords_obj = c.box(typ.coords_type, coo.coords)
+
+        then, otherwise = stack.enter_context(c.builder.if_else(cgutils.is_null(c.builder, coords_obj), likely=False))
+        with then:
+            c.pyapi.decref(data_obj)
```
hameerabbasi: Would it be possible to somehow create a context manager that did this `decref` and then remove all these?

eric-wieser: No, because the thing to remember is that `with then` is not a branch, it always runs every time.

However, we could keep a list that we iterate over the same way in each `then`.

hameerabbasi: > However, we could keep a list that we iterate over the same way in each `then`.

Let's do this then.

eric-wieser: To wit:
```diff
--- C:\Users\wiese\Repos\forks\sparse\sparse\_coo\numba_extension.py	Mon Jan 20 15:37:12 2020
+++ C:\Users\wiese\Repos\forks\sparse\sparse\_coo\numba_extension.py	Mon Jan 20 15:42:57 2020
@@ -213,36 +213,42 @@
 
     with ExitStack() as stack:
         coo = cgutils.create_struct_proxy(typ)(c.context, c.builder, value=val)
+        to_cleanup = []
 
         data_obj = c.box(typ.data_type, coo.data)
         with early_exit_if_null(c.builder, stack, data_obj):
-            c.builder.store(fail_obj, ret_ptr)
+            for o in to_cleanup[::-1]:
+                c.pyapi.decref(o)
+            c.builder.store(fail_obj, ret_ptr)
+        to_cleanup.append(data_obj)
 
         coords_obj = c.box(typ.coords_type, coo.coords)
         with early_exit_if_null(c.builder, stack, coords_obj):
-            c.pyapi.decref(data_obj)
-            c.builder.store(fail_obj, ret_ptr)
+            for o in to_cleanup[::-1]:
+                c.pyapi.decref(o)
+            c.builder.store(fail_obj, ret_ptr)
+        to_cleanup.append(coords_obj)
 
         shape_obj = c.box(typ.shape_type, coo.shape)
         with early_exit_if_null(c.builder, stack, shape_obj):
-            c.pyapi.decref(coords_obj)
-            c.pyapi.decref(data_obj)
-            c.builder.store(fail_obj, ret_ptr)
+            for o in to_cleanup[::-1]:
+                c.pyapi.decref(o)
+            c.builder.store(fail_obj, ret_ptr)
+        to_cleanup.append(shape_obj)
 
         fill_value_obj = c.box(typ.fill_value_type, coo.fill_value)
         with early_exit_if_null(c.builder, stack, fill_value_obj):
-            c.pyapi.decref(shape_obj)
-            c.pyapi.decref(coords_obj)
-            c.pyapi.decref(data_obj)
-            c.builder.store(fail_obj, ret_ptr)
+            for o in to_cleanup[::-1]:
+                c.pyapi.decref(o)
+            c.builder.store(fail_obj, ret_ptr)
+        to_cleanup.append(fill_value_obj)
 
         class_obj = c.pyapi.unserialize(c.pyapi.serialize_object(COO))
         with early_exit_if_null(c.builder, stack, class_obj):
-            c.pyapi.decref(shape_obj)
-            c.pyapi.decref(coords_obj)
-            c.pyapi.decref(data_obj)
-            c.pyapi.decref(fill_value_obj)
-            c.builder.store(fail_obj, ret_ptr)
+            for o in to_cleanup[::-1]:
+                c.pyapi.decref(o)
+            c.builder.store(fail_obj, ret_ptr)
+        to_cleanup.append(class_obj)
 
         args = c.pyapi.tuple_pack([coords_obj, data_obj, shape_obj])
         c.pyapi.decref(shape_obj)
```
Not sure it's worth it.

eric-wieser:
> You need to run `black` to fix the format checks. 😄 Otherwise this is good to go.

Decided that a better choice was to avoid having such long expressions - updated to not need any line-wrapping.

sparse/_coo/numba_extension.py
eric-wieser: I think the conclusion was that we're on the hook for checking for python exceptions ourselves. My latest commits handle this, so marking this conversation as resolved.

sparse/_coo/numba_extension.py
```diff
@@ -0,0 +1,269 @@
...
+        }
+        <everything after this call>
+
+    However, this "return" will just break out of the current `ExitStack`,
+    rather than out of the whole function
+    """
+    then, otherwise = stack.enter_context(builder.if_else(cond, likely=False))
+    with then:
+        yield
+    stack.enter_context(otherwise)
```
hameerabbasi:
```python
<iterate through list>
c.builder.store(fail_obj, ret_ptr)
```

Could be here.

eric-wieser: I'm wary of building too much abstraction here, because @stuartarchibald has suggested that something like this should end up in numba itself. The more specific we get, the harder it will be to gut out this code and replace it with numba stuff if they pick a different level of specificity.

sparse/_coo/numba_extension.py
eric-wieser: @hameerabbasi, note that this results in more lines not fewer. Do you still want me to go ahead with this?

sparse/_coo/numba_extension.py
eric-wieser: I attempted this as:
```
@intrinsic
def _make_coo(typingctx, coords, data, shape, fill_value):
    # TODO: accept a fill_value kwarg
    result_type = COOType(
        coords_dtype=numba.numpy_support.as_dtype(coords.dtype),
        data_dtype=numba.numpy_support.as_dtype(data.dtype),
        ndim=len(shape),
    )
    args_types = [
        result_type.coords_dtype,
        result_type.data_dtype,
        result_type.shape_type,
        result_type.fill_value_type
    ]
    if fill_value == types.none:
        args_types[-1] = types.none
        def codegen(context, builder, sig, args):
            typ = sig.return_type
            coords, data, shape, fill_value = args
            coo = cgutils.create_struct_proxy(typ)(context, builder)
            coo.coords = coords
            coo.data = data
            coo.shape = shape
            coo.fill_value = context.get_constant_generic(
                builder, typ.fill_value_type, _zero_of_dtype(typ.fill_value_type)
            )
            return impl_ret_borrowed(context, builder, sig.return_type, coo._getvalue())
    else:
        def codegen(context, builder, sig, args):
            typ = sig.return_type
            coords, data, shape, fill_value = args
            coo = cgutils.create_struct_proxy(typ)(context, builder)
            coo.coords = coords
            coo.data = data
            coo.shape = shape
            coo.fill_value = fill_value
            return impl_ret_borrowed(context, builder, sig.return_type, coo._getvalue())

    return sig, codegen


@overload(COO)
def impl_COO(coords, data, shape, *, fill_value=None):
    # convert a keyword to positional argument
    return _make_coo(coords, data, shape, fill_value)
```
but got the error
```
Invalid use of Function(<class 'sparse._coo.core.COO'>) with argument(s) of type(s): (array(int64, 2d, A), array(int32, 1d, A), UniTuple(int64 x 1))
 * parameterized
In definition 0:
    NotImplementedError: <intrinsic _make_coo> is not usable in pure-python
    raised from C:\Users\wiese\AppData\Local\Programs\Python\Python38\lib\site-packages\numba\extending.py:304
In definition 1:
    NotImplementedError: <intrinsic _make_coo> is not usable in pure-python
    raised from C:\Users\wiese\AppData\Local\Programs\Python\Python38\lib\site-packages\numba\extending.py:304
This error is usually caused by passing an argument of a type that is unsupported by the named function.
[1] During: resolving callee type: Function(<class 'sparse._coo.core.COO'>)
[2] During: typing of call at <ipython-input-2-523b502ede43> (5)


File "<ipython-input-2-523b502ede43>", line 5:
def get_data(c):
    return sparse.COO(c.coords, c.data, c.shape)
```

Let's leave this to a follow-up, there's enough complexity here already.

sparse/_coo/numba_extension.py
hameerabbasi: I was trying to build more-or-less the equivalent of what happens when a Python statement is executed in the interpretor, but I'm happy either way.

hameerabbasi: Which should be generic enough to end up in Numba.

eric-wieser: This ends up adding 3 more arguments to this function though (the list, the fail object, the return pointer), which i think obscures clarity.

sparse/_coo/numba_extension.py
hameerabbasi: No, I think this is fine as-is now. Your points are valid, it might be better to build an object holding this data in a follow-up PR.

sparse/_coo/numba_extension.py
eric-wieser: Resolving, I don't think keyword-only args are possible at all in the latest numba.

eric-wieser:
@hameerabbasi, let me know if you want anything else here, I'm losing track.

hameerabbasi:
Thanks, @eric-wieser!

sparse/_coo/numba_extension.py
stuartarchibald: There's a problem in here: https://github.com/numba/numba/blob/master/numba/typing/templates.py#L154 that will prevent it from working. Resolving as-is is probably the best option for now.

PR #320, opened 2020-02-06
Implement deterministic dask tokenization.

### LINKED ISSUE #300: dask deterministic hashing
sparse.COO should implement dask deterministic hashing:
https://docs.dask.org/en/latest/custom-collections.html#deterministic-hashing

The current situation is that dask falls back on a random uuid4:
```python
>>> import dask.array
>>> import numpy
>>> import sparse
>>> a = numpy.array([1,2])
>>> sa = sparse.COO(a)
>>> dask.array.from_array(a).__dask_keys__()
[('array-de932becc43e72c010bc91ffefe42af1', 0)]
>>> dask.array.from_array(a).__dask_keys__()                                                                                                                                                                                                              
[('array-de932becc43e72c010bc91ffefe42af1', 0)]
>>> dask.array.from_array(sa).__dask_keys__()                                                                                                                                                                                                              
[('array-77964bd4315f9a3eb43953586a91d273', 0)]
>>> dask.array.from_array(sa).__dask_keys__()                                                                                                                                                                                                             
[('array-cf0d4b78cf3a2b277c351ea3d534a790', 0)]
```
Suggested implementation:
```python
class COO:
    def __dask_tokenize__(self):
        from dask.base import normalize_token
        return normalize_token((type(self), self.coords, self.data, self.shape, self.fill_value))
```

danielballan:
Closes #300

Credit to @crusaderky for suggesting this implementation in that issue.
Here, I add a test. I have verified that it fails as expected before the change
in 136bc85 is introduced.

### DIFF
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,4 +1,4 @@
-dask
+dask[array]
 pytest>=3.5
 pytest-black
 pytest-cov
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,3 +1,4 @@
+dask
 pytest>=3.5
 pytest-black
 pytest-cov
--- a/sparse/_coo/core.py
+++ b/sparse/_coo/core.py
@@ -284,6 +284,12 @@ class COO(SparseArray, NDArrayOperatorsMixin):  # lgtm [py/missing-equals]
         self.coords, self.data, self.shape, self.fill_value = state
         self._cache = None
 
+    def __dask_tokenize__(self):
+        "Produce a deterministic, content-based hash for dask."
+        from dask.base import normalize_token
+        return normalize_token(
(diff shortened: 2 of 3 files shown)


requirements/tests.txt
```diff
@@ -1,3 +1,4 @@
+dask
```
hameerabbasi: `dask[array]` I believe?

danielballan: I don't think we are using anything from `dask.array`. The only imports are from `dask.base`.

danielballan: Oops, I was completely wrong about that. Indeed, `dask[array]` is what we needed. Pushed new commit. :-D

hameerabbasi:
Thanks, @danielballan

PR #321, opened 2020-02-06
Fix setting into numpy slice

danielballan:
Includes test that exercises the bug. Illustration:

```
In [1]: import numpy                                                                                                                                                                          

In [2]: import sparse                                                                                                                                                                         

In [3]: a = numpy.zeros((5, 5))                                                                                                                                                               

In [4]: b = sparse.COO(data=[1, 1], coords=(2, 4), shape=(5,))                                                                                                                                

In [5]: a[:, 0] = b                                                                                                                                                                           
TypeError                                 Traceback (most recent call last)
<ipython-input-5-ebf8ebabc230> in <module>
----> 1 a[:, 0] = b

TypeError: __array__() takes 1 positional argument but 2 were given

In [6]: %env SPARSE_AUTO_DENSIFY                                                                                                                                                              
Out[6]: '1'
```

As shown, `SPARSE_AUTO_DENSIFY` was set to `1`. The same error occurs with `SPARSE_AUTO_DENSIFY` unset. It _should_ raise `RuntimeError: Cannot convert a sparse array....` instead.

While I was here, I noticed that the test `test_failed_densification` seemed to be actually testing _successful_ densification. I took the liberty of renaming it and adding a second test that actually tests for failure.

### DIFF
--- a/sparse/_sparse_array.py
+++ b/sparse/_sparse_array.py
@@ -215,7 +215,7 @@ class SparseArray:
         True
         """
 
-    def __array__(self, **kwargs):
+    def __array__(self, *args, **kwargs):
         from ._settings import AUTO_DENSIFY
 
         if not AUTO_DENSIFY:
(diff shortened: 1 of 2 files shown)


hameerabbasi:
Thanks, @danielballan.

PR #322, opened 2020-02-07
Implement asarray.

### LINKED ISSUE #319: A densifying function that accepts sparse and dense inputs
Thanks for this great library!

I would like the ability to create certain sparse arrays that can be handled as `__array__`-protocol-abiding (auto-densifying) arrays without changing a global setting that would cause _all_ sparse arrays in the current process, including ones that might be managed by other libraries, to auto-densify.

In my use case, the arrays are small enough to fit in memory, so the benefit of the sparse representation is faster performance for certain operations. At the end, I may densify them so that standard visualization tools like `plt.imshow` know what to do with them. The value of the `__array__` protocol would be greatly diminished if I had to program each consumer check the type of the input and conditionally call `todense()`.

To configure auto-densification on a per-instance basis, we could add a parameter to `__init__` or define a subclass. Either way, I think it would be best if the setting were immutable after instantiation: it should be easy to know which Kind Of Thing I have and not worry about it changing under us.

Looking through the [issues and PRs mentioning densification](https://github.com/pydata/sparse/issues?utf8=%E2%9C%93&q=densify) I see that there has been significant prior discussion and development on this subject. I like @shoyer and @asmeurer's suggestions at the bottom of https://github.com/pydata/sparse/issues/218:

> Ideally the method would also be available on non-sparse arrays

> At the very least it would be nice to have sparse.todense() as a function so that we can call it universally on sparse and dense inputs.

That would be useful as well. If this seems worthwhile, I would be happy to submit PRs.

danielballan:
This is modeled on ``cupy.asarray``. Future work may use numpy protocols
to let duck arrays of sparse arrays pass through properly, such as a
dask array of sparse arrays. See #319 and issues linked therein for
discussion. Closes #319

### DIFF
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -438,3 +438,21 @@ def outer(a, b, out=None):
            [0, 3, 6, 9]])
     """
     return np.multiply.outer(a, b, out=out)
+
+
+def asnumpy(a, dtype=None, order=None):
+    """Returns a dense numpy array from an arbitrary source array.
+
+    Args:
+        a: Arbitrary object that can be converted to :class:`numpy.ndarray`.
+        order ({'C', 'F', 'A'}): The desired memory layout of the output
+            array. When ``order`` is 'A', it uses 'F' if ``a`` is
+            fortran-contiguous and 'C' otherwise.
+    Returns:
+        numpy.ndarray: Converted array on the host memory.
+    """
+    from ._sparse_array import SparseArray
+
+    if isinstance(a, SparseArray):
+        a = a.todense()
+    return np.array(a, dtype=dtype, copy=False, order=order)
--- a/sparse/tests/test_coo.py
+++ b/sparse/tests/test_coo.py
@@ -2226,3 +2226,14 @@ def test_flatten(in_shape):
     e = x.flatten()
 
     assert_eq(e, a)
+
+
+def test_asnumpy():
+    s = sparse.COO(data=[1], coords=[2], shape=(5,))
+    assert_eq(sparse.asnumpy(s), s.todense())
+    assert_eq(
+        sparse.asnumpy(s, dtype=np.float64), np.asarray(s.todense(), dtype=np.float64)
+    )
+    a = np.array([1, 2, 3])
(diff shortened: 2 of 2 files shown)


danielballan:
Closes #319

sparse/_common.py
```diff
@@ -438,3 +438,33 @@ def outer(a, b, out=None):
            [0, 3, 6, 9]])
     """
     return np.multiply.outer(a, b, out=out)
+
+
+def _is_dask_collection(x):
```
hameerabbasi: We don't need this function anymore.

sparse/_common.py
```diff
@@ -438,3 +438,33 @@ def outer(a, b, out=None):
...
+
+    Vendored from dask.base
+    """
+    try:
+        return x.__dask_graph__() is not None
+    except (AttributeError, TypeError):
+        return False
+
+
+def asarray(a, dtype=None, order=None):
```
hameerabbasi: Should be named `asnumpy`

sparse/_common.py
```diff
@@ -438,3 +438,33 @@ def outer(a, b, out=None):
...
+            array. When ``order`` is 'A', it uses 'F' if ``a`` is
+            fortran-contiguous and 'C' otherwise.
+    Returns:
+        numpy.ndarray: Converted array on the host memory.
+    """
+    from ._sparse_array import SparseArray
+
+    if isinstance(a, SparseArray):
+        a = a.todense()
+    return np.array(a, dtype=dtype, copy=False, order=order)
```
hameerabbasi: Should be `np,asarray` because `np.array` always copies, which is wasteful.

danielballan: I believe that `copy=False` prevents the copy, as verified by this part of the test:

```py
    a = np.array([1, 2, 3])
    # Array passes through with no copying.
    assert sparse.asarray(a) is a
```

danielballan:
Thanks for the review. Sorry for the messy first pass.

hameerabbasi:
Thanks, @danielballan!

PR #324, opened 2020-02-11
Try to simplify the numba extension a little

eric-wieser:
Rather than using `contextlib.ExitStack` to manage our state, this uses LLVM basic blocks instead.

cc @stuartarchibald

---

Unfortunately this didn't appear to improve compile-time performance.

### DIFF
--- a/sparse/_coo/numba_extension.py
+++ b/sparse/_coo/numba_extension.py
@@ -13,13 +13,12 @@ from numba.extending import (
     NativeValue,
     make_attribute_wrapper,
     type_callable,
-    lower_builtin,
 )
-from numba.targets.imputils import impl_ret_borrowed, lower_constant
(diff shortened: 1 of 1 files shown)


hameerabbasi:
CI fails with: `NameError: name 'lower_builtin' is not defined`.

hameerabbasi:
Is this ready for review?

eric-wieser:
> Is this ready for review?

Yeah, I think so. There are some open questions about `NativeValue.cleanup` that I was hoping to resolve here, but I'm realizing that that opens a can of worms that numba needs to look at - so I'll leave that to a follow-up.

hameerabbasi:
Pinging @stuartarchibald for a review.

hameerabbasi:
I'll wait a day or two for review before merging.

hameerabbasi:
Whoops, forgot about this one. Thanks, @eric-wieser!

PR #327, opened 2020-02-17
Add Code of Conduct

hameerabbasi:


.github/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,3 @@
+# Code of Conduct
+
+Please see [`docs/conduct.md`](../docs/conduct.rst)
```
rgommers: Better to hard-link to the rendered docs?

rgommers:
@hameerabbasi a major part of adding a CoC is expressing _common_ values within the project, so it would have been good to leave this open for discussion for a bit.

Any reasonable CoC is better than no CoC of course. Just curious why you decided to go with the Contributor Covenant rather than thee NumFOCUS or SciPy ones in the end?

hameerabbasi:
A factor was that the NumFocus and SciPy ones had a lot of interconnected links that were rather hard to change, and a lot of rather stringent requirements I wasn’t sure I’d be able to meet alone. I could just link to the SciPy/NumFocus ones, delegating the enforcement to them. I’m fine with that too.

rgommers:
Fair enough. Maybe staying with Contributor Covenant for now, and moving to a nicer / more elaborate CoC once the developer and user community grows, is the right thing to do.

PR #328, opened 2020-03-03
Respect `copy=False` in `astype`

eric-wieser:
Previously this would make a copy anyway.

### DIFF
--- a/sparse/_coo/core.py
+++ b/sparse/_coo/core.py
@@ -2186,6 +2186,9 @@ class COO(SparseArray, NDArrayOperatorsMixin):  # lgtm [py/missing-equals]
         :obj:`COO.elemwise`: Apply an arbitrary element-wise function to one or two
             arguments.
         """
+        # this matches numpy's behavior
+        if self.dtype == dtype and copy == False:
+            return self
         return self.__array_ufunc__(
             np.ndarray.astype, "__call__", self, dtype=dtype, copy=copy
         )
--- a/sparse/_coo/core.py
+++ b/sparse/_coo/core.py
@@ -2187,7 +2187,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):  # lgtm [py/missing-equals]
             arguments.
         """
         # this matches numpy's behavior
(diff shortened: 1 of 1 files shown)


sparse/_coo/core.py
```diff
@@ -2186,6 +2186,9 @@ def astype(self, dtype, copy=True):
         :obj:`COO.elemwise`: Apply an arbitrary element-wise function to one or two
             arguments.
         """
+        # this matches numpy's behavior
+        if self.dtype == dtype and copy == False:
```
perimosocordiae: Comparing dtypes can be tricky. To be safe:

```
if not copy and self.dtype.type == np.dtype(dtype).type:
    return self
```

eric-wieser: Your suggestion doesn't match the behavior of regular numpy arrays.

Worse, it does the wrong thing for reversed-endian types.

hameerabbasi: Comparing booleans to booleans is a weird pattern.

```suggestion
        if self.dtype == dtype and not copy:
```

eric-wieser: This makes `coo.astype(int, copy=[])` succeed, whereas the way I've written it it gives `TypeError: an integer is required (got type list)`

hameerabbasi: I'd still prefer that, anyone who writes that (or equivalent) code is in much hotter waters.

perimosocordiae: Good point. I'll amend my recommendation to only coerce the user input, then:

```
if not copy and self.dtype == np.dtype(dtype):
```

eric-wieser: Feel free to commit this change yourself if you feel that way. I'd usually lean towards preserving numpy errors when possible, simply because it makes it easier for users to spot typos.

eric-wieser: That coercion happens automatically anyway, and is probably faster at the C level where it doesn't need a python function call.

eric-wieser:
Thanks!

hameerabbasi:
Thank you, @eric-wieser!

PR #330, opened 2020-03-03
Replace linear_loc with ravel_multi_index, which is 3x faster

eric-wieser:
```
In [42]: %timeit np.ravel_multi_index(o.coords, o.shape)
11.2 µs ± 1.32 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [43]: %timeit sparse._coo.common.linear_loc(o.coords, o.shape)
35.2 µs ± 2.84 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

### DIFF
--- a/sparse/_coo/common.py
+++ b/sparse/_coo/common.py
@@ -54,14 +54,14 @@ def asCOO(x, name="asCOO", check=True):
 
 
 def linear_loc(coords, shape):
-    out = np.zeros(coords.shape[1], dtype=np.intp)
-    tmp = np.zeros(coords.shape[1], dtype=np.intp)
-    strides = int(1)
-    for i, d in enumerate(shape[::-1]):
-        np.multiply(coords[-(i + 1), :], strides, out=tmp)
-        np.add(tmp, out, out=out)
-        strides *= d
-    return out
+    if shape == () and len(coords) == 0:
+        # `np.ravel_multi_index` is not aware of arrays, so cannot produce a
+        # sensible result here (https://github.com/numpy/numpy/issues/15690).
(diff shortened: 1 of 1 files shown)


eric-wieser:
Failure caused by numpy/numpy#15690

sparse/_coo/common.py
```diff
@@ -53,17 +53,6 @@ def asCOO(x, name="asCOO", check=True):
...
 
-def linear_loc(coords, shape):
-    out = np.zeros(coords.shape[1], dtype=np.intp)
-    tmp = np.zeros(coords.shape[1], dtype=np.intp)
-    strides = int(1)
-    for i, d in enumerate(shape[::-1]):
-        np.multiply(coords[-(i + 1), :], strides, out=tmp)
-        np.add(tmp, out, out=out)
-        strides *= d
-    return out
```
eric-wieser: @hameerabbasi: want me to revise this as:
```python
def ravel_multi_index(coords, shape):
    if shape == ():
        # https://github.com/numpy/numpy/issues/15690
        assert coords.shape[0] == 0
        return np.zeros(coords.shape[1:], dtype=np.intp)
    else:
        return np.ravel_multi_index(coords, shape)
```

hameerabbasi: Yes. I arrived at the same conclusion you did, and you beat me to reporting the bug.

eric-wieser: Want me to rename it or leave the function name as `linear_loc`?

hameerabbasi: Either is fine with me.

eric-wieser:
Left the name the same, it makes the patch much smaller

PR #343, opened 2020-05-14
Added multi-axis advanced indexing support

mikeymezher:
Added support for multi-axis advanced indexing. Speed is comparable to single axis advanced indexing.


hameerabbasi:
Thanks! This is missing some tests, those are pretty much essential.

mikeymezher:
It seems like the only failed test is test_slicing_errors(index). 
Seems like an issue with the tests. This is likely because an IndexError is no longer being raised if len(adv_idx) is greater than 1 in _coo.indexing.py

What's the protocol for modifying the test cases? 

Would you be able to handle this?

hameerabbasi:
> Seems like an issue with the tests. This is likely because an IndexError is no longer being raised if len(adv_idx) is greater than 1 in _coo.indexing.py

That's because the test was "testing" if we didn't have advanced indexing at all. You can get rid of the bad tests.

hameerabbasi:
> What's the protocol for modifying the test cases?

The contributor usually adds tests, and the maintainer/reviewer reviews the code and tests.

mikeymezher:
Alright, I'll add a few test cases to test_slicing(index) and remove test_slicing_error are submit a new PR

hameerabbasi:
Oh, you can just push to your branch and it'll modify this PR. Don't need to close at all.

mikeymezher:
Bug found in returned coordinates. Will attempt to fix tomorrow

multi_adv_ix_test.py
```diff
@@ -0,0 +1,36 @@
+import sparse
```
hameerabbasi, line 1: Can we remove this file?

multi_adv_ix_test2.py
```diff
@@ -0,0 +1,34 @@
+import sparse
```
hameerabbasi, line 1: And this one?

multi_adv_ix_test.py
mikeymezher: Yes sorry, I removed them in latest commit. Those were there just for local testing and I had forgot to remove them.

hameerabbasi: No need to apologize. 😄 I tend to be a bit too quick in mt reviews.

sparse/_coo/indexing.py
```diff
@@ -387,6 +464,8 @@ def _compute_mask(coords, indices):  # pragma: no cover
     n_matches = np.intp(coords.shape[1])
 
     i = 0
+    #print('COORD SHAPE: {}'.format(coords.shape))
+    #print('IXS SHAPE: {}'.format(indices.shape))
```
hameerabbasi, line 468: Can we remove these lines?

hameerabbasi:
Thanks, @mikeymezher!

Issue #348, opened 2020-05-17
sparse output from tensordot


lueckem:
Hello!
If one of the arguments of sparse.tensordot is a dense np.ndarray, then the output will be dense as well. However, the result of a tensordot between a very sparse and a dense array can still be fairly sparse.
In one of my projects I deal with very large and sparse arrays, so that the tensordot with a small dense array results in a memory error. This would not be the case if tensordot had the option to return a sparse array in this case.

So the new feature that I request is to add an argument to sparse.tensordot to specify if the output should be returned as a np.ndarray or as a COO sparse array.
If you are open to this change, I already implemented this feature for the case that the first argument is dense and the second argument is sparse. I can open a pull request where further implementation can be discussed.
The performance is not that good (in my benchmark returning sparse is ~4 times slower than the version that returns dense), but performance is not the point of this new feature.

Regards

hameerabbasi:
Hello,

Yes, you can make a PR if you like, I'm willing to accept this change.

hameerabbasi:
Solved by #349.

PR #349, opened 2020-05-17
Choose sparse/dense type when doing tensordot

lueckem:
Enhancement of sparse.tenserdot: choose return type. See #348.
The goal is to let the user decide of which type the output of tensordot is. The options are returntype="sparse", "dense" or "auto" (sparse if both arguments are sparse, otherwise dense).

To date only works for returntype="sparse" in the case that the first array is dense and the second array is sparse. The other cases will be implemented in the future.

I am very thankful for suggestions for improvement, especially concerning perfomance.

### DIFF
--- a/sparse/_coo/common.py
+++ b/sparse/_coo/common.py
@@ -314,13 +314,13 @@ def _dot(a, b, return_type=None):
     if isinstance(a, np.ndarray) and isinstance(b, COO):
         b = b.T
         a = a.view(type=np.ndarray)
+
         if return_type == COO:
-            coords, data, dtype = _dot_ndarray_coo_type_sparse(a.dtype, b.dtype)(
+            coords, data = _dot_ndarray_coo_type_sparse(a.dtype, b.dtype)(
                 a, b.coords, b.data, out_shape
             )
-            coords = np.array(coords).T
-            data = np.array(data, dtype=dtype)
-            return COO(coords, data, shape=out_shape, prune=True)
+            return COO(coords, data, shape=out_shape, has_duplicates=False, sorted=True)
+
         return _dot_ndarray_coo_type(a.dtype, b.dtype)(a, b.coords, b.data, out_shape)
 
 
@@ -1324,14 +1324,28 @@ def _dot_ndarray_coo_type_sparse(dt1, dt2):
         out_data = []
         out_coords = []
 
-        for oidx1 in range(out_shape[0]):
-            for didx2 in range(len(data2)):
-                oidx2 = coords2[0, didx2]
-                out_coords.append([oidx1, oidx2])
-                out_data.append(array1[oidx1, coords2[1, didx2]] * data2[didx2])
-                # attention: data can be 0.0
+        # coords2.shape = (2, len(data2))
+        # coords2[0, :] = columns, sorted
+        # coords2[1, :] = rows
 
-        return out_coords, out_data, dtr
+        for oidx1 in range(out_shape[0]):
+            data_curr = 0.0
+            current_col = 0
+            for didx2 in range(len(data2)):
+                if coords2[0, didx2] != current_col:
+                    if data_curr != 0.0:
+                        out_data.append(data_curr)
+                        out_coords.append([oidx1, current_col])
+                        data_curr = 0.0
+                    current_col = coords2[0, didx2]
+
+                data_curr += array1[oidx1, coords2[1, didx2]] * data2[didx2]
+
+            if data_curr != 0.0:
+                out_data.append(data_curr)
+                out_coords.append([oidx1, current_col])
+
+        return np.array(out_coords).T, np.array(out_data, dtype=dtr)
 
     return _dot_ndarray_coo
 
--- a/sparse/_coo/common.py
+++ b/sparse/_coo/common.py
@@ -1166,44 +1166,45 @@ def _dot_coo_coo_type(dt1, dt2):
         coords_out = []
         data_out = []
         didx1 = 0
+        data1_end = len(data1)
+        data2_end = len(data2)
 
-        while didx1 < len(data1):
+        while didx1 < data1_end:
             oidx1 = coords1[0, didx1]
             didx2 = 0
             didx1_curr = didx1
 
             while (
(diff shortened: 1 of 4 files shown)


hameerabbasi:
Can we raise a `NotImplementedError` for unsupported configurations at present?

sparse/_coo/common.py
```diff
@@ -64,7 +64,7 @@ def linear_loc(coords, shape):
         return np.ravel_multi_index(coords, shape)
 
 
-def tensordot(a, b, axes=2):
+def tensordot(a, b, axes=2, returntype="auto"):
```
hameerabbasi, line 67: I'd change this to `return_type=None`

sparse/_coo/common.py
```diff
@@ -74,6 +74,9 @@ def tensordot(a, b, axes=2):
         The arrays to perform the :code:`tensordot` operation on.
     axes : tuple[Union[int, tuple[int], Union[int, tuple[int]], optional
         The axes to match when performing the sum.
+    returntype : {"auto", "dense", "sparse"}, optional
```
hameerabbasi, line 77: `{None, COO, numpy.ndarray}`

sparse/_coo/common.py
```diff
@@ -1287,6 +1299,42 @@ def _dot_ndarray_coo(array1, coords2, data2, out_shape):  # pragma: no cover
...
+        data2, coords2 : np.ndarray
+            The data and coordinates of ``s2``.
+
+        out_shape : Tuple[int]
+            The output shape.
+        """
+        out_data = []
+        out_coords = []
+
+        for oidx1 in range(out_shape[0]):
```
hameerabbasi, line 1326: I'd perform the accumulation inline rather than relying on the deduplication/coalescing. I believe this is what is slowing down your performance.

sparse/_coo/common.py
```diff
@@ -1287,6 +1299,42 @@ def _dot_ndarray_coo(array1, coords2, data2, out_shape):  # pragma: no cover
...
+            for didx2 in range(len(data2)):
+                oidx2 = coords2[0, didx2]
+                out_coords.append([oidx1, oidx2])
+                out_data.append(array1[oidx1, coords2[1, didx2]] * data2[didx2])
```
hameerabbasi, line 1330: As well as filtering zeros here.

lueckem:
> Can we raise a `NotImplementedError` for unsupported configurations at present?

I will implement them in the next few days, so maybe we could just keep the PR open until then?

hameerabbasi:
> I will implement them in the next few days, so maybe we could just keep the PR open until then?

Of course. Feel free to work on it when you like. I may be quick to respond; you don't have to be. 😄

sparse/_coo/common.py
lueckem: Thanks for the suggestions! Could you please elaborate on this one?

hameerabbasi: Well, matrix multiplies are a multiply of some elements plus a sum over a given axis, right? What you're doing is adding all the multiplies in, and relying on the deduplication to do the sum. I would also do the sum here as well, and filter out zero entries from the get-go. This would improve performance by a lot.

sparse/_coo/common.py
```diff
@@ -303,12 +307,20 @@ def _dot(a, b):
...
     if isinstance(a, np.ndarray) and isinstance(b, COO):
         b = b.T
         a = a.view(type=np.ndarray)
+        if return_type == COO:
+            coords, data, dtype = _dot_ndarray_coo_type_sparse(a.dtype, b.dtype)(
+                a, b.coords, b.data, out_shape
+            )
+            coords = np.array(coords).T
+            data = np.array(data, dtype=dtype)
+            return COO(coords, data, shape=out_shape, prune=True)
```
hameerabbasi, line 323: In relation to https://github.com/pydata/sparse/pull/349/files#r426539625, don't rely on `prune=True` here.

sparse/_coo/common.py
lueckem: I am not sure what you mean exactly, but the summing seems to be optimal already because for every row of the output (oidx1) we only iterate over the nonzero elements of the sparse array once (didx). How can you improve that? Could you provide some code or formula for your idea?

To the filtering out: I already tried that and it didnt seem to have any effect on the performance in the case that I tested. As the first array is dense, we should assume that it does not have a lot of zero entries, maybe none at all. Thus, the test for zero would almost always be negative. The additional overhead created by this test seems to be of the same order as the one call of prune in the creation of the COO array.

If you prefer, I can switch to the version with tests for zero and without prune, but I have not seen a performance increase in that change.

sparse/_coo/common.py
hameerabbasi: I'd say your tests should pass with `has_duplicates=False, sorted=True`

sparse/_coo/common.py
hameerabbasi: You are adding way more elements than necessary IIUC, one element is added multiple times. If you follow my suggestion in the other comment you'll see that.

lueckem: You are right! I updated the algorithm accordingly and the performance now matches the return_type=None version.
Unfortunately the test fails in Python 3.6. Any idea why?

sparse/_coo/common.py
```diff
@@ -64,7 +64,7 @@ def linear_loc(coords, shape):
         return np.ravel_multi_index(coords, shape)
 
 
-def tensordot(a, b, axes=2):
+def tensordot(a, b, axes=2, return_type=None):
```
hameerabbasi, line 67:
```suggestion
def tensordot(a, b, axes=2, *, return_type=None):
```

hameerabbasi: This will make it so that if NumPy adds any additional `kwargs`, we remain compatible in the future.

lueckem:
I added all the missing implementations. The performance of the newly added implementations match the old performance for return_type=None. I added some benchmarks for comparison.

hameerabbasi:
This is in, thanks so much @lueckem!

PR #350, opened 2020-05-18
Improve dot performance.

hameerabbasi:


hameerabbasi:
On my machine this brings down the time needed to do a matrix multiply on two `(5000, 5000)`-shaped inputs with density `0.01` from 12.4 seconds to 9.1 seconds.

mrocklin:
Nice.

PR #357, opened 2020-06-11
@box and @unbox for dense level

guilhermeleobas:
The following now works

```python
@njit
def foo(N, ordered, unique):
    d = Dense(N=N, ordered=ordered, unique=unique)
    return d

@njit
def bar(d):
	print(d.N)

d = foo(3, True, True)
bar(d) # prints 3
```


hameerabbasi:
Thanks, @guilhermeleobas!

PR #362, opened 2020-06-20
added method moveaxis

### LINKED ISSUE #361: No implementation for 'numpy.moveaxis'
**Describe the bug**
There is currently no moveaxis implementation for sparse tensors.

**Minimal code to Reproduce**
```python
import sparse
import numpy as np

shape = (10, 10, 10)

tensor = sparse.random(shape)

np.moveaxis(tensor, 1, 0)
```

**Expected behavior**
This should permute the dimensions as documented in the [numpy.moveaxis](https://numpy.org/doc/1.18/reference/generated/numpy.moveaxis.html) function.

**System**
 - OS and version: osx
 - `sparse` version: '0.10.0' (also verified with '0.10.0+9.g24d26c8')
 - NumPy version: '1.18.1'
 - Numba version: '0.50.0'

**Additional context**
Moveaxis is used in TensorLy: https://github.com/tensorly/tensorly/issues/131
Currently, a workaround is being used: 
https://github.com/tensorly/tensorly/blob/f9be446ec9f7fd5827dcbab8efad58ead99d124e/tensorly/contrib/sparse/backend/numpy_backend.py#L31

lueckem:
Fixes #361. I added the method moveaxis to COO for numpy compatibility.

I more or less copied the code from numpy, do we have to acknowledge that somewhere?

### DIFF
--- a/docs/generated/sparse.COO.rst
+++ b/docs/generated/sparse.COO.rst
@@ -77,7 +77,6 @@ COO
       COO.resize
       COO.transpose
       COO.swapaxes
-      COO.moveaxis
       COO.nonzero
 
    .. rubric:: Utility functions
--- a/docs/generated/sparse.COO.rst
+++ b/docs/generated/sparse.COO.rst
@@ -77,6 +77,7 @@ COO
       COO.resize
       COO.transpose
       COO.swapaxes
+      COO.moveaxis
       COO.nonzero
 
    .. rubric:: Utility functions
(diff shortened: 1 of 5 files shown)


sparse/_coo/core.py
```diff
@@ -1561,6 +1561,53 @@ def swapaxes(self, axis1, axis2):
         axes[axis1], axes[axis2] = axes[axis2], axes[axis1]
         return self.transpose(axes)
```
hameerabbasi, line 1563: I'd add a comment here indicating where the code was taken from, including a link and line numbers.

docs/generated/sparse.COO.rst
```diff
@@ -77,6 +77,7 @@ COO
       COO.resize
       COO.transpose
       COO.swapaxes
+      COO.moveaxis
```
hameerabbasi, line 80: This will fail to build docs, you should move it to the main module.

lueckem: yes, thanks for the reminder!

PR #382, opened 2020-07-11
GCXS Slicing and reshape/resize/transpose/change_compressed_axes

daletovar:
reshape/resize/transpose/change_compressed_axes for GCXS:
- previously we just converted to COO to do all of these operations, which is pretty memory inefficient. 
- all of these operations can be done with variations on the same algorithm.
- I have plans to add np.unravel_index and np.ravel_multi_index to numba. For now, I implemented a small subset of their functionality in sparse.

Slicing and fancy indexing for GCXS:
- for indexing with positive slices and sorted lists/arrays along compressed dimensions we can make some simplifying assumptions and get a lot of speedup.


hameerabbasi:
Is this ready for review?

sparse/_compressed/compressed.py
```diff
@@ -237,6 +168,33 @@ def nbytes(self):
...
+
+    @property
+    def compressed_shape(self):
+        row_size = np.prod(self.reordered_shape[: self.axisptr])
+        col_size = np.prod(self.reordered_shape[self.axisptr :])
+        return (row_size, col_size)
+
+    @property
+    def reordered_shape(self):
+        return np.array(self.shape)[self.axis_order]
```
hameerabbasi, line 192: Can we make all of these private API?

hameerabbasi, line 192: Nit: We shouldn't use NumPy for manipulating shapes, it's a lot slower than using Python directly, and it sort-of implies heavy lifting. For example, I'd replace
```
np.array(self.shape)[self.axis_order]
```

with
```
tuple(self.shape[i] for i in axis_order)
```

sparse/_compressed/compressed.py
```diff
@@ -248,7 +206,13 @@ def __str__(self):
...
     def change_compressed_axes(self, new_compressed_axes):
         """
-        changes the compressed axes of an array.
+        Changes the compressed axes of this array. This operation is similar to converting 
+        a scipy.sparse.csc_matrix to a scipy.sparse.csr_matrix.
+
+        Returns
+        -------
+        GCXS
+            The output array.
```
hameerabbasi, line 215: Can we document that this doesn't modify in-place but returns a transposed version.

sparse/_compressed/compressed.py
daletovar: How do I do that? Do you mean something like `def _axisptr(self):`?

daletovar: Thanks, this is helpful.

hameerabbasi: Yes. In Python, anything prepended with an underscore is implicitly marked as private.

daletovar: I see. Thank you.

sparse/_compressed/convert.py
```diff
@@ -58,3 +63,250 @@ def uncompress_dimension(indptr):  # pragma: no cover
...
+    """
+    for i in range(len(arr) - 1):
+        if arr[i + 1] <= arr[i]:
+            return False
+    return True
+
+
+def _1d_reshape(x, shape, compressed_axes):
+    @numba.jit(nopython=True, nogil=True)
+    def _linearize(
```
hameerabbasi, line 82: You want to either define this outside the function (if possible) or somehow cache it. The reason is that Numba/LLVM don't properly garbage collect this function once it goes out of memory.

daletovar: That's helpful. Thanks

sparse/_compressed/indexing.py
```diff
@@ -62,13 +62,13 @@ def getitem(x, key):
         count += 1
 
     # reorder the key according to the axis_order of the array
-    reordered_key = [Nones_removed[i] for i in x.axis_order]
+    reordered_key = [Nones_removed[i] for i in x._axis_order]
```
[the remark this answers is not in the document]
daletovar: Thanks @hameerabbasi. Some of the values get of `reordered_key` get modified shortly after this. I think it makes sense to keep it as a list.

daletovar:
I'm not sure why the MacOS test was cancelled.

hameerabbasi:
Merging as CI is unrelated.

hameerabbasi:
No, wait. It re-ran.

daletovar:
Gotcha. Perfect.

hameerabbasi:
Thanks, @daletovar.

Issue #331, opened 2020-03-10
Performance bencmarks comparison with scipy.sparse


rth:
I see there is an asv setup in the repo but I couldn't find any results.

How does pydata/sparse currently compare to scipy.sparse (e.g. for dot product)?  Or is most of linalg operations expected to use `scipy.sparse.linlag` in the long run and so the performance would be equivalent?

hameerabbasi:
Tl;dr: It's bad. Sometimes 10 times worse, even for element-wise operations.

We're looking at approaches that essentially guarantee optimal performance and GPU support in all cases (#326). However, we don't have a ETA for this.

hameerabbasi:
```ipython
In [3]: x = sparse.random((5000, 5000), density=0.001)

In [4]: y = sparse.random((5000, 5000), density=0.001)

In [5]: %timeit x + y  # Numba compilation
5.86 ms ± 190 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit x + y
5.97 ms ± 70.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [7]: xs = x.tocsr()

In [8]: ys = y.tocsr()

In [9]: %timeit xs + ys
408 µs ± 8.44 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

hameerabbasi:
For `__matmul__`:

```ipython
In [11]: %timeit x @ y
1.05 s ± 18.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [12]: %timeit xs @ ys
1.09 ms ± 17.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

rth:
Thanks for the explanations @hameerabbasi !

rgommers:
It seems like `matmul` could easily be improved though right? I just had a peek at the SciPy implementation (`csr_matmul` in `sparsetools/csr.h`) and it's a relatively straightforward triple for-loop in C++. So I guess no one has put much effort into a similar Numba implementation here, otherwise it wouldn't be 1000x slower.

rgommers:
Not that I'm suggesting to optimize that right now, I just want to say there's no fundamental reason why it's slow, just no one got around to it since this package is still new.

hameerabbasi:
[...]

PR #389, opened 2020-08-03
Improve dot performance

daletovar:
- improves dot for two COO arrays, addresses #331 
    - I used the same algorithm that scipy uses.
    - It's a bit slower than scipy because to ensure sorted coordinates, we have to do some sorting.

- add dot for GCXS arrays
    -  Uses the same algorithm.
    - Works for `csr @ csr` and  `csc @ csc`
    - I found that the `csr @ csc` dot product-based algorithm was just to slow, so we convert one of the arrays. 

- add io for GCXS arrays


sparse/_compressed/indexing.py
```diff
@@ -232,6 +233,12 @@ def get_slicing_selection(
     indices = np.array(indices)
     data = arr_data[ind_list]
     return (data, indices, indptr)
+=======
+        arg,
+        shape=shape,
+        compressed_axes=compressed_axes,
+        fill_value=x.fill_value)
+>>>>>>> change gxcs with gcxs
```
hameerabbasi, line 241: Merging issue here.

rgommers:
> It's a bit slower than scipy because to ensure sorted coordinates, we have to do some sorting.

That sounds promising! Would be great to see some benchmark results.

rgommers:
@daletovar also looking at the PR description, there's a ton of stuff in here that looks like it should be split into multiple PRs. One topic (fix / feature / performance improvement ), one PR is the usual rule.

hameerabbasi:
>     * It's a bit slower than scipy because to ensure sorted coordinates, we have to do some sorting.

If you're converting from GCXS, no sorting should be necessary, in theory.

hameerabbasi:
LGTM caught some valid concerns -- Could you address those?

hameerabbasi:
And the code coverage is a bit lacking. 😕

daletovar:
Thanks @rgommers, I'll make sure to do that going forward.

daletovar:
With regards to the sorting, the issue is that the dot algorithm produces results that are not sorted:
```
In [1]: import scipy.sparse as ss
In [2]: x = ss.random(1000,1000,density=.2,format='csr')
In [3]: y = ss.random(1000,1000,density=.2,format='csr')
In [4]: res = x @ y
In [5]: res.indptr[1]
Out[5]: 1000
In [6]: res.indices
Out[6]: array([554, 328, 723, ...,  14,   7,   6], dtype=int32)
```
The indices for each row of the resulting matrix are out-of-order. So I added something like this:
```
for start, stop in zip(res.indptr[:-1], res.indptr[1:]):
    order = np.argsort(res.indices[start:stop])
    res.data[start:stop] = res.data[start:stop][order]
    res.indices[start:stop] = res.indices[start:stop][order]
```

daletovar:
@hameerabbasi, yes I'll see if I can fix those.

daletovar:
Here's an initial mark:
```
In [2]: x = sparse.random((5000, 5000), density=0.001)
In [3]: y = sparse.random((5000, 5000), density=0.001)
In [4]: xs = x.tocsr()
In [5]: ys = y.tocsr()
In [6]: x @ y # jit
Out[6]: <COO: shape=(5000, 5000), dtype=float64, nnz=124139, fill_value=0.0>
In [7]: %timeit x @ y
10.5 ms ± 104 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [8]: %timeit xs @ ys
1.14 ms ± 2.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

sparse/_common.py
```diff
@@ -25,11 +28,1137 @@
...
+                next_[temp] = -1
+                sums[temp] = 0
+
+            indptr[i + 1] = nnz
+            # ensure sorted indices
+            order = np.argsort(indices[indptr[i] : indptr[i + 1]])
+            data[indptr[i] : indptr[i + 1]] = data[indptr[i] : indptr[i + 1]][order]
+            indices[indptr[i] : indptr[i + 1]] = indices[indptr[i] : indptr[i + 1]][
+                order
+            ]
```
hameerabbasi, line 647: Can we possibly sort before putting in the data? I suspect it'll be a lot faster.

sparse/tests/test_dot.py
```diff
@@ -0,0 +1,243 @@
...
+
+    a = sa.todense()
+    b = sb.todense()
+
+    assert_eq(a.dot(b), sa.dot(sb))
+    assert_eq(np.dot(a, b), sparse.dot(sa, sb))
+    assert_eq(sparse.dot(sa, b), sparse.dot(a, sb))
+    assert_eq(np.dot(a, b), sparse.dot(sa, sb))
+
+    if hasattr(operator, "matmul"):
```
hameerabbasi, line 200: This was a compat check for Python 2. It can be removed.

sparse/_io.py
```diff
@@ -49,11 +50,17 @@ def save_npz(filename, matrix, compressed=True):
 
     nodes = {
         "data": matrix.data,
-        "coords": matrix.coords,
         "shape": matrix.shape,
         "fill_value": matrix.fill_value,
     }
 
+    if type(matrix) == COO:
```
hameerabbasi, line 57: Change docstring to say:

1. That it returns a `sparse.SparseArray`
2. That the format isn't stable at this time.

sparse/tests/test_io.py
```diff
@@ -11,8 +11,9 @@
 
 
 @pytest.mark.parametrize("compression", [True, False])
-def test_save_load_npz_file(compression):
-    x = sparse.random((2, 3, 4, 5), density=0.25)
+@pytest.mark.parametrize("format", ["coo", "gcxs"])
+def test_save_load_npz_file(compression, format):
```
hameerabbasi, line 15: Can you rewrite this using `tmp_path`: https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture

hameerabbasi: Same for the next test.

sparse/_common.py
daletovar: I'm not quite sure what you mean. Could you give an example?

hameerabbasi: Well, the original CSR-CSR multiplication algorithm doesn't require sorting, correct? I suspect this one only requires it because you're passing in unsorted data in the first place (or iterating over it in a way that's unsorted).

Could you possibly pre-sort the data so the output here will be sorted?

daletovar: The original CSR-CSR multiplication algorithm iterates over the data in a way that's unsorted. In scipy, the resulting csr_matrix has indices that are out-of-order:
```
In [13]: xs.has_sorted_indices
Out[13]: 1
In [14]: res = xs @ ys
In [15]: res.has_sorted_indices
Out[15]: 0
```
Since many of the operations for COO and GCXS rely on maintaining sorted coordinates, I was thinking it would be best to ensure that the result of dot has sorted coords.

hameerabbasi: In that case could you post a benchmark which includes sorting for the SciPy version?

sparse/_io.py
daletovar: To clarify, I'm noting that the GCXS format isn't stable at this time? Or that the format of the of the output array isn't stable? Thanks for your patience, Hameer.

sparse/_common.py
daletovar:
```
%%timeit
x @ y # coo coo
10.2 ms ± 51.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
res = xs @ ys
res.sort_indices()
4.05 ms ± 8.83 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
I'm going to test doing just one sort at the end to see how that does.

daletovar:
```
%%timeit
x @ y # coo coo - one sort at the end
7.55 ms ± 96.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
res = xs @ ys
res.sort_indices()
4.04 ms ± 15.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
It looks like doing one sort at the end is better.

hameerabbasi: Could you also post some GCXS-GCXS benchmarks for direct comparison with SciPy?

sparse/_io.py
hameerabbasi: My mistake, I should've been more clear: I mean that the output binary format of `save_npz`/`load_npz` isn't stable at this time.

daletovar: I see. Thanks for clarifying.

sparse/_common.py
daletovar:
```
%%timeit
x @ y # csr csr
6.32 ms ± 107 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

x = x.change_compressed_axes([1])
y = y.change_compressed_axes([1])
x @ y

%%timeit
x @ y # csc csc
6.11 ms ± 46.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

x = x.tocoo()
y = y.tocoo()
x @ y

%%timeit
x @ y # coo coo
7.23 ms ± 93.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
res = xs @ ys
res.sort_indices()
4.08 ms ± 49.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

daletovar:
Thanks so much for the review, @hameerabbasi. I think those were really helpful changes. Is there anything else you'd like to see?

sparse/tests/test_dot.py
```diff
@@ -197,14 +197,6 @@ def test_dot(a_shape, b_shape, a_format, b_format, a_comp_axes, b_comp_axes):
...
     assert_eq(np.dot(a, b), sparse.dot(sa, sb))
 
-    if hasattr(operator, "matmul"):
-        # Basic equivalences
-        assert_eq(operator.matmul(a, b), operator.matmul(sa, sb))
-        # Test that SOO's and np.array's combine correctly
-        # Not possible due to https://github.com/numpy/numpy/issues/9028
-        # assert_eq(eval("a @ sb"), eval("sa @ b"))
-
-
```
hameerabbasi: This should be added back, only the "if" was unnecessary.

hameerabbasi:
I have one final question: Does `dot` now return `GCXS` by default for COO/ndarray inputs?

And one possibility for future work: Writing a sparse-dense matrix multiplication kernel separately.

daletovar:
No, the default behavior is the same. I did make it so that GCXS @ COO results in a GCXS though.

> And one possibility for future work: Writing a sparse-dense matrix multiplication kernel separately.

+1

I'm away from my computer the next couple of days so I apologize if I don't get back right away about anymore suggestions.

hameerabbasi:
Thanks, @daletovar!

PR #397, opened 2020-08-19
add reductions

daletovar:
This PR adds reduction operations for GCXS arrays. I can add mean, var, and std after I add elementwise operations.


sparse/_compressed/compressed.py
```diff
@@ -1,18 +1,26 @@
...
+    normalize_axis,
+    equivalent,
+    check_zero_fill_value,
+    check_compressed_axes,
+)
 from .._coo.core import COO
 from .convert import uncompress_dimension, _transpose, _1d_reshape
 from .indexing import getitem
 
+_reduce_super_ufunc = {np.add: np.multiply, np.multiply: np.power}
```
hameerabbasi, line 22: This should be in a central place somewhere, not duplicated.

sparse/_compressed/compressed.py
```diff
@@ -700,3 +708,344 @@ def __rmatmul__(self, other):
...
+            if out.shape != result.shape:
+                raise ValueError(
+                    "non-broadcastable output operand with shape %s "
+                    "doesn't match the broadcast shape %s" % (out.shape, result.shape)
+                )
+
+            out._make_shallow_copy_of(result)
+            return out
+
+        return result
```
hameerabbasi, line 752: Can we also factor out the common logic here?

sparse/_compressed/compressed.py
```diff
@@ -700,3 +708,344 @@ def __rmatmul__(self, other):
...
+        if keepdims:
+            shape = list(self.shape)
+            for ax in axis:
+                shape[ax] = 1
+            out = out.reshape(shape)
+
+        if out.ndim == 0:
+            return out[()]
+
+        return out
```
hameerabbasi, line 871: Same here.

sparse/_compressed/compressed.py
```diff
@@ -700,3 +708,344 @@ def __rmatmul__(self, other):
...
+            The reduced output sparse array.
+
+        See Also
+        --------
+        :obj:`numpy.prod` : Equivalent numpy function.
+        COO.prod : Equivalent operation on COO arrays.
+        """
+        return np.multiply.reduce(
+            self, out=out, axis=axis, keepdims=keepdims, dtype=dtype
+        )
```
hameerabbasi, line 1026: It'd be nice to see all these on `SparseArray`, as the logic is common.

hameerabbasi (commented):
I've given this a shallow review so far. My comment in general is, we should factor out the common logic somewhere (copy-paste is bad). Where the implementations are the same we should move them to `SparseArray`. Where not the same, we should add them and `raise NotImplementedError`.

daletovar:
Thanks so much, @hameerabbasi. I moved everything to _sparse_array.

sparse/_sparse_array.py
```diff
@@ -259,3 +264,436 @@ def __array_function__(self, func, types, args, kwargs):
...
+                        self.data,
+                        reduce_super_ufunc(self.fill_value, self.size - self.nnz),
+                    )
+                    val = method.reduce(temp, **kwargs)
+                if keepdims:
+                    out = np.array(val)
+                    return out.reshape(np.ones(self.ndim, dtype=np.intp))
+                return val
+            else:  # coo
+                axis = tuple(range(self.ndim))
```
hameerabbasi, line 412: Can we move this to a method that does `raise NotImplementedError` on `SparseArray` and implement it on `COO` and `GCXS`

sparse/_sparse_array.py
```diff
@@ -259,3 +264,436 @@ def __array_function__(self, func, types, args, kwargs):
...
+            out = type(self)(
+                coords,
+                data,
+                shape=(a.shape[0],),
+                has_duplicates=False,
+                sorted=True,
+                prune=True,
+                fill_value=result_fill_value,
+            )
+            out = out.reshape(tuple(self.shape[d] for d in neg_axis))
```
hameerabbasi, line 481: Actually, that applies to all of this as well.

daletovar:
@hameerabbasi, is this more in line with what you were thinking?

sparse/_sparse_array.py
```diff
@@ -340,6 +340,51 @@ def reduce(self, method, axis=(0,), keepdims=False, **kwargs):
...
+            for ax in axis:
+                shape[ax] = 1
+            out = out.reshape(shape)
+
+        if out.ndim == 0:
+            return out[()]
+
+        return out
+
         raise NotImplementedError
```
hameerabbasi: These lines can be removed.

hameerabbasi (commented):
LGTM, @daletovar! One minor change more. Let's add a `_review_return` and `_review_calc` method to `SparseArray` that raises a `NotImplementedError`

sparse/_sparse_array.py
daletovar: Oh it looks like I forgot to delete that.

daletovar:
@hameerabbasi, thanks for everything. What about something like this:
```
try:
    out = self._reduce_calc(method, axis, keepdims, **kwargs)
except:
    raise NotImplementedError
```

hameerabbasi:
More like this:

```python
class SparseArray:
    def _reduce_calc(...):
        raise NotImplementedError
```

This way, the method actually exists on `SparseArray`.

hameerabbasi (approved):
LGTM, thanks, @daletovar!

PR #407, opened 2020-09-09
Fix a number of random bugs.

### LINKED ISSUE #406: Multiply COO by numpy array gives numpy array when fill_value = 0
**Describe the bug**
When multiplying a COO by a numpy array of the same shape, being the fill_value of COO equal to zero, the result is a normal numpy array. However, [the documentation seems to imply](https://sparse.pydata.org/en/stable/operations.html#operations-with-numpy-arrays) that, under those circumstances, the result should be a sparse COO array.

**To Reproduce**

Run the following code:

```python
import numpy as np
import sparse as sp

a = np.random.random((30, 20))
a[a < 0.7] = 0.0
b = sp.COO.from_numpy(a)

c = b * a
print(type(c))
```

This code indicates that the type of c is `numpy.ndarray`.

**Expected behavior**
I would expect `c` to be a COO array, given that:

> - The operation must produce a consistent fill-values. In other words, the resulting array must also be sparse.
> - Operating on the NumPy arrays must not increase the size when broadcasting the arrays.

**System**
 - OS and version: MacOS Catalina 10.15.6
 - `sparse` version (`sparse.__version__`): 0.11.2
 - NumPy version (`np.__version__`):  1.19.0
 - Numba version (`numba.__version__`): n/a

**Additional context**
If the above is actually the designed behavior, then the documentation needs to be ament to clarify what is the output of that operation. Although, to be honest, it looses parts of its usfulness if such operation results in dense array when, in reality, it is not.

hameerabbasi:
Fixes #406

### DIFF
--- a/sparse/tests/test_array_function.py
+++ b/sparse/tests/test_array_function.py
@@ -64,6 +64,7 @@ def test_ternary(func, arg_order):
     xx = func(x, x, x)
     args = [(x, y)[i] for i in arg_order]
     yy = func(*args)
+    sparse.save_npz("y.npz", y)
     assert_eq(xx, yy)
 
 
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -274,20 +274,6 @@ def _dot(a, b, return_type=None):
     from ._compressed.convert import uncompress_dimension
     from ._sparse_array import SparseArray
 
-    def sort_indices(data, indices, indptr, shape):
-        """
-        Several of the dot algorithms produce indices that
-        are out of order. So we have to do a sort of indices
-        and data.
-        """
-        coords = np.empty((2, indices.shape[0]), dtype=np.intp)
-        coords[0, :] = uncompress_dimension(indptr)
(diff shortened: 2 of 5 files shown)


sparse/tests/test_array_function.py
```diff
@@ -64,6 +64,7 @@ def test_ternary(func, arg_order):
     xx = func(x, x, x)
     args = [(x, y)[i] for i in arg_order]
     yy = func(*args)
+    sparse.save_npz("y.npz", y)
```
perimosocordiae, line 67: Was this intended to be included in the PR?

hameerabbasi: Whoops. No. I didn't know you kept an eye on this repo! I was debugging a particular test that failed in some scenarios so I dumped the array to repro it outside the test.

hameerabbasi: Fixed in 62e35a5ac4677e2922c088aadbf27cf89c9050ae.

PR #410, opened 2020-09-22
Add nnz parameter to sparse.random

emilmelnikov:
Currently, `sparse.random` uses the `density` argument to determine
the number of non-zero elements in the generated array. If the exact
number of non-zero elements is required, one might try to compute
density as `nnz / np.prod(shape)`, which might give a wrong result
for large values due to a floating-point error.

This PR adds a new, optional `nnz` argument to `sparse.random` that,
if specified, overrides `density` and allows a caller to create
a random array with the exact number of non-zero elements.

### DIFF
--- a/sparse/_utils.py
+++ b/sparse/_utils.py
@@ -1,3 +1,4 @@
+import contextlib
 import functools
 from collections.abc import Iterable
 from numbers import Integral
@@ -456,3 +457,28 @@ def check_consistent_fill_value(arrays):
                 "is different from a fill_value of {!s} in the first "
                 "argument.".format(i, arg.fill_value, fv)
             )
+
+
+try:
+    nullcontext = contextlib.nullcontext
+except AttributeError:
+    # Shim for Python 3.6.
+    class nullcontext(contextlib.AbstractContextManager):
+        """Context manager that does no additional processing.
+
+        Used as a stand-in for a normal context manager, when a particular
+        block of code is only sometimes used with a normal context manager:
+
+        cm = optional_cm if condition else nullcontext()
+        with cm:
+            # Perform operation, using optional_cm if condition is True
+        """
+
+        def __init__(self, enter_result=None):
+            self.enter_result = enter_result
+
+        def __enter__(self):
+            return self.enter_result
+
+        def __exit__(self, *excinfo):
+            pass
--- a/sparse/_utils.py
+++ b/sparse/_utils.py
@@ -79,6 +79,7 @@ def _zero_of_dtype(dtype):
 def random(
     shape,
     density=0.01,
+    nnz=None,
     random_state=None,
     data_rvs=None,
     format="coo",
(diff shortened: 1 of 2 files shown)


sparse/_utils.py
```diff
@@ -79,6 +79,7 @@ def _zero_of_dtype(dtype):
 def random(
     shape,
     density=0.01,
```
hameerabbasi, line 81:
```suggestion
    density=None,
```

sparse/_utils.py
```diff
@@ -140,9 +144,19 @@ def random(
     # See https://github.com/scipy/scipy/blob/master/LICENSE.txt
     from ._coo import COO
 
+    if not (0 <= density <= 1):
```
hameerabbasi, line 147: Add a check here that ensures that both are not provided.

hameerabbasi (changes requested):
Hello. This is a nice addition, I'd just like to suggest tests for:

* Err on something outside the unit interval.
* Err when both `density` and `nnz` are provided.

Use `pytest.raises` (there are a couple of usages in this file).

sparse/_utils.py
emilmelnikov: Wouldn't this break backwards compatibility?

emilmelnikov: Wouldn't this break backwards compatibility if neither `density` nor `nnz` are provided?

hameerabbasi:
I think there was a misunderstanding about what I meant: If exceptions are raised, they fail the test anyway (which is what we want). I meant you should test that:

1. The exceptions you're raising are actually raised (use `pytest.raises`, grep for it in the codebase).
2. Change the default `density` to `None`.
3. Raise an error if both `nnz` and `density` are not `None`.
4. If both are `None`, assume a density of `0.01`.

sparse/_utils.py
hameerabbasi: See https://github.com/pydata/sparse/pull/410#issuecomment-696813370

hameerabbasi (approved):
One minor change more, otherwise LGTM.

sparse/_utils.py
```diff
@@ -140,9 +145,23 @@ def random(
     # See https://github.com/scipy/scipy/blob/master/LICENSE.txt
     from ._coo import COO
 
+    if density is not None and nnz is not None:
+        raise ValueError("'density' and 'nnz' are mutually exclusive")
+
+    if density is None:
+        density = 0.01
+    if not (0 <= density <= 1):
+        raise ValueError("density {} is not in the unit interval".format(density))
```
hameerabbasi, line 154: Can you test this error as well, because it's new?

emilmelnikov: It is checked by `test_random_invalid_density_and_nnz` with `density=1.01` and `density=-0.01`.

hameerabbasi:
Thanks, @emilmelnikov!

PR #412, opened 2020-09-24
Add elemwise ops for GCXS

daletovar:
- Moves element-wise operations
- For now, just converts GCXS to COO to compute elemwise and then converts again at the end
- If COO and GCXS interact, the output will be COO
- Moves methods that rely on elemwise to _sparse_array: `round, clip, astype, mean, var, std
`
In the future I like the idea of converting all SparseArrays to a linearized representation and do the computations on that and then convert the result to the desired format. It looks like this is basically what's done anyway considering that `_match_arrays` works with linearized coords. I just think it's currently wasteful to convert GCXS -> COO -> linearized -> COO -> GCXS.


hameerabbasi:
> In the future I like the idea of converting all SparseArrays to a linearized representation and do the computations on that and then convert the result to the desired format.

Unfortunately linearized coords don't work for broadcasting cases.

sparse/_compressed/compressed.py
```diff
@@ -141,6 +142,24 @@ def __init__(
...
+            copied. Set to ``False`` to only make a shallow copy.
+        """
+        return _copy.deepcopy(self) if deep else _copy.copy(self)
+
+    def _make_shallow_copy_of(self, other):
+        self.data = other.data
+        self.indices = other.indices
+        self.indptr = other.indptr
+        self.compressed_axes = other.compressed_axes
+        super().__init__(other.shape, fill_value=other.fill_value)
```
hameerabbasi, line 161:
```suggestion
        return self.copy(deep=False)
```

sparse/_sparse_array.py
```diff
@@ -541,3 +538,312 @@ def prod(self, axis=None, keepdims=False, dtype=None, out=None):
...
+        See also
+        --------
+        scipy.sparse.coo_matrix.astype : SciPy sparse equivalent function
+        numpy.ndarray.astype : NumPy equivalent ufunc.
+        :obj:`COO.elemwise`: Apply an arbitrary element-wise function to one or two
+            arguments.
+        """
+        # this matches numpy's behavior
+        if self.dtype == dtype and not copy:
+            return self
```
hameerabbasi, line 593: A test for this branch would be great, checking for object identity.

sparse/tests/test_elemwise.py
```diff
@@ -0,0 +1,707 @@
...
+    class A:
+        pass
+
+    s1 = sparse.random((2, 3, 4), density=0.5)
+    x2 = A()
+
+    with pytest.raises(TypeError):
+        s1 + x2
+
+    assert sparse.elemwise(operator.add, s1, x2) is NotImplemented
```
hameerabbasi, line 120: Thanks for this test!

sparse/tests/test_elemwise.py
```diff
@@ -0,0 +1,707 @@
...
+def test_elemwise_mixed_empty():
+    s1 = sparse.random((2, 0, 4), density=0.5)
+    x2 = np.random.rand(2, 0, 4)
+
+    x1 = s1.todense()
+
+    assert_eq(s1 * x2, x1 * x2)
+
+
+def test_elemwise_unsupported():
```
hameerabbasi, line 110: Can we parametrize this over the format?

sparse/tests/test_elemwise.py
```diff
@@ -0,0 +1,707 @@
...
+    s1 = sparse.random((2, 3, 4), density=0.5)
+    x2 = A()
+
+    with pytest.raises(TypeError):
+        s1 + x2
+
+    assert sparse.elemwise(operator.add, s1, x2) is NotImplemented
+
+
+def test_elemwise_mixed_broadcast():
```
hameerabbasi, line 123: Can we parametrize this over the format?

sparse/tests/test_elemwise.py
```diff
@@ -0,0 +1,707 @@
...
+
+    x = xs.todense()
+    y = ys.todense()
+    z = zs.todense()
+
+    fs = sparse.elemwise(func, xs, ys, zs)
+    if COO in formats:
+        assert isinstance(fs, COO)
+    else:
+        assert isinstance(fs, GCXS)
```
hameerabbasi, line 200: I would remove any guarantees here about mixed formats. If the formats are the same, then, yes, please, test that `format_in == format_out`.

hameerabbasi (changes requested):
Thanks, @daletovar. Overall, this looks pretty awesome. I've suggested a couple of minor changes.

sparse/_compressed/compressed.py
daletovar: I don't quite see how this makes a copy of `other`. I tried this and it makes most of the tests fail.

daletovar:
Thanks so much for the review, @hameerabbasi.

sparse/_compressed/compressed.py
hameerabbasi: This was a misunderstanding on my part.

daletovar: Gotcha, no worries.

hameerabbasi: Could you try `self.__dict__ = other.__dict__.copy()`?

daletovar: Yeah, definitely. Should I change it for COO as well?

PR #417, opened 2020-11-06
Move conj, real, and imag to SparseArray

daletovar:
Since these are the same for COO and GCXS I think it makes sense to move these to SparseArray. conj is needed for GCXS to work in Scipy.sparse.linalg.


hameerabbasi:
Thanks, @daletovar!

PR #418, opened 2020-11-11
use static arrays for indexing

daletovar:
I've replaced the dynamic structures with static ones for GCXS indexing. This improves the speed pretty significantly (twice as fast in a number of cases). Also, even allocating extra space in arrays seems to outperform lists in terms of memory for the most part.

### DIFF
--- a/sparse/_compressed/indexing.py
+++ b/sparse/_compressed/indexing.py
@@ -100,6 +100,12 @@ def getitem(x, key):
 
     starts = x.indptr[:-1][rows]  # find the start and end of each of the rows
     ends = x.indptr[1:][rows]
+
+    # determine how much memory to allocate for indexing
+    total_nnz = (ends - starts).sum() + 1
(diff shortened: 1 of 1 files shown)


hameerabbasi:
Can you post benchmarks as well?

hameerabbasi:
Also, we might want to hold off on merging this, Numba 0.52 introduces a number of improvements for typed lists, so switching to those might be worthwhile instead.

daletovar:
Yeah, I can definitely add them to the ASV suite.

hameerabbasi:
@daletovar I believe Numba 0.52 is out. Can we re-run the benchmarks on this with that version installed, against the old ones as well?

daletovar:
Sure thing. I'll get right on that.

hameerabbasi:
I ran the benchmarks on this a long time ago and this seemed to be obsolete. Closing for now.

PR #434, opened 2021-02-25
Use smaller dtypes in coords when possible

### LINKED ISSUE #249: Allow uint32 dtype for coords
Hey awesome library! I'm really having a blast with it.

One question / comment:

For my purposes, I need the COO array with the largest dtype for `coords` of `uint32`.

When I create an array, the COO constructor seems to convert all coord arrays to `intp` or `int64` in my case.
https://github.com/pydata/sparse/blob/master/sparse/coo/core.py#L234

Any thoughts on skipping that `astype` if coords is already an array?

daletovar:
Closes #249.

Allows different sized dtypes to be used to store `coords`, `indices`, and `indptr` of `COO` and `GCXS` arrays.

### DIFF
--- a/sparse/_compressed/common.py
+++ b/sparse/_compressed/common.py
@@ -36,7 +36,11 @@ def concatenate(arrays, axis=0, compressed_axes=None):
             ptr_list.append(arr.indptr)
             continue
         ptr_list.append(arr.indptr[1:])
+
+    total_nnz = np.sum([arr.nnz for arr in arrays], dtype=np.intp)
+    indptr_dtype = np.min_scalar_type(total_nnz)
     indptr = np.concatenate(ptr_list)
+    indptr = indptr.astype(indptr_dtype)
     indices = np.concatenate([arr.indices for arr in arrays])
     data = np.concatenate([arr.data for arr in arrays])
     ptr_len = arrays[0].indptr.shape[0]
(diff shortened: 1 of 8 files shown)


ivirshup:
Hey, I would like to request that index dtypes are not computed from values of the array. There are a few reasons for this:

## numpy has strange type promotion rules with `uint`s

First, numpy is real weird about type promotion with `uint`s:

```python
>>> type(np.uint64(2) + 1)
numpy.float64
```

This is not intuitive behaviour, and will definitely catch people off guard. 

## If types are computed based on values, a compiled interface won't work

Using `np.min_scalar_type` on a computed value of an array like `nnz` means that you can't infer the dtype of `indptr` at compile time. This behaviour could not be replicated in a compiled interface (e.g. `numba`).

A solution for this is make the sparse arrays parametric on both the dtype of their values and the dtype of their indices. This would allow compile time inference of resulting index dtypes. As a point of reference, [this is how sparse arrays are typed in Julia](https://docs.julialang.org/en/v1/stdlib/SparseArrays/#man-csc).

## I would like control over the dtypes

Similar to what was requested in #249, I would like to be able to control what `dtypes` are in `indices` and `indptr`. One of the frustrating parts of `scipy.sparse` is that it decides what dtypes are appropriate for the `indices` based on it's values. If I know that I'm going to be computing indices of a much larger sparse matrix, it's pretty annoying when the `indices` are automatically converted to a smaller `dtype` by the library. 

--------------------

I think a reasonable solution here is to do what `julia` does. Index dtypes default to system width integers:

```julia
julia> using SparseArrays
julia> s_int_inds = sprand(10, 10, .05)
10×10 SparseMatrixCSC{Float64,Int64} with 4 stored entries:
  [2,  2]  =  0.574298
  [9,  3]  =  0.189517
  [5,  8]  =  0.805498
  [5, 10]  =  0.225388
```

I can specify the type of the indices:

```julia
julia> s_uint8_inds = SparseMatrixCSC{Float64, UInt8}(sprand(10, 10, .05))
10×10 SparseMatrixCSC{Float64,UInt8} with 3 stored entries:
  [7,  1]  =  0.481469
  [7,  2]  =  0.0770323
  [5, 10]  =  0.968378
```

Type promotion acts on indices

```julia
julia> typeof(hcat(s_uint8_inds, s_int_inds))
SparseMatrixCSC{Float64,Int64}
```

And I'm responsible for overflows:

```julia
julia> repeat(s_uint8_inds, 100)
ERROR: ArgumentError: number of rows (m = 1000) does not fit in Ti = UInt8
```

daletovar:
Thanks for looking at this, @ivirshup. I'll try that out on another branch and see how that looks.

daletovar:
I just added #441. It incorporates some of the feedback given here by @ivirshup . I think I'll close this PR but keep the branch going for now.

PR #435, opened 2021-03-03
Add fancy indexing support for DOK item assignment

DragaDoncila:
Modify `__setitem__` to support a subset of advanced/fancy indexing where all dimensions are indexed. This uses @stschroe's gist [here](https://gist.github.com/stschroe/7c4660774835258c06b2be968eb49204#file-sparse_dok_indexer-py).

We also need `__getitem__`, but we thought we'd submit that as a separate PR.

### DIFF
--- a/sparse/_dok.py
+++ b/sparse/_dok.py
@@ -309,39 +309,38 @@ class DOK(SparseArray):
             return self.fill_value
 
     def __setitem__(self, key, value):
+        value = np.asanyarray(value, dtype=self.dtype)
+
+        # 1D fancy indexing
         if (
-            isinstance(key, tuple)
-            and len(key) == self.ndim
-            and all(isinstance(k, Iterable) for k in key)
+            self.ndim == 1
+            and isinstance(key, Iterable)
+            and all(isinstance(i, (int, np.integer)) for i in key)
         ):
-            if all(len(key[0]) == len(k) for k in key[1:]):
-                self._fancy_setitem(key, value)
-                return
+            key = (key,)
+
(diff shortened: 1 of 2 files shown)


hameerabbasi:
It would be nice to get the code coverage up -- Look into `pytest.raises` (there are uses in the tests) and try to hit everything and possibly remove dead code.

DragaDoncila:
Hi @hameerabbasi I've added a few more tests and it looks like it's hitting 100% of the diff. Please let me know if there's anything else you'd like to see!

sparse/_dok.py
```diff
@@ -308,15 +309,54 @@ def __getitem__(self, key):
             return self.fill_value
 
     def __setitem__(self, key, value):
-        key = normalize_index(key, self.shape)
-        value = np.asanyarray(value)
+        value = np.asanyarray(value, dtype=self.dtype)
```
hameerabbasi, line 312:
```suggestion
        value = np.asarray(value, dtype=self.dtype)
```
`np.asanyarray` lets subclasses through. There are prominent subclasses of `np.ndarray` that don't follow the [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle), and are therefore not fit for consumption by this function.

hameerabbasi (changes requested):

hameerabbasi:
This should be the last change needed -- Sorry for the slow feedback loop.

DragaDoncila:
Not slow at all! Thank you for the prompt reviews :smiley:

hameerabbasi (approved):

hameerabbasi:
Thanks @DragaDoncila! This is in!

PR #436, opened 2021-03-04
Add slicing and subset of fancy indexing to DOK array access

DragaDoncila:
Extend DOK `__getitem__` to support slicing and fancy indexing when all dimensions are indexed.


hameerabbasi:
This is in, thanks @DragaDoncila!

PR #437, opened 2021-03-04
Document ability to slice and fancy index DOK arrays

DragaDoncila:
Document new functionality for DOK array

Should be merged after #436.


hameerabbasi:
Thanks once more, @DragaDoncila!

PR #439, opened 2021-03-09
Fix dok getitem

DragaDoncila:
@hameerabbasi  there was a bug in the new `getitem` code when slicing resulted only in fill values. Have fixed that and added a test for it.


hameerabbasi (approved):
Thanks for all your work on this, @DragaDoncila!

DragaDoncila:
@hameerabbasi it's a pleasure - sorry for not catching this one before!

PR #440, opened 2021-03-10
Implement equality comparison for DOK arrays

DragaDoncila:
Behaves as closely as possible to `np.ndarray` comparisons
- Single values are broadcast
- Broadcasting for arrays of different shapes is as per `numpy` broadcasting rules
- Always returns a new DOK array of bools

@hameerabbasi I think this is the last operation we need for our use-case over at napari. Please let me know if there's any changes you'd like made!

### DIFF
--- a/sparse/_dok.py
+++ b/sparse/_dok.py
@@ -3,13 +3,14 @@ from numbers import Integral
 from collections.abc import Iterable
 
 import numpy as np
+from numpy.lib.mixins import NDArrayOperatorsMixin
 
 from ._slicing import normalize_index
 from ._utils import equivalent
 from ._sparse_array import SparseArray
 
 
-class DOK(SparseArray):
+class DOK(SparseArray, NDArrayOperatorsMixin):
     """
     A class for building sparse multidimensional arrays.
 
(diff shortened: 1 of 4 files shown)


hameerabbasi:
In `COO` (and through it) `GCXS`, we have logic that unifies all the element-wise broadcasting n-ary operators. Would it be difficult to do something like that here? I'd hate to special case `==` here.

DragaDoncila:
Hi @hameerabbasi , more than happy to have a look at unifying these two implementations, but we're (@jni and I) just a bit confused how the dispatch works for the `__array_ufunc__`.

Given a `COO` array and a `DOK` array, we see that the `__array_ufunc__` is called and proceeds through its logic. However, if a `DOK` and `DOK` are compared (or a `DOK` against an int), the `__array_ufunc__` is never called, and we get a `TypeError`.

If both of the array types inherit from `SparseArray`, how come the `__array_ufunc__` is not called at all when the two operands are both `DOK` arrays, or when one is a `DOK` and the other an int? 

Additionally, what is the machinery that allows `__array_ufunc__` to be called when the operands include a `COO` array?

hameerabbasi:
You must define an `__array_ufunc__` implementation on DOK as well, as one isn't defined right now at all. One that accepts only DOK and nothing else (returns `NotImplemented` if any other type is passed in).

Here is a NEP with the details: https://numpy.org/neps/nep-0013-ufunc-overrides.html

jni:
@hameerabbasi what we're confused about is that COO does not implement its own `__array_ufunc__` method, it is only implemented in the base class. And how does *Python* ever get to dispatching to `__array_ufunc__` without a `__eq__` or `__lt__` pointing to the ufunc machinery, in the absence of any NumPy array?

hameerabbasi:
Oh, yes. The `__eq__` and so on methods are defined for COO, and that calls `np.equals`. Those are inherited from `np.lib.

jni:
@hameerabbasi ah! NDArrayOperatorsMixin! 🤦 I hate OOP. 😅 We're off to the races! 🙏

DragaDoncila:
@hameerabbasi we've unified the elementwise operations for DOK in the `SparseArray __array_ufunc__`. Let us know if this seems reasonable to you!

hameerabbasi (approved):
Thanks for all the awesome work, @DragaDoncila @jni!

PR #441, opened 2021-03-11
allow for use of different dtypes in coords, indices, and indptr

daletovar:
In this version, I've added a `storage_dtype` argument in the `COO` and `GCXS` constructors. For most operations, the result array will use the same storage dtype as the input array. There are a few places where the output is upcast to a larger dtype if the input dtype is too small. I decided to do this because in these situations no error is raised and the output is simply wrong. Where it doesn't make sense to upcast, I have it raise an error.

### DIFF
--- a/sparse/_compressed/common.py
+++ b/sparse/_compressed/common.py
@@ -41,7 +41,7 @@ def concatenate(arrays, axis=0, compressed_axes=None):
     data = np.concatenate([arr.data for arr in arrays])
     ptr_len = arrays[0].indptr.shape[0]
     nnz = arrays[0].nnz
-    total_nnz = sum([arr.nnz for arr in arrays[:-1]])
+    total_nnz = sum(int(arr.nnz) for arr in arrays)
     if not can_store(indptr.dtype, total_nnz):
         indptr = indptr.astype(np.min_scalar_type(total_nnz))
     for i in range(1, len(arrays)):
@@ -96,7 +96,7 @@ def stack(arrays, axis=0, compressed_axes=None):
     data = np.concatenate([arr.data for arr in arrays])
     ptr_len = arrays[0].indptr.shape[0]
     nnz = arrays[0].nnz
-    total_nnz = sum([arr.nnz for arr in arrays[:-1]])
+    total_nnz = sum(int(arr.nnz) for arr in arrays)
     if not can_store(indptr.dtype, total_nnz):
         indptr = indptr.astype(np.min_scalar_type(total_nnz))
     for i in range(1, len(arrays)):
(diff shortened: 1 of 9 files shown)


sparse/_compressed/common.py
```diff
@@ -41,6 +41,9 @@ def concatenate(arrays, axis=0, compressed_axes=None):
     data = np.concatenate([arr.data for arr in arrays])
     ptr_len = arrays[0].indptr.shape[0]
     nnz = arrays[0].nnz
+    total_nnz = sum([arr.nnz for arr in arrays[:-1]])
```
hameerabbasi, line 44:
```suggestion
    total_nnz = sum(int(arr.nnz) for arr in arrays[:-1])
```

To make sure we avoid overflows, and don't construct a list.

hameerabbasi (changes requested):
A general comment: We should strive for 100% test coverage. If we cannot hit a line with a test, it is probably dead code anyway. In this particular instance, I see that most of the code here can be hit; but I'd like to see it tested since this is a major feature/refactor.

sparse/_compressed/common.py
```diff
@@ -93,6 +96,9 @@ def stack(arrays, axis=0, compressed_axes=None):
     data = np.concatenate([arr.data for arr in arrays])
     ptr_len = arrays[0].indptr.shape[0]
     nnz = arrays[0].nnz
+    total_nnz = sum(int(arr.nnz) for arr in arrays)
+    if not can_store(indptr.dtype, total_nnz):
+        indptr = indptr.astype(np.min_scalar_type(total_nnz))
```
hameerabbasi, line 101: This branch needs hitting or removing.

sparse/_compressed/compressed.py
```diff
@@ -49,16 +50,28 @@ def _from_coo(x, compressed_axes=None):
     compressed_shape = (row_size, col_size)
     shape = x.shape
 
+    if storage_dtype and not can_store(storage_dtype, max(compressed_shape)):
+        raise ValueError(
```
hameerabbasi, line 54: So does this one.

sparse/_utils.py
```diff
@@ -190,7 +191,22 @@ def random(
...
+        data,
+        shape=elements,
+        fill_value=fill_value,
+    ).reshape(shape)
+
+    if storage_dtype:
+        if can_store(storage_dtype, max(shape)):
+            ar.coords = ar.coords.astype(storage_dtype)
+        else:
+            raise ValueError(
```
hameerabbasi, line 205: As does this one.

hameerabbasi (changes requested):
Just a few more remaining! Great work!

ivirshup:
Thanks for looking into this more @daletovar! Minor naming suggestion: something like `index_dtype`/ `idx_dtype`? I feel like `storage_dtype` maybe be a bit too generic of a name.

hameerabbasi (approved):
LGTM, thanks @daletovar!

Issue #433, opened 2021-02-22
CSC/ CSR classes


ivirshup:
### Is your feature request related to a problem? Please describe.

Could explicit `CSR` and `CSC` classes be added to this library?

I would love to stop using `scipy.sparse.spmatrix` classes. Their `matrix` API is a source of constant frustration. Lack of CSC and CSR arrays with an np.ndarray API has been a barrier to adoption not only of this library, but of `xarray` and `dask` for the `scanpy`/ `anndata` projects 

### Describe the solution you'd like

I would like there to be explicit `CSR` and `CSC` classes in this package.  These should perform at least as well as `scipy.sparse.spmatrix` counterparts.

While higher dimensional algorithms are developed, performant implementations for these cases can be taken from `scipy.sparse` or developed from reference implementations. I think this is a low effort/ high pay-off path towards performant implementations for common formats. They wouldn't be n-dimensional, but there's a lot of value in being able to start consolidating efforts to this library before an n-dimensional generalization has the features and performance of the common 2d cases.

#### No explicit CSC/ CSR classes. These are subsets of more generic structures so why special case them?

I think this is the wrong approach due to how common these structures are. Even a flexible framework like `taco` exports these formats (`pytaco.csc` and `pytaco.csr`).

We could also consider what code looks like if these classes are not defined. All dispatch here could change from `isinstance(x, sparse.csr_matrix)` to something like `isinstance(x, sparse.GXCS) and x.compressed_axes == (0,) and x.ndims == 2)`, but something closer to the first case is probably an easier sell.

#### CSC/ CSR classes as thin wrappers of an n-dimensional format

I think this is a great goal. However, it requires an implementation of that format which is at least as performant in most (probably all) cases as the `scipy` equivalents for broad adoption. As mentioned above, performant implementations can be (1) sourced from a current dependency (`scipy.sparse`) or (2) easily adapted from existing reference implementations.

Code specialized for these cases could gradually be replaced with more generic versions as they are implemented. However, performant implementation for these common cases would not be blocked by implementation of n-dimensional algorithms.

#### I could expand on why I think this is the right way to go

I could go on about how important I think it is for there to be CSR and CSC classes in this library, and why it's important to implement these separately from higher dimensional generalizations. I find that makes this post too long. I'd be happy to add these arguments if asked.

#### Type hierarchy

[...]

PR #442, opened 2021-03-15
Initial CSC/ CSR classes

ivirshup:
As discussed in #433. The goal of this PR is primarily to get skeletons for these classes that work, while fast implementations of indices and broadcasting can be added in subsequent PRs.

### Class structure

Current design: `CSC` and `CSR` classes subclass `GCXS`. Is this a good idea, or should a non-user visible super class be added? It may be strange that:

```python
gcxs = sparse.random((50, 30), .2, format="gcxs")
csc = CSC(gcxs)

assert isinstance(csc, GCXS) and isinstance(gcxs, GCXS)
assert not np.array_equal(csc.data, gcxs.data)
```

### Return types

What is the result type of `CSR + CSC`? What about `COO + CSC`?

Do we follow `scipy` convention here?


sparse/_compressed/compressed.py
```diff
@@ -788,3 +799,87 @@ def _prune(self):
...
+        self.data, self.indices, self.indptr = arg
+
+        if self.data.ndim != 1:
+            raise ValueError("data must be a scalar or 1-dimensional.")
+
+        self.shape = shape
+        self.fill_value = fill_value
+
+        if prune:
+            self._prune()
```
hameerabbasi, line 848: Can we somehow re-use `super().__init__` here?

sparse/_compressed/compressed.py
```diff
@@ -382,6 +383,12 @@ def tocoo(self):
             .transpose(order)
         )
 
+    def tocsr(self):
+        return CSR(self)
+
+    def tocsc(self):
+        return CSC(self)
+
```
hameerabbasi, line 391: I'd prefer for these to not exist, we could direct users to `asformat` instead.

sparse/_compressed/compressed.py
```diff
@@ -1,4 +1,5 @@
 import copy as _copy
+from numba.core.errors import CompilerError
```
hameerabbasi, line 2: Do we need this anywhere?

hameerabbasi (changes requested):
I've given this a shallow review, deeper one to come later.

sparse/_compressed/compressed.py
ivirshup: A couple things stopped this for me:

* `self.compressed_axes = compressed_axes` in `GCXS.__init__`
* Being able to error once I knew the shape was wrong, instead of waiting until it's constructed

sparse/_compressed/compressed.py
ivirshup: :+1: I wasn't sure how many of the construction methods to support.

sparse/_compressed/compressed.py
ivirshup: No, probably just overactive vscode autocomplete on my end.

sparse/_compressed/compressed.py
hameerabbasi: Can we put the shape-checking logic above `super().__init__`?

hameerabbasi: As concerns `compressed_axes`, that's constant anyway for these classes.

hameerabbasi: Concretely, we can use it as:

```python
# Shape-checking goes here
super().__init__(..., compressed_axes=(0,)) # Or (1,) for CSC.
```

ivirshup: `compressed_axes` is mutable for `GCXS`, but it really shouldn't be for `CSR` and `CSC`. Making a setter for `compressed_axes` that checks the value isn't different feels wrong to me. `compressed_axes` cannot be made immutable for `GCXS` because it's mutated by methods like `resize`.

I think I can do this with a little bit of introspection, but I'm not sure it's a good idea overall. For example, I thinks it's fairly simple to have a faster method for creating 2d compressed sparse arrays from 2d dense arrays, but am not sure this is the case for nd arrays.

There's specialization that can happen for the 2d case in the constructor, but sharing that code will add indirection for the nd case. This is a bit of a broader implementation question. For code that works with nd arrays, do they get a check at the front that says: `if 2d: fast_path_2d(...)` or does the 2d case just call the fast path directly?

ivirshup: I've made the change in 01caa5c9781134fc4ec3e, but let me know what you think

ivirshup:
Could `CSR(arg: scipy.sparse.spmatrix)` work for conversion? I feel like it's unintuitive that `CSR(: Union[ndarray, SparseArray])` works, but construction from `spmatrix` has to be through a separate method.

sparse/_compressed/compressed.py
hameerabbasi: > I've made the change in [01caa5c](https://github.com/pydata/sparse/commit/01caa5c9781134fc4ec3e745162372290e2dfd94), but let me know what you think

We should make `compressed_axes` immutable, that's the solution I see. All these attributes should be immutable.

hameerabbasi:
> Does this mean modifying all of:

Yes.

> I'm wondering if these can be split up into separate PRs, since they touching different parts of the codebase.

Yes, I'm fine with that.

> Right now I'm thinking that dimension preserving operations with `CSR`/ `CSC` should generally prefer to return these types.

I'd say they _should_ return these types, but I'd prefer not to be _locked into_ that decision. Which is why we return the correct type, but only document we're returning a `SparseArray`.

> Could `CSR(arg: scipy.sparse.spmatrix)` work for conversion? I feel like it's unintuitive that `CSR(: Union[ndarray, SparseArray])` works, but construction from `spmatrix` has to be through a separate method.

That should be done, yes, at the `GCXS` level (it will then trickle down to `CSR`/`CSC`). `COO` already does do it. Pretty sure `DOK` doesn't.

sparse/_compressed/compressed.py
ivirshup: What about `array.resize`? That's part of the "array interface" and can modify the shape and compressed axes inplace.

I also expect @daletovar may have some insight here.

Right now, `compressed_axes` is immutable for `CSR` and `CSC`, so any changes to the mutability of `GCXS` attributes is orthogonal. Could this be handled separately from this PR?

hameerabbasi: > Right now, `compressed_axes` is immutable for `CSR` and `CSC`, so any changes to the mutability of `GCXS` attributes is orthogonal. Could this be handled separately from this PR?

Yes, I'm okay splitting it out. But we should create a tracking issue for all this. I'll do that.

> What about `array.resize`? That's part of the "array interface" and can modify the shape and compressed axes inplace.

I personally feel it's okay if we don't implement `resize` at all, I'm fine with `reshape` only.

sparse/_compressed/compressed.py
```diff
@@ -137,9 +151,13 @@ def __init__(
...
-        self.compressed_axes = (
-            tuple(compressed_axes) if isinstance(compressed_axes, Iterable) else None
-        )
+
+        if not isinstance(self, Compressed2d):
+            self.compressed_axes = (
+                tuple(compressed_axes)
+                if isinstance(compressed_axes, Iterable)
+                else None
+            )
```
hameerabbasi, line 160: Why is this change necessary?

hameerabbasi: In particular, I don't like it because it makes the superclass depend on the subclass -- That's bad coding practice, usually.

hameerabbasi (commented):
A general question

sparse/_compressed/compressed.py
```diff
@@ -115,6 +116,9 @@ def __init__(
         self, arg, shape=None, compressed_axes=None, prune=False, fill_value=0
     ):
 
+        if isinstance(arg, ss.spmatrix):
+            arg = self.from_scipy_sparse(arg)
+
```
hameerabbasi, line 121: Can we make `scipy` a soft requirement here?

sparse/_compressed/compressed.py
ivirshup: Otherwise `self.compressed_axes` would be mutable for `Compressed2d` classes, since we'd have assignment here.

I'm open to other suggestions. Alternatives I've considered are:

* Separate `__init__` methods – my initial solution
* `setattr(csr. "compressed_axes", x)` is a no-op – but it really should be an error
* Check that the new value is correct – this is adding moving parts to an immutable property. I don't like making changes to the subclasses based purely on the implementation of super classes.
* Higher level super class. This might be adding too much indirection.

> I don't like it because it makes the superclass depend on the subclass

I agree, but think that super classes have to be designed around allowing subclassing. I think this is a perennial python problem (e.g. how easy is it to subclass an ndarray or dataframe?).

hameerabbasi: One way to go about this is to make it immutable in this PR. Just rename it to `_compressed_axes` and also add a `@property`.

hameerabbasi: Another is to just have it mutable for now, and make it immutable later.

sparse/_compressed/compressed.py
ivirshup: Two things:

* Do you have a preferred way to make dummy imports in this package?
* I think it'd be really nice to be able to lean on `scipy` to get these classes feature complete. Is there a rush to remove it as a dependency before then?

ivirshup:
> CSR(arg: scipy.sparse.spmatrix)

Added in 8f21503

> I'd say they should return these types, but I'd prefer not to be locked into that decision

:+1:. Where would this have been documented? Is it sufficient that `elemwise` returns a `SparseArray`?

sparse/_compressed/compressed.py
hameerabbasi: > Two things:
> 
>     * Do you have a preferred way to make dummy imports in this package?
> 
>     * I think it'd be really nice to be able to lean on `scipy` to get these classes feature complete. Is there a rush to remove it as a dependency before then?

Not really, but I was avoiding making it harder when we do do it, there are very few hard links to it now. But I agree, the path of least resistance can be taken here.

hameerabbasi:
> Where would this have been documented? Is it sufficient that `elemwise` returns a `SparseArray`?

We don't have to do it in this PR, but the docstring of `CSR`/`CSC` classes as well as the getting started page are two possible places to put it.

sparse/_compressed/compressed.py
ivirshup: I don't like the idea of modifying GCXS to support CSR, CSC too much. Especially since I fully expect CSR, CSC code to be volatile for a bit. I also really like keeping CSR, CSC implementation as simple as possible (e.g. `compressed_axes` is constant).

How about I split the shared code into an outside function? It's a bit of indirection, but I don't think we can have a single method that works for one class.

I've also been thinking about it more, and I'm not sure `issubtype(Compressed2d, GCXS)` is always going to be a good idea. For instance, my impression is that `GCXS`, `CSF` and how `taco`'s representations for these arrays are not equivalent though can be equivalent `CSR` and `CSC`. If either of the other implementations happens, does it still make sense for `CSR`/ `CSC` to subtype `GCXS`?

hameerabbasi: We can use composition instead of inheritance and only subclass, `SparseArray`.

hameerabbasi: To answer your question: I don't see anyone relying on subclass behaviour too much, just the fact that it has all the operations.

ivirshup: > just the fact that it has all the operations.

Yeah, this is an important part. I would like to think more about the type structure before making any big decisions on it, and would like to get this merged so operations can be defined and test/ benchmark suites can be built. I imagine experience while adding to the classes – as well as getting more eyes on it – will help clarify here.

To that end, what's the minimum fix here? To me it's:

1. Leave as is
2. Split common initialization into it's own function

I like **1**, since it's the smallest amount of code that allows both behaviors. It also has very little indirection. I'd be happy to implement **2** if that's a better solution to you. Either of these would come with an issue to think about type structure.

> I don't see anyone relying on subclass behaviour too much

I think it will come up, especially as typing continues to mature. I bet it's going to come up while writing the dispatch methods.

hameerabbasi: I think the best fix is to leave it mutable (https://github.com/pydata/sparse/pull/442#discussion_r595011333) and move mutability to the next PR. I'll mark the issue as a release blocker if immutability is that important to you. `GCXS` wasn't released, and so we're fine changing that behavior.

As a side note we may be overthinking the immutability, `COO` is most often used, and it holds even `data` and `coords` as attributes. Hardly anyone I've come across modifies them.

ivirshup: I've added a commit where it's kept as immutable, but we don't need the instance check. Just makes sure it's not set to anything other that what it should be.

hameerabbasi (approved):

hameerabbasi:
This is in, thanks @ivirshup!

mrocklin:
Quick comment from the gallery: this is very fun to see :)

PR #447, opened 2021-03-19
Fix ZeroDivisionError on html_table.

### LINKED ISSUE #424: ZeroDivisionError in jupyter repr
**Describe the bug**
In jupyter, with a size=0 sparse array, I got an error in the repr

```python-traceback

ZeroDivisionError                         Traceback (most recent call last)
~/conda/envs/dask-upstream-dev/lib/python3.8/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/conda/envs/dask-upstream-dev/lib/python3.8/site-packages/sparse/_sparse_array.py in _repr_html_(self)
    170         Renders in Jupyter.
    171         """
--> 172         return html_table(self)
    173 
    174     @abstractmethod

~/conda/envs/dask-upstream-dev/lib/python3.8/site-packages/sparse/_utils.py in html_table(arr)
    319         str(arr.shape),
    320         str(arr.nnz),
--> 321         str(arr.nnz / arr.size),
    322     ]
    323 

ZeroDivisionError: division by zero
```

**To Reproduce**
```python
import numpy as np
import sparse

sparse.COO.from_numpy(np.array(()))
```

**System**
 - `sparse` version (`sparse.__version__`) = `0.11.2+12.g440809f`

**Additional context**
This came up when testing dask since the `_meta` is an empty array.

H4R5H1T-007:
Fix #424

### DIFF
--- a/sparse/_utils.py
+++ b/sparse/_utils.py
@@ -327,12 +327,9 @@ def html_table(arr):
     table = "<table>"
     table += "<tbody>"
     headings = ["Format", "Data Type", "Shape", "nnz", "Density", "Read-only"]
-    
-    if arr.size > 0:
-        density = (arr.nnz / arr.size)
-    else:
-        density = np.NaN
-    
+
+    density = np.float_(arr.nnz) / np.float_(arr.size)
+
     info = [
         type(arr).__name__.lower(),
         str(arr.dtype),
@@ -343,20 +340,18 @@ def html_table(arr):
 
     # read-only
     info.append(str(not hasattr(arr, "__setitem__")))
-    
-    if any(x == 0 for x in arr.shape):
-        storage_ratio = np.NaN
-    else:
-        storage_ratio = float("{:.1f}".format(arr.nbytes / (reduce(operator.mul, arr.shape, 1) * arr.dtype.itemsize)))
-    
+
     if hasattr(arr, "nbytes"):
         headings.append("Size")
         info.append(human_readable_size(arr.nbytes))
         headings.append("Storage ratio")
         info.append(
-            str(storage_ratio)
+            "%.1f"
+            % (
+                np.float_(arr.nbytes)
+                / np.float_(reduce(operator.mul, arr.shape, 1) * arr.dtype.itemsize)
+            )
         )
-
 
     # compressed_axes
     if type(arr).__name__ == "GCXS":
--- a/sparse/_utils.py
+++ b/sparse/_utils.py
@@ -327,25 +327,36 @@ def html_table(arr):
     table = "<table>"
     table += "<tbody>"
     headings = ["Format", "Data Type", "Shape", "nnz", "Density", "Read-only"]
+    try:
+        density = (arr.nnz / arr.size)
+    except ZeroDivisionError:
+        density = 'undefined'
+    
     info = [
         type(arr).__name__.lower(),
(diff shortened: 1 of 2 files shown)


sparse/_utils.py
```diff
@@ -327,26 +327,37 @@ def html_table(arr):
     table = "<table>"
     table += "<tbody>"
     headings = ["Format", "Data Type", "Shape", "nnz", "Density", "Read-only"]
+    try:
+        density = (arr.nnz / arr.size)
+    except ZeroDivisionError:
+        density = 'undefined'
```
hameerabbasi, line 333:
```suggestion
        density = np.Nan
```

sparse/_utils.py
```diff
@@ -327,26 +327,37 @@ def html_table(arr):
...
+        str(density),
     ]
 
     # read-only
     info.append(str(not hasattr(arr, "__setitem__")))
 
+    try:
+        storage_ratio = "{:.1f}".format(arr.nbytes / (reduce(operator.mul, arr.shape, 1) * arr.dtype.itemsize))
+    except ZeroDivisionError:
+        storage_ratio = 'undefined'
```
hameerabbasi, line 349:
```suggestion
        storage_ratio = np.man
```

hameerabbasi (changes requested):
Maybe cast these things to `float` so things happen automatically, without a try-catch?

sparse/tests/test_coo.py
```diff
@@ -1606,3 +1606,10 @@ def test_coo_valerr():
 def test_random_idx_dtype():
     with pytest.raises(ValueError):
         sparse.random((300,), density=0.1, format="coo", idx_dtype=np.int8)
+
+
+def test_html_for_size_zero():
+    try:
+        html_table(sparse.COO.from_numpy(np.array(())))
```
hameerabbasi, line 1613: The try catch is unnecessary, the test fails on an exception.

Also, we should check the result of density and storage ratio, that's where the original bug was.

hameerabbasi (changes requested):
The test needs fixing.

H4R5H1T-007:
Tests are failing in Windows environment @hameerabbasi sir. can you check what is error because I am unable to understand it.

sparse/tests/test_coo.py
```diff
@@ -1606,3 +1607,43 @@ def test_coo_valerr():
 def test_random_idx_dtype():
     with pytest.raises(ValueError):
         sparse.random((300,), density=0.1, format="coo", idx_dtype=np.int8)
+
+
+def test_html_for_size_zero():
+    arr = sparse.COO.from_numpy(np.array(()))
+    ground_truth = "<table>"
```
hameerabbasi, line 1614: It'd be ideal if ground truth was a literal string, and you didn't have to construct it here.

Copy-pasting code from the function into the test is usually Bad.

hameerabbasi (changes requested):

hameerabbasi:
Thanks for your work on this, @H4R5H1T-007!

H4R5H1T-007:
Thank You @hameerabbasi sir for your guidance to solve this issue. It helped me to learn a lot.

PR #448, opened 2021-03-21
Modify GCXS docstring

sayandip18:
The `GCXS` docstring now has some information on the structure and advantages of GCRS/GCCS scheme.

### DIFF
--- a/sparse/_compressed/compressed.py
+++ b/sparse/_compressed/compressed.py
@@ -97,7 +97,7 @@ class GCXS(SparseArray, NDArrayOperatorsMixin):
     significantly reducing storage.
 
     Let the 3 arrays be RO, CO and VL. The first element
-    of array RO is the integer 1 and later elements are the number of
+    of array RO is the integer 0 and later elements are the number of
     cumulative non-zero elements in each row for GCRS, column for
     GCCS. CO stores column indexes of non-zero elements at each row for GCRS, column for GCCS.
     VL stores the values of the non-zero array elements.
--- a/sparse/_compressed/compressed.py
+++ b/sparse/_compressed/compressed.py
@@ -89,11 +89,21 @@ class GCXS(SparseArray, NDArrayOperatorsMixin):
 
     This is stored in GCXS format, a generalization of the GCRS/GCCS formats
     from 'Efficient storage scheme for n-dimensional sparse array: GCRS/GCCS':
-    https://ieeexplore.ieee.org/document/7237032. GCXS generalizes the csr/csc
-    sparse matrix formats. For arrays with ndim == 2, GCXS is the same csr/csc.
+    https://ieeexplore.ieee.org/document/7237032. GCXS generalizes the CRS/CCS
+    sparse matrix formats.
+
+    For arrays with ndim == 2, GCXS is the same CSR/CSC.
     For arrays with ndim >2, any combination of axes can be compressed,
     significantly reducing storage.
 
+    Let the 3 arrays be RO, CO and VL. The first element
+    of array RO is the integer 1 and later elements are the number of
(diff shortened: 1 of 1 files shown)


hameerabbasi:
Black is failing -- Run `black .` in the source to fix it.

Also we don't need an overview of the research, we're trying to keep things simple for the user. Ideally, we should just tell them what arrays are stored and what they mean, and how it relates to CCS/CRS.

sayandip18:
> and how it relates to CCS/CRS.

Then we should add some docstring to the recently implemented CCS/CRS classes too. Do you agree?

hameerabbasi:
> Then we should add some docstring to the recently implemented CCS/CRS classes too so that we don't have to explain the structure of CCS/CRS again in GCXS docstring. Do you agree?

If you're willing to do that, it's a great idea. 😄

sayandip18:
Okay, I'm on it!

sparse/_compressed/compressed.py
```diff
@@ -90,24 +90,18 @@ class GCXS(SparseArray, NDArrayOperatorsMixin):
...
 
     For arrays with ndim == 2, GCXS is the same CSR/CSC.
     For arrays with ndim >2, any combination of axes can be compressed,
     significantly reducing storage.
 
+    Let the 3 arrays be RO, CO and VL. First element
+    of array RO is 1 and later elements are the number of
+    cumulative non-zero elements in each row for GCRS, column for
+    GCCS. CO stores column indexes of non-zero elements at each row for GCRS, column for GCCS.
+    VL stores the values of the non-zero array elements.
```
hameerabbasi, line 103: This should be in the `CSR` docstring, then you can say `CSC transpose = CSR` and `GCXS transpose + reshape = CSR`.

sayandip18: > This should be in the `CSR` docstring
CSR/CCS has another set of arrays KOi which stores the indices of the values. Whereas GCXS has no such array. I deliberately wrote that.

hameerabbasi: It still generalizes down. ;) If you ignore the label of the array.

sayandip18: Instead of simply writing CCS is the transpose of CSR, I intended to write a brief description of each kind in their respective docstrings.

hameerabbasi: SGTM, as long as you mention that fact.

sayandip18: > It still generalizes down. ;) If you ignore the label of the array.

Agreed. I'll make the changes as you mentioned.

sparse/_compressed/compressed.py
```diff
@@ -89,11 +89,21 @@ class GCXS(SparseArray, NDArrayOperatorsMixin):
...
-    sparse matrix formats. For arrays with ndim == 2, GCXS is the same csr/csc.
+    https://ieeexplore.ieee.org/document/7237032. GCXS generalizes the CRS/CCS
+    sparse matrix formats.
+
+    For arrays with ndim == 2, GCXS is the same CSR/CSC.
     For arrays with ndim >2, any combination of axes can be compressed,
     significantly reducing storage.
 
+    Let the 3 arrays be RO, CO and VL. The first element
+    of array RO is the integer 1 and later elements are the number of
```
hameerabbasi, line 100:
```suggestion
    of array RO is the integer 0 and later elements are the number of
```

We use 0-based indexing in Python.

hameerabbasi (commented):
Last comment, thanks for the changes @sayandip18!

sparse/_compressed/compressed.py
sayandip18: Ah, thanks!

hameerabbasi (approved):

PR #450, opened 2021-03-22
Docstrings to CSC, CSR classes

sayandip18:


sparse/_compressed/compressed.py
```diff
@@ -845,6 +845,12 @@ def ndim(self) -> int:
 
 
 class CSR(_Compressed2d):
+    """
+    The CSR or CRS scheme stores a n-dimensional array using n+1 one-dimensional arrays.
```
hameerabbasi, line 849: This isn't strictly true, we support only 2D CSR/CSC.

hameerabbasi (approved):
Thanks, @sayandip18!

PR #451, opened 2021-03-22
Document elementwise functions for DOK arrays

DragaDoncila:
@hameerabbasi I've updated the docs to include elementwise operations for DOK arrays. 

At this point I'm not sure whether you want to keep the DOK documentation in Construct Sparse Arrays, or whether you'd like to pull it out as its own section, but please let me know if that's the case, and whether you want more detail on the elementwise operations e.g. more detail on comparisons, etc. As it is, I tried to follow the convention already used for the COO arrays.

### DIFF
--- a/docs/construct.rst
+++ b/docs/construct.rst
@@ -140,7 +140,7 @@ DOK arrays also support fancy indexing assignment if and only if all dimensions 
    s[[0, 2], [2, 1], [0, 1]] = 5
    s[[0, 3], [0, 4], [0, 1]] = [1, 5]
 
-Alongside indexing assignment and retrieval, DOK arrays support any arbitrary broadcasting function
+Alongside indexing assignment and retrieval, :obj:`DOK` arrays support any arbitrary broadcasting function
 to any number of arguments where the arguments can be :obj:`SparseArray` objects, :obj:`scipy.sparse.spmatrix`
 objects, or :obj:`numpy.ndarrays`. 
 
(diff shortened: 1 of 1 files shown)


hameerabbasi (commented):
It'd be nice to note all standard ufuncs and operators are now supported on DOK minus reduce.

docs/construct.rst
```diff
@@ -140,8 +140,31 @@ DOK arrays also support fancy indexing assignment if and only if all dimensions
...
+to any number of arguments where the arguments can be :obj:`SparseArray` objects, :obj:`scipy.sparse.spmatrix`
+objects, or :obj:`numpy.ndarrays`. 
+
+.. code-block:: python
+
+   x = sparse.random((10, 10), 0.5, format="dok")
+   y = sparse.random((10, 10), 0.5, format="dok")
+   sparse.elemwise(np.add, x, y)
+
+:obj:`DOK` arrays also support standard `__ufunc__` operations and operators, including comparison operators,
```
hameerabbasi, line 153: I don't think `__ufunc__` is a thing.

docs/construct.rst
```diff
@@ -140,8 +140,31 @@ DOK arrays also support fancy indexing assignment if and only if all dimensions
...
+the following code will perform elementwise equality comparison on the two arrays
+and return a new boolean :obj:`DOK` array.
+
+.. code-block:: python
+
+   x = sparse.random((10, 10), 0.5, format="dok")
+   y = np.random.random((10, 10))
+   x == y
+
+:obj:`DOK` arrays are returned from elemwise functions and standard `__ufunc__` operations if and only if all
```
hameerabbasi, line 164: Same here.

DragaDoncila: I changed it to just say universal... I'm not sure if there's a better technical term to refer to it by

hameerabbasi:
I think the technical term would be operators and ufuncs, without the double-underscore.

DragaDoncila:
@hameerabbasi I've made that change. Is there anything else you'd like documented?

hameerabbasi (approved):

hameerabbasi:
Thanks, @DragaDoncila!

PR #454, opened 2021-03-26
Fix typo in var function docstring

simonthor:
In the docstring of the `var` function, it says "gi66ven" instead of "given". I assume that this is a typo?
I created a pull request but maybe you would have preferred that I opened an issue instead.


hameerabbasi (approved):
Just waiting on CI to turn green!

hameerabbasi:
Thanks, @simonthor!

PR #455, opened 2021-03-29
Update docs for GCXS and CSR/CSC

daletovar:
Adds GCXS to operations.rst and adds GCXS and CSR/CSC to the completed tasks section of the roadmap.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,7 +18,7 @@ where scipy.sparse matrices are not well suited,
 but it should not be considered a full replacement.
 The data structures in pydata/sparse complement and can
 be used in conjunction with the fast linear algebra routines
-inside scipy.sparse.
+inside scipy.sparse. A format conversion or copy may be required.
 
 
 Motivation
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -16,8 +16,9 @@ rather than the :obj:`numpy.matrix` interface used in :obj:`scipy.sparse`
 These differences make this project useful in certain situations
 where scipy.sparse matrices are not well suited,
 but it should not be considered a full replacement.
-It lacks layouts that are not easily generalized like CSR/CSC
-and depends on scipy.sparse for some computations.
+The data structures in pydata/sparse complement and can
+be used in conjunction with the fast linear algebra routines
+inside scipy.sparse.
 
 
 Motivation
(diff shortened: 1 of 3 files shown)


docs/operations.rst
```diff
@@ -24,11 +24,6 @@ However some operations are not supported, like operations that
 implicitly cause dense structures, or numpy functions that are not
 yet implemented for sparse arrays.
 
-.. code-block:: python
-
-   np.svd(x)  # sparse svd not implemented
-
-
```
hameerabbasi, line 31: I wouldn't remove this.

docs/operations.rst
```diff
@@ -173,12 +168,19 @@ currently work:
    np.max(x)
    np.min(x, axis=(0, 2))
    x.prod()
+   x.argmax()
+   x.argmin()
+   np.mean()
+   np.var()
+   np.std()
+   x.any()
+   x.all()
```
hameerabbasi, line 177: This is already in the API docs; we don't need to make this page a complete reference.

docs/operations.rst
```diff
@@ -173,12 +168,19 @@ currently work:
...
 
 .. note::
    If you are performing multiple reductions along the same axes, it may
    be beneficial to call :obj:`COO.enable_caching`.
```
hameerabbasi, line 179: Caching was removed; I'd remove this statement.

docs/roadmap.rst
```diff
@@ -90,3 +90,10 @@ creation functions) becomes available through array protocols, it will be automa
 
 Support for ``scipy.sparse.linalg`` has been completed. We hope to add support for ``scipy.sparse.csgraph``
 in the future.
+
+More Storage Formats
+-------------------------------------
+
+GCXS, a compressed n-dimensional array format based on the GCRS/GCCS formats of
+`Shaikh and Hasan 2015 <https://ieeexplore.ieee.org/document/7237032>`_, has been added.
+In conjunction with this work, the CSR/CSC matrix formats have been are now a part of pydata/sparse.
```
hameerabbasi, line 99: I'd note here that a lot of operations still need speeding up.

hameerabbasi (commented):
In general the changes are helpful, but I'd make a couple of changes.

docs/operations.rst
daletovar: For sure. I think if we include that it should be `np.linalg.svd` and we should make it clear that sparse svd is supported by Scipy.

daletovar: Thinking about it a bit more, because we're trying to highlight "numpy functions that are **not yet** implemented for sparse arrays", perhaps we could use `np.linalg.cholesky`?

daletovar:
I'm seeing that I should update index.rst as well. I'll make a commit for that and if you think it's good we can merge after that.

daletovar:
@hameerabbasi, thanks for reviewing. Let me know if this looks good to you.

docs/index.rst
```diff
@@ -16,8 +16,9 @@ rather than the :obj:`numpy.matrix` interface used in :obj:`scipy.sparse`
 These differences make this project useful in certain situations
 where scipy.sparse matrices are not well suited,
 but it should not be considered a full replacement.
-It lacks layouts that are not easily generalized like CSR/CSC
-and depends on scipy.sparse for some computations.
+The data structures in pydata/sparse complement and can
+be used in conjunction with the fast linear algebra routines
+inside scipy.sparse.
```
hameerabbasi, line 21: I'd mention here that a conversion or a copy may be required.

hameerabbasi (changes requested):
This should be the last comment. Thanks so much for all the hard work, @daletovar!

daletovar:
Thanks so much, @hameerabbasi. How does this look?

hameerabbasi (approved):

hameerabbasi:
Thanks, @daletovar!

PR #456, opened 2021-04-07
Attempt to reformat docstring to match Numpydoc standard.

Carreau:
This uses an automatic reformatter to fix some of the common mistakes.
A few manual were needed, for example Numpydoc really want section to be separated by empty lines, or it start to say that `Return`, `------` are parameter of a function.

Spaces around colons or not in parameters have a different semantic meaning when parsing – it may not show in sphinx html, but internally the data is different and is messed up for crosslinks.

This also fixed a couple of parameter names typos (x/a or coord/shape/ndim inversion ...)

### DIFF
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -1341,6 +1341,7 @@ def full(shape, fill_value, dtype=None, format="coo", compressed_axes=None):
         A format string.
     compressed_axes : iterable, optional
         The axes to compress if returning a GCXS array.
+
     Returns
     -------
     out : SparseArray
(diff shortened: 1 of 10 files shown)


hameerabbasi (approved):
Thanks, @Carreau! Would it be possible to add a lint to CI? If not, I'll just put this in as-is.

Carreau:
You could run `velin . --check`, but I'm still actively working on it, meaning that failures would likely be due to me implementing more checks/heuristics. It's also mor designed to be a pre-commit hook as it automatically fixes things, but so far there are still fixes that are incorrect, so I still check manually every changes.

You can follow the progress or report bugs here https://github.com/carreau/velin

PR #458, opened 2021-04-15
Slicing and advanced indexing for DOK

### LINKED ISSUE #298: Support for DOK slicing and advanced indexing
Since the DOK array is versatile applicable I use them frequently in my projects. I needed support for slicing and advanced indexing and implemented it myself, see [gist](https://gist.github.com/stschroe/7c4660774835258c06b2be968eb49204#file-sparse_dok_indexer-py). 

The indexing supports slicing, advanced (fancy) indexing and as I need it most of the time "array indexing". This is a fast version for advanced indexing. Only vectors (1d sequences) are accepted for indexing. The getitem  method always returns dense ndarrays in contrast to normal advanced indexing of sparse arrays which create new copies of sparse arrays. The setitem method allows only vectors and scalars as values. This allows fast calculations over ndarray methods without the expensive creation of temporary sparse arrays between operations. This is of course only useful for small vectors, which fit easily in to memory.

If there is interest in adding this to pydata.sparse, I could help. But as I have never contributed to any open project, I have no idea if I could handle the necessary amount of work.

sayandip18:
Fixes #298

### DIFF
--- a/sparse/_dok_indexer.py
+++ b/sparse/_dok_indexer.py
@@ -1,0 +1,295 @@
+from itertools import product, repeat
+from numbers import Integral
+
+import numpy as np
+from sparse import DOK, SparseArray
+from sparse._slicing import normalize_index
+
+
+class DOKVectorIndexer:
+    """Vector indexing
+    Allows so called "fancy" or advanced indexing with one dimensional
+    sequences (here named as vectors) as array indices. As opposed to
+    numpy legacy indexing, the vector result dimensions are always the
+    first axes of the result array. This is inspired by NEP 21:
+    https://www.numpy.org/neps/nep-0021-advanced-indexing.html
+    It is not fully implemented yet. Especially the setting of array
+    elements.
+    """
+
+    def __init__(self, dok):
+        self._dok = dok
+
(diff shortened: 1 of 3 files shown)


hameerabbasi (commented):
Can we integrate this into the `DOK` type? I see some monkeypatching going on.

sayandip18:
> Can we integrate this into the `DOK` type?

Do you mean including `vix` as a function in `DOK`?

hameerabbasi:
> Do you mean including `vix` as a function in `DOK`?

I mean as an property.

sayandip18:
Calling the function `add_vector_indexer_to_sparse_dok()` essentially does that. Maybe we can import and call the function in `_dok.py` which will give us the desired behaviour.

hameerabbasi:
> Calling the function `add_vector_indexer_to_sparse_dok()` essentially does that. Maybe we can import and call the function in `_dok.py` which will give us the desired behaviour.

I'd prefer for `vix` to be directly added to all `DOK` objects.

hameerabbasi:
@sayandip18 Do you plan to continue this?

PR #466, opened 2021-04-24
Miscellaneous improvements in docs

sayandip18:
Changes which should have been included in #448 are continued here.


hameerabbasi (changes requested):
I'd just say `SparseArray` instead of promising any of the subclasses.

PR #467, opened 2021-04-24
Improve matmul benchmark

sayandip18:
This is WIP. More commits incoming.

### DIFF
--- a/benchmarks/benchmark_matmul.py
+++ b/benchmarks/benchmark_matmul.py
@@ -12,7 +12,7 @@ class Matmul_Sparse:
 
         if(dens_arg==1):
             self.x = self.x.todense()
-        elif(dens_arg==0)
+        elif(dens_arg==0):
             self.y = self.y.todense()
 
         self.x @ self.y
--- a/benchmarks/benchmark_matmul.py
+++ b/benchmarks/benchmark_matmul.py
@@ -1,0 +1,16 @@
+import numpy as np
+import sparse
+
+
+class Matmul_Sparse:
+    params = (["coo", "gcxs"], [0.01, 0.33, 0.5, 1.0])
+
+    def setup(self, p, dens_arg):
+        np.random.seed(0)
+        self.x = sparse.random((100, 100), density=0.01, format=p)
+        self.y = sparse.random((100, 100), density=dens_arg, format=p)
+
+        self.x @ self.y
+
+    def time_matmul(self, p, dens_arg):
+        self.x @ self.y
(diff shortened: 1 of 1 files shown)


hameerabbasi (changes requested):
Maybe you could parametrize things instead of using multiple separate benchmarks?

sayandip18:
If the changes are okay, I'll squash the commits

benchmarks/benchmark_matmul.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
+import sparse
+
+
+class Matmul_Sparse:
```
hameerabbasi, line 5: Maybe these two classes could also be parameterised.

Something like `dense_arg = [0, 1, None]`.

hameerabbasi (commented):
I'd like one last change, parametrise all the things!

benchmarks/benchmark_matmul.py
```diff
@@ -0,0 +1,16 @@
...
+    params = (["coo", "gcxs"], [0.01, 0.33, 0.5, 1.0])
+
+    def setup(self, p, dens_arg):
+        np.random.seed(0)
+        self.x = sparse.random((100, 100), density=0.01, format=p)
+        self.y = sparse.random((100, 100), density=dens_arg, format=p)
```
hameerabbasi, line 11: This still produces a GCXS when the dense argument is `1.0`, what I meant in https://github.com/pydata/sparse/pull/467/files#r633073348 was that `0` means `self.x` should be `.todense()`, and `1` means the same for `self.y`.

benchmarks/benchmark_matmul.py
```diff
@@ -3,12 +3,17 @@
...
-    params = (["coo", "gcxs"], [0.01, 0.33, 0.5, 1.0])
+    params = (["coo", "gcxs"], [0, 1, None])
 
     def setup(self, p, dens_arg):
         np.random.seed(0)
         self.x = sparse.random((100, 100), density=0.01, format=p)
-        self.y = sparse.random((100, 100), density=dens_arg, format=p)
+        self.y = sparse.random((100, 100), density=0.01, format=p)
+
+        if(dens_arg==1):
```
hameerabbasi, line 13:
```suggestion
        if(dens_arg==0):
```

benchmarks/benchmark_matmul.py
```diff
@@ -3,12 +3,17 @@
...
+            self.x = self.x.todense()
+        elif(dens_arg==0):
```
hameerabbasi, line 15:
```suggestion
        elif(dens_arg==1):
```

hameerabbasi (approved):
Just waiting on CI to turn green.

PR #471, opened 2021-05-03
Pytest failures - fixed with black re-formatting

GenevieveBuckley:
This is super minor, but I noticed when I forked the repository (just to look at things), a couple of tests fail when you run `pytest --pyargs sparse` as described in the [contributing guide](https://github.com/pydata/sparse/blob/master/docs/contributing.rst).

The reason for the two test failures is some whitespace that the black formatting tool doesn't like. So it's nothing important, but it's always nice to have things work smoothly when new people go through the contributing guide.

### DIFF
--- a/sparse/_dok.py
+++ b/sparse/_dok.py
@@ -393,7 +393,7 @@ class DOK(SparseArray, NDArrayOperatorsMixin):
         )
 
     def _filter_by_key(self, coords, slice_key):
-        """Filter data coordinates to be within given slice """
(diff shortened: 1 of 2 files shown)


hameerabbasi:
Ah, yes. Black adds new rules every now and then, and things break.

GenevieveBuckley:
> Ah, yes. Black adds new rules every now and then, and things break.

Yeah. Slightly annoying, but what can you do

hameerabbasi:
Thanks, @GenevieveBuckley!

PR #472, opened 2021-05-05
Add hypothesis tests

sayandip18:


sparse/tests/test_dot.py
```diff
@@ -316,3 +317,10 @@ def test_complex(dtype1, dtype2, format1, format2):
     x1, x2 = dense_convertor(s1), dense_convertor(s2)
 
     assert_eq(x1 @ x2, s1 @ s2)
+
+
+@given(x=st.integers(min_value=0), y=st.integers(min_value=0))
```
hameerabbasi, line 322: Do you know the maximum value here? We don't want to create arrays that are too large.

hameerabbasi (changes requested):
This is a good start! Thanks! My comment on Gitter was more about moving the existing test suite.

sparse/tests/test_dot.py
sayandip18: Will an upper limit of 100 be okay?

hameerabbasi: 10 is more than enough, I think. 😄

sayandip18:
> My comment on Gitter was more about moving the existing test suite.

I intended this to be a starting point since this is the first time I'm using Hypothesis. I'll keep pushing more changes.

hameerabbasi:
On the next commit, CI should trigger, I just pushed an Azure Pipelines config that should allow it.

hameerabbasi:
Changed base to `master`, since this is generally useful.

sparse/tests/test_compressed.py
```diff
@@ -79,8 +82,10 @@ def test_reductions_float16(random_sparse, reduction, kwargs, axis):
 
 
 @pytest.mark.parametrize("reduction,kwargs", [("any", {}), ("all", {})])
```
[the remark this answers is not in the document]
sayandip18: Yes I understand, more commits are on the way. My class tests concluded yesterday and I'm aiming to finish the remaining work in a day or two.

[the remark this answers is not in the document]
hameerabbasi: These aren't necessarily criticisms, more like reminders of what still needs to be done, for both of us. 😄 

Don't take them as such -- They're only meant to guide future me/you.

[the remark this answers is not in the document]
sayandip18: I knew it wasn't a criticism, I'm just informing that I'll be able to devote more time on this from today :grin:

sayandip18:
Do we intend to replace all instances of `pytest.mark.parametrize` with `hypothesis.given`?

There are some inherent problems regarding the randomized nature of Hypothesis. Consider this example
https://github.com/pydata/sparse/blob/146f1129835547f2efb6ba8a55eb4eca320cbcdd/sparse/tests/test_dok.py#L59-L75

The hypothesis strategy I implemented looks like this
```
@given(
    st.one_of(
        st.tuples(
            st.tuples(st.integers(min_value=1, max_value=5)),
            st.dictionaries(
                st.integers(min_value=0, max_value=0),
                st.integers(min_value=0, max_value=4),
            ),
        ),
        st.tuples(
            st.tuples(
                st.integers(min_value=1, max_value=5),
                st.integers(min_value=1, max_value=5),
            ),
            st.dictionaries(
                st.tuples(
                    st.integers(min_value=0, max_value=0),
                    st.integers(min_value=0, max_value=0),
                ),
                st.integers(min_value=0, max_value=4),
            ),
        ),
    )
)
def test_construct(sd):
    shape, data = sd
    s = DOK(shape, data)
    x = np.zeros(shape, dtype=s.dtype)

    for c, d in data.items():
        x[c] = d

    assert_eq(x, s)
```
 The strategy is very complex and its randomized nature ensures that I cannot put `max_value` > 0 in any of the `st.dictionaries` to avoid error. Is there some simpler way out?

hameerabbasi:
Perhaps you could post the error that happens and we can work it out together?

hameerabbasi:
But if I understand correctly, you can use [`basic_indices`](https://hypothesis.readthedocs.io/en/latest/numpy.html?highlight=numpy#hypothesis.extra.numpy.basic_indices), perhaps.

hameerabbasi:
Combined with [composite strategies](https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies).

sayandip18:
The error message is quite large. The brief one is 
```
============================================================== short test summary info ==============================================================
FAILED sparse/tests/test_dok.py::test_construct - IndexError: Index is not smaller than dimension 1 >= 1

```

I am getting an out of bound index due to the randomness.

hameerabbasi:
I would define a two-way [composite strategy](https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies):

Generate a tuple of shape, data:
1. Generate a [shape](https://hypothesis.readthedocs.io/en/latest/numpy.html?highlight=shape#hypothesis.extra.numpy.array_shapes)
2. Generate a [dictionary](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.fixed_dictionaries) of unique [indices](https://hypothesis.readthedocs.io/en/latest/numpy.html?highlight=shape#hypothesis.extra.numpy.integer_array_indices) for the keys and [floats](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats) for the values.

sayandip18:
Getting a new issue

I replaced this
https://github.com/pydata/sparse/blob/146f1129835547f2efb6ba8a55eb4eca320cbcdd/sparse/tests/test_dok.py#L103-L110
with the following
```
@given(data())
def test_getitem(data):
    n = st.integers(min_value=1, max_value=5)
    shape = st.tuples(n, n)
    density = st.floats(min_value=0, max_value=1)
    print(type(density))
    indices = st.slices(st.integers(min_value=1, max_value=3))
    s = sparse.random(shape, density, format="dok")
    x = s.todense()

    sparse_sliced = s[indices]
    dense_sliced = x[indices]

    assert_eq(sparse_sliced.todense(), dense_sliced)
```
which is leading to the following error
```
    
        if density is None:
            density = 0.01
>       if not (0 <= density <= 1):
E       TypeError: '<=' not supported between instances of 'int' and 'LazyStrategy'

sparse/_utils.py:151: TypeError

FAILED sparse/tests/test_dok.py::test_getitem - TypeError: '<=' not supported between instances of 'int' and 'LazyStrategy'
```

hameerabbasi:
Don't be afraid of having CI red on a WiP PR. Just push up your code and it's easier to pull it down and take a look.

sparse/tests/test_dok.py
```diff
@@ -88,19 +119,24 @@ def test_getitem_single(shape, density):
...
+#         ((10, 10), 0.2, (slice(5), slice(0, 10, 3))),
+#         ((5, 5), 0.5, (slice(0, 4, 4), slice(0, 4, 4))),
+#         ((5, 5), 0.4, (1, slice(0, 4, 1))),
+#         ((10, 10), 0.8, ([0, 4, 5], [3, 2, 4])),
+#         ((10, 10), 0, (slice(10), slice(10))),
+#     ],
+# )
+@given(data())
+def test_getitem(data):
+    n = st.integers(min_value=1, max_value=5)
```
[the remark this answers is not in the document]
sayandip18: Yup this works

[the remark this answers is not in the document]
hameerabbasi: This can be simplified and generalized at the same time:

```python
@composite
def dok_shape_items():
    shape = draw(array_shapes(min_dims = 1, max_dims=3, min_side=1, max_side=10))
    items = draw(dictionaries(
        keys=basic_indices(shape),
        values=floats(),
        min_size=0,
        max_size=0.5 * np.prod(shape)
    ))
    return shape, items
```

And then use it as any other strategy. This also allows it to be re-usable.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 numpy>=1.17
 scipy>=0.19
 numba>=0.49
+hypothesis>=6.5
```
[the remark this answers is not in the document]
hameerabbasi: This is missing to get CI green.

hameerabbasi:
Right now I think CI will fail because `hypothesis` isn't a test dependency. Please add it to https://github.com/pydata/sparse/blob/master/requirements/tests.txt as per https://github.com/pydata/sparse/pull/472#discussion_r626936838.

sayandip18:
A lot of tests are failing now regardless of hypothesis being there is `test.txt`. Most Flaky, meaning they are exceeding the deadline of 200.00ms on the first run.

sayandip18:
**Note**: I have added the utility file

sayandip18:
Only a handful left now. I am hoping to finish this before the official coding period starts.

sayandip18:
I will squash the commits if the tests pass. The commit message should be something meaningful instead of how it is now.

hameerabbasi:
@sayandip18 Two things remain:

1. The tests run forever. Right now it hangs on `sparse/tests/test_compressed.py::test_reshape_2`.
2. There are a lot of `sparse.random` in the tests that should be replaced by `gen_sparse_random`.

hameerabbasi:
Can we replace the [`tmp_path`](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture with [`this`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)?

Replace

```python
def test_abc(tmp_path):
    ...
```

with

```python
def test_abc():
    with tempfile.TemporaryDirectory() as temp_path:
        ...
```

sparse/tests/_utils.py
```diff
@@ -0,0 +1,325 @@
...
+        array_shapes(min_dims=4, max_dims=6, min_side=3, max_side=5),
+    )
+
+    return shape, index, value_shape
+
+
+@composite
+def gen_transpose(draw):
+    a = draw(st.lists(st.integers(min_value=1, max_value=5), min_size=2, max_size=4))
+    if len(a) is 2:
```
hameerabbasi, line 98:
```suggestion
    if len(a) == 2:
```

sparse/tests/_utils.py
```diff
@@ -0,0 +1,325 @@
...
+        b = draw(st.sampled_from([(1, 0), (0, 1)]))
+    elif len(a) is 3:
```
hameerabbasi, line 100:
```suggestion
    elif len(a) == 3:
```

sparse/tests/_utils.py
```diff
@@ -0,0 +1,325 @@
...
+    if function:
+        reduction = getattr(np, reduction)
+
+    return reduction, kwargs
+
+
+@composite
+def gen_broadcast_shape(draw):
+    shape1 = draw(
+        st.lists(st.integers(min_value=2, max_value=5), min_size=2, max_size=3).map(
```
hameerabbasi, line 138:
```suggestion
        st.lists(st.integers(min_value=2, max_value=30), min_size=1, max_size=3).map(
```

sparse/tests/_utils.py
```diff
@@ -0,0 +1,325 @@
...
+            tuple
+        )
+    )
+    shape2 = draw(broadcastable_shapes(shape1, min_side=3))
```
hameerabbasi, line 142:
```suggestion
    shape2 = draw(broadcastable_shapes(shape1, min_side=2))
```

hameerabbasi (changes requested):
A few suggestions.

sparse/tests/_utils.py
```diff
@@ -0,0 +1,325 @@
...
+        )
+    else:
+        b = draw(st.sampled_from([sparse.random((100, 100), density=0.01)]))
+
+    return a, b
+
+
+@composite
+def gen_broadcast_shape_dot(draw):
+    a_shape = draw(
```
hameerabbasi, line 187: 1. Generate ANY two shapes with `1 <= ndim <= 4`.
2. If `ndim_a == 1: shape_b[-1] = shape_a[0]`
3. If `ndim_b == 1: shape_b[0] = shape_a[-2 if ndim_a != 1 else -1]`
4. Else `shape_b[-1] = shape_a[-2]`.

hameerabbasi: 1. For `matmul`, do the same as dot but with `ndim` fixed to `1 <= ndim <= 2`.
2. Then generate two separate broadcastable shapes `shape_sa, shape_sb` s.t. both have ndim between 0 and 3. If `ndim_a == 2: shape_a += shape_sa`
4. Repeat step 3 for `b`.

sparse/tests/_utils.py
```diff
@@ -0,0 +1,325 @@
...
+    kwargs["axis"] = draw(st.sampled_from([None, 0, 1, 2, (0, 2), -3, (1, -1)]))
+    kwargs["keepdims"] = draw(st.sampled_from([True, False]))
+    if function:
+        reduction = getattr(np, reduction)
+
+    return reduction, kwargs
+
+
+@composite
+def gen_broadcast_shape(draw):
```
hameerabbasi, line 136: https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.mutually_broadcastable_shapes

hameerabbasi:
> Can we replace the [`tmp_path`](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture with [`this`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)?
> 
> Replace
> 
> ```python
> def test_abc(tmp_path):
>     ...
> ```
> 
> with
> 
> ```python
> def test_abc():
>     with tempfile.TemporaryDirectory() as temp_path:
>         ...
> ```

This still needs doing, if I'm not mistaken.

sayandip18:
I'm getting these two failures for `test_io.py`

```python
===================================================================== FAILURES =====================================================================
_____________________________________________________________ test_save_load_npz_file ______________________________________________________________

    @given(
>       compression=st.sampled_from([True, False]), format=st.sampled_from(["coo", "gcxs"])
    )
    def test_save_load_npz_file(compression, format):

sparse/tests/test_io.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

compression = True, format = 'coo'

    @given(
        compression=st.sampled_from([True, False]), format=st.sampled_from(["coo", "gcxs"])
    )
    def test_save_load_npz_file(compression, format):
        with tempfile.TemporaryDirectory() as tmp_path:
            x = sparse.random((2, 3, 4, 5), density=0.25, format=format)
            y = x.todense()
    
>           filename = tmp_path / "mat.npz"
E           TypeError: unsupported operand type(s) for /: 'str' and 'str'

sparse/tests/test_io.py:20: TypeError
-------------------------------------------------------------------- Hypothesis --------------------------------------------------------------------
Falsifying example: test_save_load_npz_file(
    compression=True, format='coo',
)
_________________________________________________________ test_load_wrong_format_exception _________________________________________________________

    def test_load_wrong_format_exception():
        with tempfile.TemporaryDirectory() as tmp_path:
            x = np.array([1, 2, 3])
    
>           filename = tmp_path / "mat.npz"
E           TypeError: unsupported operand type(s) for /: 'str' and 'str'

sparse/tests/test_io.py:31: TypeError

```

hameerabbasi:
How about doing `with ... as tmp_path_str` and then `tmp_path = pathlib.Path(tmp_path_str)`?

sayandip18:
Passed!

hameerabbasi:
I think you need to merge master.

hameerabbasi:
@sayandip18 Do you plan on continiuing this?

PR #474, opened 2021-05-10
Added implementation of sparse.pad()

### LINKED ISSUE #438: Implement sparse.COO.pad
**Is your feature request related to a problem? Please describe.**
I would like to be able to use the xarray.DataArray API when the data is sparse.COO.

Instead—as mentioned at https://github.com/pydata/sparse/issues/1#issuecomment-792342987 —it appears I currently need to write a custom subclass of DataArray that handles cases where sparse doesn't implement a method expected by xarray.

**Describe the solution you'd like**
For `sparse.COO.pad` and the other methods mentioned in the linked comment to be implemented so that, in general, it is possible to use the [xarray.DataArray API](http://xarray.pydata.org/en/stable/api.html#dataarray), or at least all a certain advertised set of basic functions, with sparse data.

**Describe alternatives you've considered**
As mentioned, I am currently handling this with wrapper code that converts an xr.DataArray to dense data, performs operations, then re-converts the data to sparse.

Another mitigation would be for the sparse documentation to describe which DataArray operations are not supported.

Some example code:
```python
import pandas as pd
import xarray as xr
from numpy import nan

idx = dict(
    columns=pd.Index(["x1", "x2", "x3", "x4"], name="x"),
    index=pd.Index(["y1", "y2", "y3", "y4"], name="y"),
)

# Simplified example data
s1 = pd.DataFrame(
    [
        [1.0, nan, nan, nan],
        [nan, nan, 2.0, nan],
        [nan, nan, nan, 3.0],
        [nan, 4.0, nan, nan],
    ],
    **idx
).stack()

print(s1)

da1 = xr.DataArray.from_series(s1, sparse=True)

print(da1)

# I would like to produce data like the following by using DataArray.shift()
s2 = pd.DataFrame(
    [
        [nan, 1.0, nan, nan],
        [nan, nan, nan, 2.0],
        [nan, nan, nan, nan],
        [nan, nan, 4.0, nan],
    ],
    **idx
).stack()

print(s2)

da2 = xr.DataArray.from_series(s2)

print(da2)

# This line raises an exception, below
da1.shift(x=1)
```

The exception raised:
```
  File "test.py", line 42, in <module>
    da1.shift(x=1)
  File "/home/khaeru/.local/lib/python3.8/site-packages/xarray/core/dataarray.py", line 3077, in shift
    variable = self.variable.shift(
  File "/home/khaeru/.local/lib/python3.8/site-packages/xarray/core/variable.py", line 1205, in shift
    result = result._shift_one_dim(dim, count, fill_value=fill_value)
  File "/home/khaeru/.local/lib/python3.8/site-packages/xarray/core/variable.py", line 1166, in _shift_one_dim
    data = duck_array_ops.pad(
  File "/home/khaeru/.local/lib/python3.8/site-packages/xarray/core/duck_array_ops.py", line 56, in f
    return wrapped(*args, **kwargs)
  File "<__array_function__ internals>", line 5, in pad
TypeError: no implementation found for 'numpy.pad' on types that implement __array_function__: [<class 'sparse._coo.core.COO'>]
```

H4R5H1T-007:
Fixes #438

### DIFF
--- a/docs/generated/sparse.COO.rst
+++ b/docs/generated/sparse.COO.rst
@@ -73,12 +73,12 @@ COO
       COO.copy
       COO.dot
       COO.flatten
+      COO.pad
       COO.reshape
       COO.resize
       COO.transpose
       COO.swapaxes
       COO.nonzero
-      COO.pad
 
    .. rubric:: Utility functions
    .. autosummary::
--- a/docs/generated/sparse.COO.rst
+++ b/docs/generated/sparse.COO.rst
@@ -73,7 +73,6 @@ COO
       COO.copy
       COO.dot
       COO.flatten
-      COO.pad
       COO.reshape
       COO.resize
       COO.transpose
--- a/docs/generated/sparse.COO.rst
+++ b/docs/generated/sparse.COO.rst
@@ -78,6 +78,7 @@ COO
       COO.transpose
       COO.swapaxes
       COO.nonzero
+      COO.pad
 
    .. rubric:: Utility functions
    .. autosummary::
--- a/docs/generated/sparse.GCXS.pad.rst
+++ b/docs/generated/sparse.GCXS.pad.rst
@@ -1,0 +1,6 @@
+﻿GCXS.pad
+========
+
+.. currentmodule:: sparse
+
+.. automethod:: GCXS.pad
--- a/docs/generated/sparse.GCXS.pad.rst
+++ b/docs/generated/sparse.GCXS.pad.rst
@@ -1,6 +1,0 @@
-﻿GCXS.pad
-========
-
-.. currentmodule:: sparse
-
-.. automethod:: GCXS.pad
--- a/docs/generated/sparse.GCXS.rst
+++ b/docs/generated/sparse.GCXS.rst
@@ -81,6 +81,8 @@ GCXS
       GCXS.mean
    
       GCXS.min
+
+      GCXS.pad
    
       GCXS.prod
    
--- a/docs/generated/sparse.GCXS.rst
+++ b/docs/generated/sparse.GCXS.rst
@@ -81,8 +81,6 @@ GCXS
       GCXS.mean
    
       GCXS.min
-
-      GCXS.pad
    
       GCXS.prod
    
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -1687,6 +1687,9 @@ def pad(array, pad_width, mode="constant", **kwargs):
 
     Parameters
     ----------
+    array : SparseArray
+        Sparse array which is to be padded.
+
     pad_width : {sequence, array_like, int}
         Number of values padded to the edges of each axis. ((before_1, after_1), … (before_N, after_N)) unique pad widths for each axis. ((before, after),) yields same before and after pad for each axis. (pad,) or int is a shortcut for before = after = pad width for all axes.
 
@@ -1728,16 +1731,8 @@ def pad(array, pad_width, mode="constant", **kwargs):
     if kwargs:
         raise NotImplementedError("Additional Unknown arguments present.")
 
-    from ._compressed import GCXS
-    from ._dok import DOK
     from ._coo import COO
 
-    if isinstance(array, GCXS):
-        ftype = "gcxs"
-    elif isinstance(array, DOK):
-        ftype = "dok"
-    else:
-        ftype = "coo"
     array = array.asformat("coo")
 
     pad_width = np.broadcast_to(pad_width, (len(array.shape), 2))
@@ -1749,6 +1744,4 @@ def pad(array, pad_width, mode="constant", **kwargs):
         ]
     )
     new_data = array.data
-    return COO(new_coords, new_data, new_shape, fill_value=array.fill_value).asformat(
-        ftype
-    )
+    return COO(new_coords, new_data, new_shape, fill_value=array.fill_value)
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -9,7 +9,13 @@ from numba import literal_unroll
 import warnings
 
 from ._sparse_array import SparseArray
-from ._utils import check_compressed_axes, normalize_axis, check_zero_fill_value
+from ._utils import (
+    check_compressed_axes,
+    normalize_axis,
+    check_zero_fill_value,
+    equivalent,
+    _zero_of_dtype,
+)
 
 from ._umath import elemwise
 from ._coo.common import (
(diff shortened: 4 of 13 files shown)


hameerabbasi (commented):
This is pretty good already. Just needs a few finishing touches.

In addition to this, look at `docs/generated/sparse.COO.rst` and add an entry there, as described here: https://sparse.pydata.org/en/stable/contributing.html#adding-building-the-documentation

sparse/_coo/core.py
```diff
@@ -1451,6 +1451,57 @@ def asformat(self, format, compressed_axes=None):
...
+
+        See Also
+        --------
+        :obj:`numpy.pad` : NumPy equivalent function
+
+        """
+        if mode.lower() != "constant":
+            raise NotImplementedError("Mode '{}' is not supported.".format(mode))
+
+        if kwargs.pop("constant_values", self.fill_value) != self.fill_value:
```
hameerabbasi, line 1488: This will run into some issues with `nan`s and so forth, use `equivalent` to compare these. Also the default `constant_values` is `0`, not `self.fill_value`. Use `_zero_of_dtype(self.dtype)` to get the default constant value.

sparse/_coo/core.py
```diff
@@ -1451,6 +1451,57 @@ def asformat(self, format, compressed_axes=None):
...
+            raise ValueError("constant_values can only be equal to fill value.")
+
+        if bool(kwargs):
```
hameerabbasi, line 1491:
```suggestion
        if kwargs:
```

sparse/_coo/core.py
H4R5H1T-007: But Sir in our case shouldn't constant value default be fill value?

hameerabbasi: We want to match the NumPy API as exactly as possible. 😄

sparse/_coo/core.py
```diff
@@ -1485,10 +1488,10 @@ def pad(self, pad_width, mode="constant", **kwargs):
         if mode.lower() != "constant":
             raise NotImplementedError("Mode '{}' is not supported.".format(mode))
 
-        if kwargs.pop("constant_values", self.fill_value) != self.fill_value:
+        if kwargs.pop("constant_values", _zero_of_dtype(self.dtype)) != self.fill_value:
```
hameerabbasi, line 1491: You missed checking the `constant_values` with `equivalent`. If the fill-value is `nan` this will have issues, as `np.nan != np.nan` returns `True`.

sparse/_coo/core.py
```diff
@@ -1451,6 +1451,60 @@ def asformat(self, format, compressed_axes=None):
...
+
+        Parameters
+        ----------
+        pad_width : {Sequence, Array_like, Int}
+            Number of values padded to the edges of each axis. ((before_1, after_1), … (before_N, after_N)) unique pad widths for each axis. ((before, after),) yields same before and after pad for each axis. (pad,) or int is a shortcut for before = after = pad width for all axes.
+
+        mode : str
+            Pads to a constant value which is fill value. Currently only constant mode is implemented
+
+        constant_values : Int
```
hameerabbasi, line 1467:
```suggestion
        constant_values : int
```

sparse/_coo/core.py
```diff
@@ -1451,6 +1451,60 @@ def asformat(self, format, compressed_axes=None):
...
         raise NotImplementedError("The given format is not supported.")
 
+    def pad(self, pad_width, mode="constant", **kwargs):
+        """
+        Performs the equivalent of :obj:`numpy.pad` for :obj:`COO`. Note that
+        this function returns a new array instead of a view.
+
+        Parameters
+        ----------
+        pad_width : {Sequence, Array_like, Int}
```
hameerabbasi, line 1461:
```suggestion
        pad_width : int
```

That's the only one we support.

sparse/_coo/core.py
```diff
@@ -1451,6 +1451,60 @@ def asformat(self, format, compressed_axes=None):
...
+            Number of values padded to the edges of each axis. ((before_1, after_1), … (before_N, after_N)) unique pad widths for each axis. ((before, after),) yields same before and after pad for each axis. (pad,) or int is a shortcut for before = after = pad width for all axes.
+
+        mode : str
+            Pads to a constant value which is fill value. Currently only constant mode is implemented
+
+        constant_values : Int
+            The values to set the padded values for each axis. Default is 0.
```
hameerabbasi, line 1468: Comment that it must be the same as `fill_value`.

sparse/_coo/core.py
```diff
@@ -1451,6 +1451,60 @@ def asformat(self, format, compressed_axes=None):
...
+        ValueError
+            If constant_values != self.fill_value
+
+        See Also
+        --------
+        :obj:`numpy.pad` : NumPy equivalent function
+
+        """
+        if mode.lower() != "constant":
+            raise NotImplementedError("Mode '{}' is not supported.".format(mode))
```
hameerabbasi, line 1489: Use f-strings here.
```suggestion
            raise NotImplementedError(f"Mode '{mode}' is not yet supported.")
```

docs/generated/sparse.COO.rst
```diff
@@ -78,6 +78,7 @@ COO
       COO.transpose
       COO.swapaxes
       COO.nonzero
+      COO.pad
```
hameerabbasi, line 81: Move this so the list is in alphabetical order.

hameerabbasi (commented):
A couple of small changes still.

sparse/_coo/core.py
H4R5H1T-007: Sir pad_width can be arbitrary just like numpy

hameerabbasi: Ah, okay. Then can you make it `{np.ndarray, int}`

H4R5H1T-007: Sir we can also send a sequence of tuples of lists also. I have checked it it's working.

hameerabbasi:
```suggestion
        pad_width : {sequence, array_like, int}
```

hameerabbasi (commented):
Oh, missed the most important thing. Could you change this to a function from a method, so `sparse.pad` instead of `sparse.COO.pad`?

The documentation will need to be moved too.

H4R5H1T-007:
> Could you change this to a function from a method, so sparse.pad instead of sparse.COO.pad?

Sir does that means we need to implement pad function for other data structures also?

hameerabbasi:
> Sir does that means we need to implement pad function for other data structures also like GCXS?

No, you can just do `array.asformat(COO)` and document it as returning a SparseArray. Also test arrays as well as scalars for `constant_values`.

H4R5H1T-007:
> In other words -- The method should not exist on any class.

But sir then how can we implement np.pad(COO)?

docs/generated/sparse.GCXS.pad.rst
```diff
@@ -0,0 +1,6 @@
+﻿GCXS.pad
```
hameerabbasi, line 1: Delete this file.

docs/generated/sparse.GCXS.rst
```diff
@@ -81,6 +81,8 @@ GCXS
       GCXS.mean
    
       GCXS.min
+
+      GCXS.pad
```
hameerabbasi, line 85: Delete these lines.

sparse/_common.py
```diff
@@ -1672,3 +1678,77 @@ def moveaxis(a, source, destination):
...
+    from ._compressed import GCXS
+    from ._dok import DOK
+    from ._coo import COO
+
+    if isinstance(array, GCXS):
+        ftype = "gcxs"
+    elif isinstance(array, DOK):
+        ftype = "dok"
+    else:
+        ftype = "coo"
```
hameerabbasi, line 1740: No need to convert/convert back.

sparse/_common.py
```diff
@@ -1672,3 +1678,77 @@ def moveaxis(a, source, destination):
...
+    new_shape = tuple(
+        [
+            array.shape[i] + pad_width[i, 0] + pad_width[i, 1]
+            for i in range(len(array.shape))
+        ]
+    )
+    new_data = array.data
+    return COO(new_coords, new_data, new_shape, fill_value=array.fill_value).asformat(
+        ftype
+    )
```
hameerabbasi, line 1754: Same here.

sparse/_compressed/compressed.py
```diff
@@ -810,6 +810,42 @@ def _prune(self):
...
+
+        ValueError
+            If constant_values != self.fill_value
+
+        See Also
+        --------
+        :obj:`numpy.pad` : NumPy equivalent function
+
+        """
+        return pad(self, pad_width, mode=mode, **kwargs)
```
hameerabbasi, line 847: No need for this function at all. `np.pad` calls `__array_function__` which works just fine with `sparse.pad`

sparse/_coo/core.py
```diff
@@ -1451,6 +1451,42 @@ def asformat(self, format, compressed_axes=None):
 
         raise NotImplementedError("The given format is not supported.")
 
+    def pad(self, pad_width, mode="constant", **kwargs):
+        """
```
hameerabbasi, line 1455: Same here, no need for this function.

sparse/_common.py
```diff
@@ -1672,3 +1678,77 @@ def moveaxis(a, source, destination):
...
     return result
+
+
+def pad(array, pad_width, mode="constant", **kwargs):
+    """
+    Performs the equivalent of :obj:`numpy.pad` for :obj:`SparseArray`. Note that
+    this function returns a new array instead of a view.
+
+    Parameters
+    ----------
```
hameerabbasi, line 1689: Add `array: SparseArray` here.

PR #487, opened 2021-06-10
Use cmake and scikit-build

sayandip18:

PR #496, opened 2021-06-27
supercedes #487

leej3:
sets CMAKE_RC_COMPILER
supports building pybind11 from source if missing.

cc: @hameerabbasi


leej3:
Looks like windows is failing due to this problem:
https://stackoverflow.com/questions/31629966/how-to-set-cmake-cxx-variable-on-windows

hameerabbasi:
Can we put an if condition in the bash script to check if it's Windows and execute that batch file?

leej3:
I'm inclined to try and use a github action. Knowing where that bat file is on any given version of visual studio seems tricky.  This one looks like it might be useful:
https://github.com/seanmiddleditch/gha-setup-vsdevenv

I don't know windows very well but can give it a try later if you don't get to it.

.github/workflows/ci.yml
```diff
@@ -33,6 +33,7 @@ jobs:
           vswhere: vswhere.exe
       - name: Install package
         run: |
+          if grep -iq  win <(echo "${OSTYPE}");then export CXX=cl.exe;fi
```
hameerabbasi, line 34: `darwin` also includes `win`.

leej3:
Ha. silly me.

hameerabbasi:
In this goes, thanks @leej3!

leej3:
Cool. I wish I could be more help with the windows stuff. It’s pretty alien to me so far though. Let me know if the pybind11 incorporation doesn’t work as expected and I’ll take another look. As far as I could tell it was solely a development dependency and so didn’t install anything from its build but instead simply exposed functionality required for your build.

sayandip18:
Thanks @leej3, this looks good!

Issue #532, opened 2021-12-10
Support for numpy 1.21.3


plooney:
Numba doesn't support numpy 1.21 yest so perhaps the requirements need updating to restrict numpy to < 1.21

hameerabbasi:
Please report this to Numba, since this is an upstream dependency issue, it should in theory be fixed upstream.

hameerabbasi:
I'll keep it open to track it. (Pressed the wrong button, I can't computer in the mornings.)

hameerabbasi:
NumPy 1.23.5 is now supported.

PR #533, opened 2021-12-16
fixing the issue with numba requiring numpy version 1.20 or less https://github.com/pydata/sparse/issues/532

Atish03:


hameerabbasi:
I wouldn't do this change here, but in Numba instead, since it's an upstream bug, it requires an upstream fix instead of a workaround.

Atish03:
> I wouldn't do this change here, but in Numba instead, since it's an upstream bug, it requires an upstream fix instead of a workaround.

Thank you for enlightening me, i am a newbie so i dont know much stuff

hameerabbasi:
No problem, thank you for the effort!

Issue #539, opened 2022-03-11
Additional Random Generator


smldub:
Problem:
The sparse.random function is very speedy for most tensors below the size of around 1E9 elements, but after that point the two current methods run into some drawbacks:
1) random_state.choice requires enough memory to store an array of the same size as the number of elements in the tensor, and in my experience this takes quite some time for tensors larger than 1E9
2) The set/hashing method used when the density is below .3 is limited by being a python for loop and takes quite some time to generate any more than 1E6 random elements (which is below the .3 threshold of a 1E9 element tensor). I think this is likely the result of needing to refer to a hash table every loop.

Potential Solution:
1). Generate a list using np.random.randint that is nnz (the number of nonzero entries desired) long.
2). Then use np.unique to both sort and take out any repeats.
3). go back to step 1, but decrease the desired length to number of nnz elements still needed and np.hstack the resulting matrices.

np. unique also sorts the results making the construction of the COO matrix a little faster.

Example Code:
```
def rand (shape, nnz = None, density = .1):
    elements  = np.prod(shape)
    if nnz == None:
        nnz = int(elements*density)
    if nnz > elements: #my sad attempt at preventing errors
        print("fail")
        return None
    out = np.random.randint(elements, size = nnz) #generate the initial guess for indicies
    out = np.unique(out) #remove the repeated indices
    nnztemp = len(out)
    while nnztemp<nnz: #loop to get the rest of the elements
        out = np.hstack((out,np.random.randint(elements,size=int(nnz-nnztemp))
        out = np.unique(out)
        nnztemp = len(out)
    out = np.array(np.unravel_index(out,shape), dtype = np.int64) #converts index from 1d to Nd
    return sparse.COO(out, data = np.random.rand(out.shape[1]), shape = shape)
```
Here is a plot for a tensor with size 1E9 (100,100,100,1000) where I compare the speed of generating a random matrix with the default sparse function (in orange) vs the above function (in blue).

![Figure_1](https://user-images.githubusercontent.com/71188551/157924966-13da8916-effe-4537-a8d2-33339f2a8b0b.png)

The new method also works relatively well at the small limit as well for the particularly large tensors because the first guess often has no repeated values. 

I would be interested to hear any feedback on the idea before I try to implement something.

hameerabbasi:
> 1. random_state.choice requires enough memory to store an array of the same size as the number of elements in the tensor, and in my experience this takes quite some time for tensors larger than 1E9

[...]

PR #540, opened 2022-03-13
Rand

smldub:
Arising from #539 
Added the random number generation algorithms from this paper:
https://www.cs.emory.edu/~cheung/Courses/584/Syllabus/papers/RandomSampling/1984-Vitter-Faster-random-sampling.pdf
Seems to run fine if the new functions are not compiled with numba, but that step introduces quite a few new errors.

### DIFF
--- a/sparse/_utils.py
+++ b/sparse/_utils.py
@@ -254,7 +254,9 @@ def random(
         data_rvs = random_state.rand
 
     # Use the algorithm from python's random.sample for k < mn/3.
-    if nnz == 1:
+    if nnz == elements or density >= 1:
+        ind = np.arange(elements)
+    elif nnz < 2:
         ind = random_state.choice(elements, nnz)
     elif nnz > elements / 2:
         nnztemp = elements - nnz
--- a/sparse/_utils.py
+++ b/sparse/_utils.py
@@ -5,6 +5,7 @@ from functools import reduce
 
 import operator
 import numpy as np
+import numba
 
 
 def assert_eq(x, y, check_nnz=True, compare_dtype=True, **kwargs):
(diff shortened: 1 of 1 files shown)


smldub:
I believe the main problem arises from Numba not supporting RandomState. It does however support np.random.seed, so I think the work around is feed a random number from random_state into functions as the seed.

hameerabbasi:
Can you post an empirical CDF sampled with the new function, just for comparison?

smldub:
Screen Shot 2022-03-14 at 10 27 42 AM

Code based off newest commit:
```
import numpy as np
import sparse
import matplotlib.pyplot as plt
fig = plt.figure()
ax1 = fig.add_subplot(121)
x2 = fig.add_subplot(122)
N = int(1E4)
for i in range(1,30,4):
    a=sparse.random((N),i/100).coords.reshape(-1)
    ax1.plot(a, np.linspace(0, 1, len(a), endpoint=False))
    a = np.random.choice(N,int(N*i/100),replace = False)
    ax2.plot(np.sort(a), np.linspace(0, 1, len(a), endpoint=False))

ax1.legend([('dens=',i/100) for i in range(1,30,4)])
ax2.legend([('dens=',i/100) for i in range(1,30,4)])
ax1.set_title("New Function")
ax2.set_title("np.random.choice")
ax1.plot(np.linspace(0,N,N),np.linspace(0,1,N), 'k--')
ax2.plot(np.linspace(0,N,N),np.linspace(0,1,N), 'k--')
plt.show()
```

hameerabbasi:
Thanks! Always good to check if there are any major numerical inaccuracies. Would you happen to know the numerical stability of the algorithm and how it affects the resulting distribution, just out of curiosity? If not, the plot should suffice.

hameerabbasi:
Also, let me know when you're done making changes, and I'll merge.

hameerabbasi (approved):
LGTM!

hameerabbasi:
Okay! If you're willing, you could post to the SciPy Developers mailing list and request a review.

I'm not particularly concerned about breaking random stability guarantees for the same seed, just about the distribution being okay.

That said, if one considers the cardinality of all sequences that can be sampled, and the size of the seed (which still exists internally even if we pass nothing) some sparse matrices were ALWAYS going to be avoided.

hameerabbasi:
One potential problem I spot right away is that Numba doesn't pass the seed back to the code, and `data_rvs` could be called with a stale state (the two random bit streams could be correlated).

It might make sense to return the RNG state from each algorithm manually and seed the `RandomState` object with that.

smldub:
If you just want the seed, you don't need to pass it through any of the Numba functions, it is generated outside of them. Also numba doesn't support np.random.get_state. 
I might also not be understanding the issue, but when I run np.random.get_state before calling the functions, I get a different state (in the integer position) than after I call algD or algA.

smldub:
Yea unfortunately Numba doesn't support np.random.get_state, but the numba functions aren't calling the same seed that makes the random state. The seed for those functions is 
seed = random_state.choice(np.iinfo(np.int32).max),
so the likelihood (from my pretty bad understanding of statistics) that the states will be the same is 1/2^64 (assuming the first seed is just randomly generated). Also numba functions don't advance the counter for random_state (global or local) or change the global np.random.seed. 
I am going to post one more (hopefully) commit changing some comments, and if you are fine with how it looks/my above argument. Feel free to merge it.

sparse/_utils.py
```diff
@@ -168,19 +290,31 @@ def random(
...
+        ind = random_state.choice(elements, nnz)
+    # Faster to find non-sampled indices and remove them for dens > .5
+    elif elements - nnz < 2:
+        ind = reverse(random_state.choice(elements, elements - nnz), elements)
+    elif nnz > elements / 2:
+        nnztemp = elements - nnz
+        # Using algorithm A for dens > .1
+        if elements > 10 * nnztemp:
+            ind = reverse(
+                algD(nnztemp, elements, random_state.choice(np.iinfo(np.int32).max)),
```
hameerabbasi, line 305: Have you tried changing these to `int64`?

smldub: That was my original implementation before commit  #2bdd17a, but np.choice can only take 32 bit integers for the seed

hameerabbasi:
Thank-you for your excellent work on this, @smldub! You're a fast learner. 😄

PR #551, opened 2022-06-01
Parallelizing reverse function

smldub:
Rewrote the function that finds unsampled indices for matrices with a density > .5 to be faster (if run without parallel flag), but also parallel (which is faster for large nnz and not appreciably slower for low nnz).

### DIFF
--- a/sparse/_utils.py
+++ b/sparse/_utils.py
@@ -87,13 +87,15 @@ def algD(n, N, random_state=None):
         N = size of system (elements)
         random_state = seed for random number generation
     """
-
+    _rnd = np.random.rand
+    _exp = np.exp
+    _log = np.log
     if random_state != None:
         np.random.seed(random_state)
     n = np.int64(n + 1)
     N = np.int64(N)
     qu1 = N - n + 1
-    Vprime = np.exp(np.log(np.random.rand()) / n)
+    Vprime = _exp(_log(_rnd()) / n)
     i = 0
     arr = np.zeros(n - 1, dtype=np.int64)
     arr[-1] = -1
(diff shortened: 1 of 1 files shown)


smldub:
I think I am breaking more than I am fixing trying to correct this sphinx issue, so maybe someone with more experience can handle this part?

hameerabbasi:
Hello, we don't have a precedent for parallelization in this library. In addition, I'm concerned about the reproducibility of this algorithm as well as a potential problem with the same seed being used across threads.

I would suggest keeping the speed-up but removing the parallilization feature. I would also keep all your other fixes.

smldub:
Heyo,
I might be misunderstanding your concern, but this section of the algorithm doesn't involve any random seeds. It is just finding the elements in a range(0,N) that are not contained in an array. Or in mathy terms, it is finding the complement of the array. Because the array we obtain from the random number algorithm is always ordered, we are able to take advantage of that structure to write parallel code. This code doesn't change how the original random numbers are selected.

hameerabbasi:
Hey, even if you don't internally use seeds they will exist. I would not parallelise at all -- If one needs parallelism they can easily use Dask or Ray.

smldub:
Its an internal function, so the end user would need to have a good understanding of how the random sparse matrices are produced in order to parallelize this specific aspect for dense sparse matrices. But I don't really know much about seed related issues. 

If we aren't aiming for speed, the same function can be written in 3 lines of pure python code and will be only a few seconds slower for ~10**9 nnz elements. This would make things way more readable, less error prone, etc... Would that perhaps be a better path?

hameerabbasi:
I'd say so, yes. 😀

PR #555, opened 2022-08-03
Fix nanmean for complex arrays.

### LINKED ISSUE #553: wrong nanmean with complex64
**Describe the bug**
np.nanmean of sparse array does not match numpy result

**To Reproduce**
```python
import numpy as np
import sparse as sp

array = np.array([0. +0.j, 0.+np.nan * 1j], dtype=np.complex64)
sparray = sp.COO.from_numpy(array)

print("numpy:", np.nanmean(array), "sparse:", np.nanmean(sparray))
```
```
numpy: 0j sparse: (nan+nanj)
```

**Expected behavior**
Match the numpy result

**System**
 - OS and version: [e.g. Windows 10] macOS latest
 - `sparse` version (`sparse.__version__`) 0.13.0
 - NumPy version (`np.__version__`) 1.22.4
 - Numba version (`numba.__version__`) 0.55.2

**Additional context**
Seen when using `hypothesis` to test out array wrapping in xarray (https://github.com/pydata/xarray/pull/4972)

hameerabbasi:
Fixes #553

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -81,7 +81,7 @@ release = __version__
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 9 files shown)


keewis:
it seems `complex128` has the same issue, so it might be good to check that as well?

hameerabbasi:
> it seems `complex128` has the same issue, so it might be good to check that as well?

Thanks, done in https://github.com/pydata/sparse/pull/555/commits/4cdf8c0cbeb81eeb48bb01fe4963312875c52ba6.

Issue #554, opened 2022-08-03
SVG logo


headtr1ck:
It would be nice to have an SVG version of the sparse logo.

While at it:
- It seems that the lines are not really parallel, not sure if that was intentional?

In case you are up for a complete rework:
- The aspect ratio is a bit on the too tall side, maybe the lower 2D block can be made horizontal (doesn't help much though)?
- It does not really represent sparse arrays, maybe several data-patches should be missing?

hameerabbasi:
Hello, and thanks for the issue.

Unfortunately I do not have the original, I received it in high-res PNG format.

I would be willing to accept a contribution, but my knowledge of graphics design is limited.

headtr1ck:
Just noticed that the logo is looking quite similar to the [@pydata](https://github.com/pydata) logo :)

jamestwebber:
how's this?
![pydata-sparse-logo](https://user-images.githubusercontent.com/963647/184554571-f488b62e-6d89-4d76-801e-65cd78b7e7b4.svg)

headtr1ck:
Looks great!

If it would be up to me, I would drop the top and bottom 2D parts altogether.
Would make the logo more square.

Maybe some core devs / pydata people have an opinion about this?

(Just opened the image standalone in chrome and it was displayed all black, not sure if that's a chrome bug? In the GitHub post it renders correctly...)

jamestwebber:
> (Just opened the image standalone in chrome and it was displayed all black, not sure if that's a chrome bug? In the GitHub post it renders correctly...)

Yeah I saw that too, dunno why. When I open the image in Chrome directly (from the finder) it looks okay. I think GitHub must have done some automatic processing or something?

> If it would be up to me, I would drop the top and bottom 2D parts altogether.
Would make the logo more square.

> Maybe some core devs / pydata people have an opinion about this?

I don't have a strong opinion about this. I assume it should basically match the pydata main logo, and a cube would look pretty boring there.

headtr1ck:
> Just opened the image standalone in chrome and it was displayed all black, not sure if that's a chrome bug? In the GitHub post it renders correctly...

When I use inspect and disable a fill style and enable it again it works... Really seems like a chrome bug to me.

jamestwebber:
> I don't have a strong opinion about this. I assume it should basically match the pydata main logo, and a cube would look pretty boring there.

Just realized the PyData logo looks the way it does because it's a stylized "pd" 😂

[...]

PR #556, opened 2022-08-16
added svg logo and logo-generating script

jamestwebber:
A PR to address #554. I may have gone a little bit crazy with it 😂 it was fun though

I wrote a script to generate an SVG logo with some random opacity, and modified the docs so that it looks right. That may seem crazy but the motivations were a) it was fun to do and b) there might be a lot of bike-shedding on this, and I'm not good enough with a Affinity Designer to make changes to an image quickly.

One question I had to decide: in the docs, the logo is shown on two different backgrounds. If I just used opacity then it looks very different, so I added white background squares. But this also looks a little odd on the blue background in the sidebar. The original logo didn't actually use opacity so it wasn't an issue there.


hameerabbasi:
> I wrote a script to generate an SVG logo with some random opacity, and modified the docs so that it looks right. That may seem crazy but the motivations were a) it was fun to do and b) there might be a lot of bike-shedding on this, and I'm not good enough with a Affinity Designer to make changes to an image quickly.

No bike shedding from my side -- I'd just fix the seed for reproducibility but other than that, we're good.

The test failures should be resolved when #555 is in.

jamestwebber:
> No bike shedding from my side -- I'd just fix the seed for reproducibility but other than that, we're good.

Seed is set at 1 (0 didn't look random enough).

hameerabbasi:
Could you merge/rebase `master` and we should be good to go.

headtr1ck:
Currently on mobile so cannot really check much, but could you ensure that it also looks nice in dark mode?

hameerabbasi:
I don't think we have a dark mode. It doesn't seem to come vanilla.

jamestwebber:
Yeah when I view the docs after building locally I don't see an option for a dark mode. But I set the SVG background to black and took a screenshot::

hameerabbasi:
My aesthetic sense is close to none, I'll let you be the judge of how that'll blend in.

headtr1ck:
Yeah, maybe removing the white parts again and instead of using semi-transparent tiles, simply tiles with a lighter color?

jamestwebber:
> Yeah, maybe removing the white parts again and instead of using semi-transparent tiles, simply tiles with a lighter color?

This does look pretty sweet on a dark background, but IMO it looks worse on the blue background of the docs page sidebar. Given that there isn't _actually_ a dark mode for the docs, I'm inclined to keep it with the white background.

Screen Shot 2022-08-17 at 18 58 36
Screen Shot 2022-08-17 at 18 58 50

headtr1ck:
Maybe having two versions is also ok?
Alternatively, you can use CSS to change the color of the background to/from transparent to white?

headtr1ck:
I have noticed that you removed the spacing between the sides.
Maybe it was too large in the initial version but I think no spacing is confusing, at least for the cube-2D intersection.

jamestwebber:
Okay this is why I made the script 😂 I don't really have time to make additional tweaks but happy for you to play with it more.

hameerabbasi:
Merging as-is for now... -- Thanks, @jamestwebber!-

Issue #31, opened 2017-08-31
einsum implementation


aremirata:
Hi guys, do you have plans to add einsum to this library? I guess einsum from numpy  is obtained from c. There are einsum operation I cannot directly convert to using tensordot. by the way, this library is really awesome and thanks for creating this.

mrocklin:
I have no plans personally to add einsum, but it seems well within scope if
you're interested in implementing it.

bigerl:
I realized that what I really need for my project is a highly parallelized einsum. I'm implementing it now as stand-alone and when I'm done I'll check out if it is a reasonable effort to integrate it with this library again. When it's done it will be open sourced anyways. I'll report back when I've got something new.

hameerabbasi:
@aremirata @bigerl I've found in my experience that a lot of the stuff that's done using `einsum` is doable using broadcasting, reductions and `tensordot`. Maybe you can try those now that they're added?

nils-werner:
I've written a very simple proof of concept: https://github.com/nils-werner/sparse/tree/einsum

It only works for two operands and doesn't support any of the fancy einsum tricks like ellipses etc.

Internally it uses tensordot and the fact that we can cheaply diagonalize axes on sparse arrays (see the docstrings for more explanation).

nils-werner:
I've written a few small issues and added [some tests](https://github.com/nils-werner/sparse/blob/einsum/sparse/tests/test_core.py#L821-L842).

hameerabbasi:
@nils-werner It'd be great to have this in, even with partial support. More features can always be added. We can document the missing features.

One minor thing: I'd really prefer to have this support any number of arrays rather than two.

stsievert:
I'm using [opt_einsum] with their sparse backend, but opt_einsum still wants an einsum operation in sparse. It works for particular einsum strings as per [their sparse tests][1], but still relies an einsum operation for higher dimensions/more inputs.

The einsum strings that are throwing errors are `'bz,az->abz'`, `'cz,bz,az->abcz'` and `'dz,cz,bz,az->abcdz'` (and so on; always N two indices to one length N+1 index).


[1]:https://github.com/dgasmith/opt_einsum/blob/960cf7c45993077be6d4ad04a1564b6042f0e092/opt_einsum/tests/test_backends.py#L131-L158
[opt_einsum]:https://optimized-einsum.readthedocs.io/en/latest/

hameerabbasi:
[...]

PR #564, opened 2022-11-17
einsum 

jcmgray:
This is a general einsum implementation (#31) for the COO format. An implementation was provided in #31 but it doesn't seem to work for all inputs. The implementation here passes all of `opt_einsum`'s test suite edges cases, and is also reasonably performant (e.g. same order of magnitude speed as `tensordot` for cases where they overlap in functionality, sometimes faster). But certainly not the most optimized one could do. One would also probably want to use `opt_einsum` for more than 2 terms generally, but it's quite useful to have a reference, non-optimized form.

### Maybe later:
- [ ] dispatch to tensordot when compatible?
- [ ] benchmarks

### DONE
- [x] shape checks
- [x] main docs?
- [x] main tests
- [x] input checks (such as default fill value?)
- [x] decide what to do with mixed input types?
- [x] edge case tests such as mixed input types?
- [x] docstrings
- [x] decide what to do in the scalar output case: return scalar or sparse? : numpy returns scalar, this is generally convenient


hameerabbasi:
Thanks for the PR! It's much appreciated. However, I won't have much time to review before the weekend.

Assuming you're willing, please go ahead and work on it. Based on your comments in #31 and here, this is a very welcome change.

sparse/_common.py
```diff
@@ -1211,6 +1212,142 @@ def _dot_ndarray_coo(array1, coords2, data2, out_shape):  # pragma: no cover
...
+        The array to perform the einsum on.
+
+    Returns
+    -------
+    output : SparseArray
+    """
+    operand = asCOO(operand)
+
+    if lhs == rhs:
+        # XXX: take copy?
```
hameerabbasi, line 1235: No need to take a copy -- `SparseArrays` are assumed immutable.

sparse/tests/test_einsum.py
```diff
@@ -0,0 +1,95 @@
...
+    d = 4
+
+    terms = subscripts.split("->")[0].split(",")
+    arrays = [sparse.random((d,) * len(term), density=density) for term in terms]
+    sparse_out = sparse.einsum(subscripts, *arrays)
+    numpy_out = np.einsum(subscripts, *(s.todense() for s in arrays))
+
+    if hasattr(sparse_out, "todense"):
+        # scalar output is not a sparse array
+        sparse_out = sparse_out.todense()
```
hameerabbasi, line 93:
```suggestion
    if hasattr(sparse_out, "todense"):
        # scalar output is not a sparse array
        sparse_out = sparse_out.todense()
```

This should be done inside the main `einsum` function -- The convention is to return `SparseArray` instead of `ndarray` (including 0D), but a scalar for a scalar.

sparse/_common.py
```diff
@@ -1211,6 +1212,142 @@ def _dot_ndarray_coo(array1, coords2, data2, out_shape):  # pragma: no cover
...
+        new_data = operand.data[selector]
+    else:
+        new_coords = operand.coords[perm]
+        new_data = operand.data
+
+    if not rhs:
+        # scalar output - match numpy behaviour by not wrapping as array
+        return new_data.sum()
+
+    return operand.__class__(new_coords, new_data, shape=new_shape, has_duplicates=True)
```
hameerabbasi, line 1272: Why not use `COO` directly?

hameerabbasi (changes requested):
A couple of suggested changes -- The base logic looks good.

hameerabbasi:
> * main docs?

Yes, documentation would be useful, if only a section in the usage page.



> * input checks (such as default fill value?)

Shape checks and zero fill value are a must.



> * decide what to do with mixed input types?

General convention is to return `COO` if mixed, but the same type if others.



> * edge case tests such as mixed input types?

Yes, please. I've noticed bugs sometimes pop up in this area.

sparse/_common.py
jcmgray: It was just to avoid the circular import error of putting `from .coo import COO` in `_common.py`,but suggestions welcome.

sparse/tests/test_einsum.py
jcmgray: That convention should already happen - possibly could be clearer, but this line just converts non-scalar (i.e. sparse) arrays to dense for the test comparison.

jcmgray:
> Yes, documentation would be useful, if only a section in the usage page.

I've mentioned it alongside tensordot, but let me know if there should more extensive docs. I'm partly assuming the implicit "it's numpy.einsum but for sparse arrays" is enough.

> Shape checks and zero fill value are a must.
> General convention is to return COO if mixed, but the same type if others.
> Yes, please. I've noticed bugs sometimes pop up in this area.

I've added various input and shape checks. The logic of what format to output is already largely handled by the `multiply(*prepared_operands)`. The only non-trivial thing is that `DOK` didn't support `reshape`, and it was cleaner to add it rather than add logic about converting to COO and back etc. But maybe `DOK.reshape` was intentionally not implemented?

sparse/_common.py
hameerabbasi: Well, operand may not be `COO`. I'd instead import inside the function.

hameerabbasi:
> But maybe `DOK.reshape` was intentionally not implemented?

Definitely useful to have this method, so long as it's tested.

jcmgray:
@hameerabbasi, I've added various `DOK.reshape` tests. From my end I think this is good to go now.

hameerabbasi:
Waiting on CI to turn green.

hameerabbasi (approved):

hameerabbasi:
Just formatting errors from what I can see, run black. 😅

jcmgray:
Aha a sneaky single quote, done.

hameerabbasi:
Ah, still a bunch of black fixes in other files. 😅 I'll made a quick fix PR soon and then re-run the CI on and merge yours.

jcmgray:
I see, I did run `black sparse/` and it didn't change anything else, maybe there is a particular config it needs to run with?

hameerabbasi:
Just run `black .` with the latest black version.

hameerabbasi:
Thanks, @jcmgray!

jcmgray:
Nice, thanks for reviewing!

PR #565, opened 2022-11-19
Relative imports (helps vendoring)

mangecoeur:
Use relative imports, apart from being a good code style it makes vendoring sparse within another package easier.


hameerabbasi:
Thanks, @mangecoeur!

Issue #566, opened 2022-11-19
fill_value is implicitly assumed to be a subscriptable numpy type


mangecoeur:
`fill_value` is described as a scalar but in is assumed to be a subscriptable numpy type in`_Elemwise._get_fill_value`, leads to errors if you supply your own fill value.

e.g.
```python
        zero_args = tuple(
            arg.fill_value[...] if isinstance(arg, COO) else arg
            for arg in self.args
        )
```

fails with a normal float fill_value

DeepakSingh260:
Can we raise a typeerror exception here . and it raise put the varible inside a list like
try: 
   zero_args = tuple(
              arg.fill_value[...] if isinstance(arg, COO) else arg
              for arg in self.args
          )
except TypeError:
     arg.fillvalue = [arg.fillvalue]
   zero_args = tuple(
              arg.fill_value[...] if isinstance(arg, COO) else arg
              for arg in self.args
          )

hameerabbasi:
I would suggest making `self.fill_value` a read-only property, that can only be set in the ctor.

hameerabbasi:
I feel this is unsupported utilization of an internal API based on the comments. Closing as wontfix.

Issue #297, opened 2019-10-29
Boolean operation which flips my fill value


jwdebelius:
Hi!

This is an awesome package and I'm having fun using it to its full potential. Hopefully this is the right place for this. I'm having an issue where, after a series of boolean operations, my fill value becomes True and my data value becomes False. This limits my ability to do further calculations.

I have a sparse, boolean `n` x `n` x `m` array. 

```python
>>> import sparse as sp
>>> m = sp.COO(
...    coords=np.array([[0, 0, 0], [0, 0, 1],
...                     [0, 1, 0],
...                     [1, 0, 0],
...                     [1, 1, 0], [1, 1, 1],
...                     [2, 2, 1], 
...                     [3, 3, 0], [3, 3, 1]
...                     ]).T,
...    data=1,
...    shape=(4, 4, 2),
... )
>>> m
<COO: shape=(4, 4, 2), dtype=int64, nnz=9, fill_value=0>
```

I want to identify rows/columns where there is not any value that is present in all three coordinates by checking for the number of positions that have something present and then that this is less than the number of regions.

```python
>>> pres = (m.sum(axis=1) > 0).sum(axis=1) < 2
>>> pres
<COO: shape=(4,), dtype=bool, nnz=3, fill_value=True>
>>> pres.todense()
array([False, False,  True, False])
```

If I try to do something like flip my fill value, then I just get a matrix of zeros

```python
>>> pres.fill_value = 0
>>> pres
<COO: shape=(4,), dtype=bool, nnz=3, fill_value=0>
>>> pres.todense()
array([False, False,  False, False])
```

But, the fact that my array density flips seems like  a possible bug, or at least a parameter I'd like more control over.

hameerabbasi:
Your second code snippet illustrates the issue.

> ```python
> >>> pres = (m.sum(axis=1) > 0).sum(axis=1) < 2
> >>> pres
> <COO: shape=(4,), dtype=bool, nnz=3, fill_value=True>
> >>> pres.todense()
> array([False, False,  True, False])
> ```

If there's a row of all-zeros, it'll sum to zero, so the resulting fill-value is zero. When you compare an array with a fill-value of zero to 2, it's going to be less most of the time.

This is the reason that we have a fill-value of `True` in the resulting array. However, you make a point that one should be able to manipulate the fill-value in the manner you show.

jwdebelius:
[...]

Issue #287, opened 2019-09-09
Sparse removing dimensions


TomMaullin:
Apologies: Error in my coding. This was a non-issue.

PR #567, opened 2022-11-19
Explicitly wrap fill_value #566

mangecoeur:
Explicitly wrap fill_value in array instead of assuming a numpy dtype, should address #566

### DIFF
--- a/sparse/_coo/core.py
+++ b/sparse/_coo/core.py
@@ -257,7 +257,7 @@ class COO(SparseArray, NDArrayOperatorsMixin):  # lgtm [py/missing-equals]
         if isinstance(shape, np.ndarray):
             shape = tuple(shape)
 
-        if shape is not None and not self.coords.size:
+        if shape and not self.coords.size:
             warnings.warn(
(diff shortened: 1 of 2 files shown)


hameerabbasi:
Could we add a test for some functionality this would enable?

mangecoeur:
Sure, something in my code triggered this, i'll try to pull out a test case.

mangecoeur:
Ok so checking my code I realized what was triggering my issue was changing the `fill_value` after array creation, which i realise looking at #297 is not yet supported. Now I'm not sure if this change would be part of a full solution to #297 or if instead solving #287 would make this change obsolete. 

The error can be reproduced thus:
```
    s = sparse.random((2, 3, 4), density=0.5, format=format)
    s.fill_value = np.nan
    fs = s *2
```

But I don't see where in the code anything is done to `fill_value` which makes it subscriptable in the case its passed to the constructor and not subscriptable if modified after class creations.

hameerabbasi:
Yes, mutating a sparse array isn't supported, unfortunately.

hameerabbasi:
Closing as explicitly setting the `fill_value` isn't supported.

PR #568, opened 2022-11-19
Return scalar if zero dimension select

mangecoeur:
According to numpy docs, 0-dimension indexing on 0-dimension array `x[()]` should return a scalar. Since a 0d sparse array will only have one element we can safely densify and return scalar value.


hameerabbasi:
A small test would be appreciated.

rgommers:
I recommend against this change. This was a design mistake in NumPy, and other NumPy-compatible libraries like CuPy and JAX all don't follow it (they don't even have library-specific scalars). Returning 0-D arrays is better. That's also what the array API standard does.

mangecoeur:
Hm ok, I actually added this because I wanted to get the scalar value out of a 0d array and stackoverflow claimed using `x[()]` was the correct way.  What is instead the right way to get a scalar from a 0d sparse array?

hameerabbasi:
`s[()]` is the correct way. Densification should return an array, even if 0D.

hameerabbasi:
Ah, I missed https://github.com/pydata/sparse/pull/568#issuecomment-1321044063. Well, in this library we tend to use NumPy scalars -- But it sounds like the ecosystem is moving in the other direction, so I'd agree with @rgommers.

mangecoeur:
Righto, but then what is the canonical way of getting a scalar out of a single element array?

hameerabbasi:
You can do `.todense()` -- Scalars shouldn't exist, only 0D arrays should.

hameerabbasi:
Closing based on https://github.com/pydata/sparse/pull/568#issuecomment-1321044063.

PR #569, opened 2022-11-21
fix errors/crashes in multi-dimensional slicing

### LINKED ISSUE #550: GCXS slice bug
**Describe the bug**
The slice operator (`__getitem__`) produces incorrect results in some cases for the class `GCXS`.

**To Reproduce**
I found this error working with a large matrix. This is the minimal version that reproduces the error

```
dok = DOK(shape = (12, 1147, 12, 32, 1147, 3, 3))
dok[1,1,1,1,1,1,1] = 5
coo = dok.to_coo()
gcxs = coo.asformat('gcxs')

print( coo[1:11, 1:2, 1:11, 1:31, 1:1146, 1:2, 1:2])
print(gcxs[1:11, 1:2, 1:11, 1:31, 1:1146, 1:2, 1:2])
```

Results in 
```
<COO: shape=(10, 1, 10, 30, 1145, 1, 1), dtype=float64, nnz=1, fill_value=0.0>
<GCXS: shape=(10, 1, 10, 30, 1145, 1, 1), dtype=float64, nnz=0, fill_value=0.0, compressed_axes=(5,)>
```
 They should be the same.

I tried to reproduce it on small arrays, but it worked just fine. On medium-size matrices however, like the following, sometimes it works and sometimes it doesn't (running exactly the same code, seems random):

```
dok = DOK(shape = (4,5,6,7,8,9,10))
dok[1,1,1,1,1,1,1] = 5
coo = dok.to_coo()
gcxs = coo.asformat('gcxs')

print( coo[1:3, 1:2, 1:5, 1:6, 1:7, 1:8, 1:9])
print(gcxs[1:3, 1:2, 1:5, 1:6, 1:7, 1:8, 1:9])
```

Sometimes results in :
```
<COO: shape=(2, 1, 4, 5, 6, 7, 8), dtype=float64, nnz=1, fill_value=0.0>
<GCXS: shape=(2, 1, 4, 5, 6, 7, 8), dtype=float64, nnz=1, fill_value=0.0, compressed_axes=(0,)>
```

And if I keep executing the cell, sometimes it outputs:
```
<COO: shape=(2, 1, 4, 5, 6, 7, 8), dtype=float64, nnz=1, fill_value=0.0>
<GCXS: shape=(2, 1, 4, 5, 6, 7, 8), dtype=float64, nnz=0, fill_value=0.0, compressed_axes=(0,)>
```

**Expected behavior**
COO and GCXS should have consistent results. In particular for the examples above, the result should have `nnz=1` always (COO gets it right).

**System**
 - OS and version: macOS 12.3.1 
 - `sparse` version (`sparse.__version__`): 0.13.0
 - NumPy version (`np.__version__`): 1.21.5
 - Numba version (`numba.__version__`): 0.53.0

**Additional context**
None

jamestwebber:
Possibly closes #550. Needs some review as I did not fully grok what `compute_flat` was doing here, but it seemed like it was failing to account for multi-dimensional sparse arrays properly.

There might be a better way to do this (or a better way to calculate this more generally, I'm not sure) but this seems to have fixed the problem identified in #550 as well as some crashes I observed in particularly bad situations.

### DIFF
--- a/sparse/_compressed/convert.py
+++ b/sparse/_compressed/convert.py
@@ -35,11 +35,11 @@ def compute_flat(increments, cols, operations):  # pragma: no cover
     positions = np.zeros(len(increments) - 1, dtype=np.intp)
     pos = len(increments) - 2
     for i in range(operations):
-        if i != 0 and positions[pos] == increments[pos].shape[0]:
+        while i != 0 and positions[pos] == increments[pos].shape[0]:
             positions[pos] = 0
             pos -= 1
             positions[pos] += 1
-            pos += 1
+        pos = len(increments) - 2
         to_add = np.array(
             [increments[i][positions[i]] for i in range(len(increments) - 1)]
         ).sum()
--- a/sparse/tests/test_compressed_convert.py
+++ b/sparse/tests/test_compressed_convert.py
@@ -57,4 +57,6 @@ def test_compute_flat(shape, expected_subsample, subsample):
     ],
 )
 def test_transform_shape(shape, expected_shape):
-    assert_eq(convert.transform_shape(np.asarray(shape)), expected_shape)
+    assert_eq(
+        convert.transform_shape(np.asarray(shape)), expected_shape, compare_dtype=False
+    )
--- a/sparse/tests/test_compressed_convert.py
+++ b/sparse/tests/test_compressed_convert.py
@@ -29,14 +29,13 @@ def make_increments(shape):
 )
 def test_compute_flat(shape, expected_subsample, subsample):
     increments = make_increments(shape)
-    operations = np.prod(
-        [inc.shape[0] for inc in increments[:-1]], dtype=increments[0].dtype
-    )
-    cols = np.empty(increments[-1].size * operations, dtype=increments[0].dtype)
+    dtype = increments[0].dtype
+    operations = np.prod([inc.shape[0] for inc in increments[:-1]], dtype=dtype)
+    cols = np.empty(increments[-1].size * operations, dtype=dtype)
 
     assert_eq(
         convert.compute_flat(increments, cols, operations)[::subsample],
-        expected_subsample,
+        expected_subsample.astype(dtype),
     )
 
 
--- a/sparse/tests/test_compressed_convert.py
+++ b/sparse/tests/test_compressed_convert.py
@@ -1,0 +1,39 @@
+import sparse
+import pytest
+import numpy as np
+
+import sparse._compressed.convert as convert
+from sparse._utils import assert_eq
+
+
+# @pytest.mark.parametrize(
+# )
+def test_convert_to_flat():
+    ...
+
+
+# @pytest.mark.parametrize(
+# )
+def test_compute_flat():
+    ...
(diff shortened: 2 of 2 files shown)


hameerabbasi:
I'd really love to avoid silent bad results. I realise this fixes crashes, but can we possibly add tests?

jamestwebber:
> I'd really love to avoid silent bad results. I realise this fixes crashes, but can we possibly add tests?

I feel like I don't fully understand what this code is doing, yet&mdash;I just worked backwards from where I saw an exception.

Writing a test seems tricky because the problem is sporadic (I'm not even sure why!). But once I understand what's going on in `convert.py` I will try to write tests for it.

hameerabbasi:
How about testing with Numba environment variables that enable bounds checking? Or disable JIT? That should help debug.

jamestwebber:
> How about testing with Numba environment variables that enable bounds checking? Or disable JIT? That should help debug.

For me the issue wasn't finding the bug (I just removed the `jit` and it's a normal exception), I'm just still trying to understand GCXS so I know what the behavior should be. That way I can write a test for regression.

hameerabbasi:
> For me the issue wasn't finding the bug (I just removed the jit and it's a normal exception),

For context, Numba May have buffer overflows on out of range accesses of an array. These sometimes result in segfaults and crashed processes, but sometimes they just overwrite other non-important memory.

But in non-jit code, these show up as `IndexError`s. So if you fix the underlying exceptions and the tests pass with disabled JIT, I think that should be sufficient.

We already have tests in CI that test without JIT for this reason.

jamestwebber:
Understood, I just want to verify that I have test cases that failed in the current code. The slightly-random nature of the bug makes it a little trickier but I think the error is more consistent at lower levels.

> We already have tests in CI that test without JIT for this reason.

Where are these? It'd be useful to see how this is done.

jamestwebber:
I added some test cases that all failed with the previous code (at least sometimes?) but I believe are now correct.

I am verifying against a small subsample of the output because they are all very large arrays. But it should be consistent and hopefully would detect any regressions here.

As a side note, while messing with this code I found I could speed it up a fair amount (e.g. removing unnecessary array allocation in jitted function), but I'll make a new PR for that.

jamestwebber:
For posterity, I think the reason this bug was so tricky was because it requires a high-dimensional fairly large array to get tripped.

High-dimensional because it seems to require multiple _uncompressed_ axes and furthermore the number of `operations` must be large enough that it doesn't finish before hitting the overflow issue. I think it requires at least 4 dimensions for a GCXS array, maybe 5.

hameerabbasi:
> Where are these? It'd be useful to see how this is done.

https://github.com/pydata/sparse/blob/d51a1a10bff5ca93a8c8222449fc117e887e772f/.github/workflows/ci.yml#L18 and https://github.com/pydata/sparse/blob/d51a1a10bff5ca93a8c8222449fc117e887e772f/.github/workflows/ci.yml#L23

hameerabbasi (approved):

hameerabbasi:
Thanks, @jamestwebber.

PR #570, opened 2022-11-22
Improving indexing performance

jamestwebber:
While working on #569*, I realized that there were a few optimizations that could be performed, and with a little additional work I could use `numba.jit` on `_compressed.convert.convert_to_flat` as well.

\* that PR should be merged first, as this builds on it

Short summary of changes:
 * removed an array allocation and let numba optimize a loop in `compute_flat`, which was roughly half the time in that function.
 * similarly, removed an array allocation and the use of `np.prod` and let numba take care of it
 * part of the work of building `cols` can be performed up front with the equivalent of `np.tile` (but that's not a numba-compatible function so this is the equivalent). This required changing my tests a little.
 * simplified `transform_shape` a little bit--no reason to iterate backwards
 * to maintain correctness I had to add a few checks earlier in the process, but I believe all tests pass

<details>
<summary>Code for some rudimentary A/B tests (using some random shapes and simple slices)</summary>

```python
import numpy as np
from numba.typed import List
from sparse._compressed.convert import convert_to_flat

random_shapes = [
    (5, 6, 7, 8, 9),
    (13, 12, 12, 9, 7),
    (12, 15, 7, 14, 9),
    (18, 5, 12, 14, 9, 11, 8, 14),
    (11, 6, 13, 11, 17, 7, 15),
    (9, 9, 12, 7, 12)
]

# call once to compile function
_ = convert_to_flat(
    List([np.arange(1, a - 1, dtype=int) for a in random_shapes[0]]),
    random_shapes[0],
    np.int64
)

# unoptimized version (from #569)
for shape in random_shapes:
    inds = List([np.arange(1, a - 1, dtype=int) for a in shape])
    %timeit convert_to_flat(inds, shape, np.int64)

257 µs ± 12.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
2.81 ms ± 183 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
2.8 ms ± 97.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
994 ms ± 25.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
113 ms ± 3.83 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
1.07 ms ± 6.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

# this version
for shape in random_shapes:
    inds = List([np.arange(1, a - 1, dtype=int) for a in shape])
    %timeit convert_to_flat(inds, shape, np.int64)

75.6 µs ± 1.26 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
1.42 ms ± 92.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
1.38 ms ± 20.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
677 ms ± 14.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
65 ms ± 2.63 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
458 µs ± 16.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

```

</details>

### DIFF
--- a/sparse/_compressed/indexing.py
+++ b/sparse/_compressed/indexing.py
@@ -1,5 +1,6 @@
 import numpy as np
 import numba
+from numba.typed import List
 from numbers import Integral
 from itertools import zip_longest
 from collections.abc import Iterable
@@ -80,10 +81,13 @@ def getitem(x, key):
     # convert all ints and slices to iterables before flattening
     for i, ind in enumerate(reordered_key):
         if isinstance(ind, Integral):
-            reordered_key[i] = [ind]
+            reordered_key[i] = np.array([ind])
         elif isinstance(ind, slice):
             reordered_key[i] = np.arange(ind.start, ind.stop, ind.step)
+        elif isinstance(ind, np.ndarray) and ind.ndim > 1:
+            raise IndexError("Only one-dimensional iterable indices supported.")
 
+    reordered_key = List(reordered_key)
     shape = np.array(shape)
 
     # convert all indices of compressed axes to a single array index
--- a/sparse/_compressed/indexing.py
+++ b/sparse/_compressed/indexing.py
@@ -87,6 +87,8 @@ def getitem(x, key):
         elif isinstance(ind, np.ndarray) and ind.ndim > 1:
             raise IndexError("Only one-dimensional iterable indices supported.")
 
+        reordered_key[i] = reordered_key[i].astype(x.indices.dtype, copy=True)
+
     reordered_key = List(reordered_key)
     shape = np.array(shape)
 
--- a/sparse/_compressed/indexing.py
+++ b/sparse/_compressed/indexing.py
@@ -87,7 +87,7 @@ def getitem(x, key):
         elif isinstance(ind, np.ndarray) and ind.ndim > 1:
             raise IndexError("Only one-dimensional iterable indices supported.")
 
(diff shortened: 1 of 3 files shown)


hameerabbasi:
I guess there's an astype or an explicit dtype missing somewhere, from a quick glance at the CI logs.

jamestwebber:
I realized I could test for `convert_to_flat`, which is probably redundant with the test for `compute_flat`. I don't think it would hurt to remove the latter.

sparse/_compressed/indexing.py
```diff
@@ -87,6 +87,8 @@ def getitem(x, key):
         elif isinstance(ind, np.ndarray) and ind.ndim > 1:
             raise IndexError("Only one-dimensional iterable indices supported.")
 
+        reordered_key[i] = reordered_key[i].astype(x.indices.dtype, copy=True)
```
hameerabbasi, line 90: Why `copy=True` if we won't be mutating this?

hameerabbasi (commented):
Minor comment

sparse/_compressed/indexing.py
jamestwebber: brainfart 😅 the default is `True` but I wanted to set it to `False` for when it was not needed.

PR #573, opened 2023-01-13
Add .format support to GCXS and COO classes

SultanOrazbayev:
Currently, only 'dok' supports `.format`:

```python
import sparse

for format in ["dok", "gcxs", "coo"]:
    s = sparse.random((5, 5), density=0.2, format=format)
    print(s.format)
# dok
# AttributeError: 'GCXS' object has no attribute 'format'
# AttributeError: 'COO' object has no attribute 'format'
```

### DIFF
--- a/sparse/_sparse_array.py
+++ b/sparse/_sparse_array.py
@@ -192,29 +192,6 @@ class SparseArray:
             If the format isn't supported.
         """
 
-    @classmethod
-    def format(cls):
-        """
-        The storage format of this array.
-        Returns
-        -------
-        str
-            The storage format of this array.
-        See Also
-        -------
-        scipy.sparse.dok_matrix.format : The Scipy equivalent property.
-        Examples
-        -------
-        >>> import sparse
-        >>> s = sparse.random((5,5), density=0.2, format='dok')
-        >>> s.format
-        'dok'
-        >>> t = sparse.random((5,5), density=0.2, format='coo')
-        >>> t.format
-        'coo'
-        """
-        return cls.__name__.lower()
-
     @abstractmethod
     def todense(self):
         """
--- a/sparse/_sparse_array.py
+++ b/sparse/_sparse_array.py
@@ -192,8 +192,8 @@ class SparseArray:
             If the format isn't supported.
         """
 
-    @classmethod
     @property
+    @classmethod
     def format(cls):
         """
         The storage format of this array.
--- a/sparse/_sparse_array.py
+++ b/sparse/_sparse_array.py
@@ -192,8 +192,8 @@ class SparseArray:
             If the format isn't supported.
         """
 
-    @property
     @classmethod
+    @property
     def format(cls):
         """
         The storage format of this array.
--- a/sparse/_sparse_array.py
+++ b/sparse/_sparse_array.py
@@ -203,8 +203,6 @@ class SparseArray:
             The storage format of this array.
         See Also
         -------
-        COO.format : Equivalent :obj:`COO` array property.
-        GCXS.format : Equivalent :obj:`GCXS` array property.
         scipy.sparse.dok_matrix.format : The Scipy equivalent property.
         Examples
         -------
--- a/sparse/_sparse_array.py
+++ b/sparse/_sparse_array.py
@@ -192,6 +192,32 @@ class SparseArray:
             If the format isn't supported.
         """
 
+    @classmethod
+    @property
+    def format(cls):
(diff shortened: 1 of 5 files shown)


SultanOrazbayev:
To give context where this might be useful. When doing conversion from `pydata.sparse` to another object, it's useful to know the type of the sparse matrix, see https://github.com/python-graphblas/python-graphblas/pull/347#issuecomment-1381281843. Using `.format` would be consistent with the `scipy.sparse` approach.

hameerabbasi (changes requested):
Can we make this a classmethod on `SparseArray`? Perhaps the implementation could be something like

```python
return cls.__name__.lower()
```

hameerabbasi:
Also some unit tests would be awesome.

SultanOrazbayev:
Thanks, classmethod on `SparseArray` would be great, I wasn't sure how to implement it, but your suggestion looks like a way forward.

hameerabbasi:
All we need now are small unit tests. 😄

sparse/_sparse_array.py
```diff
@@ -192,6 +192,32 @@ def asformat(self, format):
...
+        """
+        The storage format of this array.
+        Returns
+        -------
+        str
+            The storage format of this array.
+        See Also
+        -------
+        COO.format : Equivalent :obj:`COO` array property.
+        GCXS.format : Equivalent :obj:`GCXS` array property.
```
hameerabbasi, line 207:
```suggestion
```

sparse/tests/test_array_function.py
```diff
@@ -92,3 +92,8 @@ def test_zeros_like_order():
 
     assert isinstance(actual, sparse.COO)
     assert_eq(actual, expected)
+
+@pytest.mark.parametrize("format", ["dok", "gcxs", "coo"])
+def test_format(func):
```
hameerabbasi, line 97:
```suggestion
def test_format(format):
```

hameerabbasi:
> Hmm, why code coverage is dropping so much?

Because the tests didn't actually run. Speaking of, we should change it to run only if tests succeed.

sparse/tests/test_array_function.py
```diff
@@ -92,3 +92,8 @@ def test_zeros_like_order():
 
     assert isinstance(actual, sparse.COO)
     assert_eq(actual, expected)
+
+@pytest.mark.parametrize("format", ["dok", "gcxs", "coo"])
+def test_format(format):
+    s = sparse.random((5, 5), density=0.2, format=format)
+    assert s.format == format
```
hameerabbasi, line 99:
```suggestion


@pytest.mark.parametrize("format", ["dok", "gcxs", "coo"])
def test_format(format):
    s = sparse.random((5, 5), density=0.2, format=format)
    assert s.format == format
```

sparse/_sparse_array.py
```diff
@@ -192,6 +192,30 @@ def asformat(self, format):
             If the format isn't supported.
         """
 
+    @classmethod
+    @property
```
hameerabbasi, line 196:
```suggestion
    @property
    @classmethod
```

sparse/_sparse_array.py
```diff
@@ -192,6 +192,30 @@ def asformat(self, format):
             If the format isn't supported.
         """
 
+    @property
+    @classmethod
```
hameerabbasi, line 196:
```suggestion
    @classmethod
    @property
```

SultanOrazbayev:
I tested it locally and seems to work. The only unrelated thing is I had to install `hypothesis` manually. (apparently was not installed by `pip install -e .[all]`)

SultanOrazbayev:
Also, the combination of `@property` and `@classmethod` is not supported by 3.11... https://docs.python.org/3.11/library/functions.html?highlight=classmethod#classmethod

hameerabbasi:
I guess let's find another solution, like the one you had before? Sorry for the wild goose chase.

SultanOrazbayev:
Looking through this thread https://github.com/python/cpython/issues/89519, I think this is an example where violating DRY might not be so bad.

SultanOrazbayev:
OK, this version passes changes locally. Hopefully it will work.

hameerabbasi (approved):

hameerabbasi:
This is in, thanks @SultanOrazbayev!

SultanOrazbayev:
Thank you for the help, @hameerabbasi !

PR #576, opened 2023-02-03
Ensure correct NNZ when a zero dimensional array contains a value

mangecoeur:
Currently COO will return NNZ=0 even if there is data in a zero dimensional array, this is a fix for that.

### DIFF
--- a/sparse/_coo/core.py
+++ b/sparse/_coo/core.py
@@ -621,6 +621,8 @@ class COO(SparseArray, NDArrayOperatorsMixin):  # lgtm [py/missing-equals]
         >>> np.count_nonzero(x) == s.nnz
         True
         """
+        if self.shape == ():
+            return len(self.data)
         return self.coords.shape[1]
 
     @property
--- a/sparse/tests/test_coo.py
+++ b/sparse/tests/test_coo.py
@@ -1608,16 +1608,18 @@ def test_random_idx_dtype():
     with pytest.raises(ValueError):
         sparse.random((300,), density=0.1, format="coo", idx_dtype=np.int8)
 
+
 def test_nnz_for_zero_dimensional_array():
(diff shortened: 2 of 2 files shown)


sparse/_coo/core.py
```diff
@@ -621,6 +621,8 @@ def nnz(self):
         >>> np.count_nonzero(x) == s.nnz
         True
         """
+        if self.shape == ():
+            return len(self.data)
         return self.coords.shape[1]
```
hameerabbasi, line 626: What I'd be concerned about is... Why is `self.coords.shape[1]` not correct for a scalar. Could we look into the constructor?

mangecoeur: Zero dimensional array also has zero dimensional coordinates at the moment:

```python
a = sparse.COO.from_numpy(np.array(1))
a.coords

#array([], shape=(0, 0), dtype=int64)
```

This makes sense to me (the length of a coordinate array in 0-dimensions would also be zero).

hameerabbasi: The shape of coords should theoretically always be `(ndim, nnz)`, so if there is a scalar array... It should be `(0, 1)`.

mangecoeur:
Looks like my tests don't work anyway 😞

hameerabbasi:
@mangecoeur Do you plan on continuing this PR?

mangecoeur:
hi, I saw in another thread that there is work on doing a substantial rewrite of sparse. If that's the case I would not contiue with the PR. Otherwise I might eventually get round to this, but not in the very near future.

hameerabbasi:
Closing based on https://github.com/pydata/sparse/pull/576#issuecomment-1885502033, feel free to re-open or create a new PR.

PR #579, opened 2023-02-24
change parsing of the input in einsum

HadrienNU:
After testing the implemented einsum (many thanks for implementing), it turns out that the ellipsis were not working (and they are used by xarray dot function, my primary motivation for making this work). So in this commit there is an added function _parse_einsum input that is borrowed from numpy with the only change that operand are not cast to numpy array but are left untouched.

### DIFF
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -1212,13 +1212,12 @@ def _dot_ndarray_coo_type_sparse(dt1, dt2):
     return _dot_ndarray_coo
 
 
+# Copied from : https://github.com/numpy/numpy/blob/59fec4619403762a5d785ad83fcbde5a230416fc/numpy/core/einsumfunc.py#L523 
+# under BSD-3-Clause license : https://github.com/numpy/numpy/blob/v1.24.0/LICENSE.txt
 def _parse_einsum_input(operands):
     """
     A copy of the numpy parse_einsum_input that 
     does not cast the operands to numpy array.
-    
-    Copied from : https://github.com/numpy/numpy/blob/main/numpy/core/einsumfunc.py
-    under BSD-3-Clause license : https://github.com/numpy/numpy/blob/main/LICENSE.txt
     
     Returns
     -------
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -1216,6 +1216,10 @@ def _parse_einsum_input(operands):
     """
     A copy of the numpy parse_einsum_input that 
     does not cast the operands to numpy array.
+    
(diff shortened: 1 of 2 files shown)


sparse/_common.py
```diff
@@ -1212,6 +1212,173 @@ def _dot_ndarray_coo(array1, coords2, data2, out_shape):  # pragma: no cover
     return _dot_ndarray_coo
 
 
+def _parse_einsum_input(operands):
```
hameerabbasi, line 1215: Please leave a comment where this was taken and a link to the airport license.

hameerabbasi (changes requested):

sparse/_common.py
HadrienNU: Sure, i have add this:

`Copied from : https://github.com/numpy/numpy/blob/main/numpy/core/einsumfunc.py
    under BSD-3-Clause license : https://github.com/numpy/numpy/blob/main/LICENSE.txt`

Do you think it is sufficient? (Licence are the same in sparse and numpy so that shouldn't be an issue and BSD-3 allow for copy and modification)

I have also removed the part of the code that was causing the fail of the test

hameerabbasi: I'd suggest adding it in a comment above the function instead of in the docstring, otherwise the note looks fine. I'd also link the exact commit, rather than Main which can have changing line numbers.

HadrienNU: Ok, done

hameerabbasi:
Seems like a formatting fail -- Perhaps update `black` to the latest version and re-run `black .`?

Also, there are a lot of lines missing from coverage, more comprehensive tests would be appreciated.

HadrienNU:
Ok, I will add more tests

hameerabbasi (changes requested):
One last change!

sparse/tests/test_einsum.py
```diff
@@ -94,12 +98,60 @@ def test_einsum(subscripts, density):
...
+@pytest.mark.parametrize("subscript", [0, [0, 0]])
+def test_einsum_type_error(subscript):
+    x = sparse.random(shape=(2,), density=0.5, format="coo")
+    y = sparse.random(shape=(2,), density=0.5, format="coo")
+    with pytest.raises(TypeError):
+        sparse.einsum(subscript, x, y)
+
+
+#
+# def test_einsum_type_error_nosubscript():
```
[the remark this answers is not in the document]
HadrienNU: Done, sorry

hameerabbasi (approved):

hameerabbasi:
Thanks, @HadrienNU for the improvement!

Issue #531, opened 2021-11-21
Convert two assignments to augmented source code


elfring:
:eyes: Some source code analysis tools can help to find opportunities for improving software components.
:thought_balloon: I propose to [increase the usage of augmented assignment statements](https://docs.python.org/3/reference/simple_stmts.html#augmented-assignment-statements "Augmented assignment statements") accordingly.

```diff
diff --git a/sparse/_compressed/indexing.py b/sparse/_compressed/indexing.py
index df35a6c..abe270a 100644
--- a/sparse/_compressed/indexing.py
+++ b/sparse/_compressed/indexing.py
@@ -151,7 +151,7 @@ def getitem(x, key):
             indptr = np.empty(shape[0] + 1, dtype=x.indptr.dtype)
             indptr[0] = 0
             np.cumsum(np.bincount(uncompressed, minlength=shape[0]), out=indptr[1:])
-            indices = indices % size
+            indices %= size
 
     arg = (data, indices, indptr)
 
diff --git a/sparse/_slicing.py b/sparse/_slicing.py
index 4e46244..7169ead 100644
--- a/sparse/_slicing.py
+++ b/sparse/_slicing.py
@@ -40,7 +40,7 @@ def normalize_index(idx, shape):
             continue
         else:
             n_sliced_dims += 1
-    idx = idx + (slice(None),) * (len(shape) - n_sliced_dims)
+    idx += (slice(None),) * (len(shape) - n_sliced_dims)
     if len([i for i in idx if i is not None]) > len(shape):
         raise IndexError("Too many indices for array")
 
```

hameerabbasi:
Hey, would you be willing to make a PR?

hameerabbasi:
This was closed.

PR #586, opened 2023-04-04
Convert two assignments to augmented source code #531

MHRasmy:


hameerabbasi:
Thanks, @MHRasmy!

elfring:
Thanks for your integration of a small source code adjustment.

PR #597, opened 2023-06-30
Minor bug: import warnings in `_sparse_array.SparseArray`

jamestwebber:
I noticed that `SparseArray.var` can potentially try to call `warnings.warn` ([here](https://github.com/pydata/sparse/blob/master/sparse/_sparse_array.py#L767)) but it was never imported.


hameerabbasi:
Thanks, @jamestwebber.

PR #599, opened 2023-07-01
Modernize packaging.

hameerabbasi:


jamestwebber (commented):
Minor suggestion: you can drop `flake8` entirely as ruff provides those on its own. 

I'm having trouble figuring out if the universal `bdist_wheel` setting can be moved to `pyproject.toml` -- it feels like that _should_ be supported, and is a bug in setuptools if not. But if it works then `setup.cfg` can be removed entirely.

hameerabbasi:
Thanks for the review! Yeah, I was considering that as well.

jamestwebber:
> Thanks for the review! Yeah, I was considering that as well.

I'm not sure I would call it a review in that I barely looked at most of the change 😂 but I will take a more thorough look later and test things out.

jamestwebber:
It looks like you already removed flake8, actually. Just a few references to it in the contributing docs.

As an experiment I removed `setup.cfg`, installed `build` and ran `python -m build`. It seems to build a universal wheel (it's named `sparse-0.14.1.dev10+gdfda838.d20230701-py3-none-any.whl`, at least) without any config, as well as an sdist. I'm not sure if that matches up with how you made releases before, though.

hameerabbasi:
This was superceded by #616 and #617. Of course, further lints might be added to Ruff.

PR #601, opened 2023-07-02
Preserve gcxs compression

### LINKED ISSUE #596: numpy ufuncs can change the compression of a gcxs array
I didn't see a redundant issue for this but I may have missed it.

**Describe the bug**

I have a GCXS array with compressed rows (aka a CSR array). I want to transform the data with e.g. `sqrt` or `log1p`. numpy ufuncs work great for this, but surprisingly I end up with a CSC array afterwards. I'm not sure why this would happen--it should basically just apply the ufunc to `data` and return the same indices.

I've been trying to write sparse-friendly code which means I need to use the array internals directly, but if the compression changes I get the wrong results (and/or will do something inefficient)

**To Reproduce**

```
import numpy as np
import sparse

m = sparse.random((10, 8), density=0.1, format="gcxs", compressed_axes=(0,))
assert m.compressed_axes == (0,)
assert np.sqrt(m).compressed_axes == (1,)
```

**Expected behavior**
I didn't expect my compression to change, and downstream methods were specialized for compressed rows.

**System**
 - OS and version = debian 11 VM on GCP
 - `sparse` version '0.14.0'
 - NumPy version '1.24.3'
 - Numba version '0.57.0'

jamestwebber:
This PR closes #596 in a fairly simple way: if all of the inputs are GCXS arrays and they have the same compression, it will return the output with the same compression. I just do this by storing a `kwargs` dictionary during the `_Elemwise` initialization.

A further enhancement that I'm exploring: if the `ufunc` has a single input, we could avoid the conversion to `COO` entirely, as well as skip some other checks. In this case we know that our argument is sparse (or we wouldn't be creating this class) and we know that it "broadcasts" to the same shape. The things to consider are the potential keywords for the ufunc: `where`, `out`, etc.

### DIFF
--- a/sparse/_umath.py
+++ b/sparse/_umath.py
@@ -412,6 +412,7 @@ class _Elemwise:
 
         processed_args = []
         out_type = GCXS
+        out_kwargs = {}
 
         sparse_args = [arg for arg in args if isinstance(arg, SparseArray)]
 
@@ -421,6 +422,7 @@ class _Elemwise:
             out_type = DOK
         elif all(isinstance(arg, GCXS) for arg in sparse_args):
             out_type = GCXS
+            out_kwargs["compressed_axes"] = sparse_args[0].compressed_axes
         else:
             out_type = COO
 
@@ -441,6 +443,7 @@ class _Elemwise:
                 return
 
         self.out_type = out_type
+        self.out_kwargs = out_kwargs
         self.args = tuple(processed_args)
         self.func = func
         self.dtype = kwargs.pop("dtype", None)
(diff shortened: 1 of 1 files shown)


hameerabbasi:
@jamestwebber do you mind if we land #599 first? Should be done by tomorrow.

jamestwebber:
Yeah no rush here!

hameerabbasi:
@jamestwebber `pre-commit` and `pyproject.toml` is in, you can rebase this PR if you wish.

hameerabbasi:
Thanks for your patience and work on this, @jamestwebber!

jamestwebber:
One thing I didn't do is add a test for regression. "when I have time" 😂 I will open a new PR with a fast unary path and I can add tests then.

PR #605, opened 2023-08-30
Use matrepr for rich `__str__` and `_repr_html_`

### LINKED ISSUE #604: Consider MatRepr for `__repr__` and `_repr_html_`
I've noticed that there appears to be no native way to print/display a sparse matrix.
The docs always convert to dense for display, and the native methods only emit some metadata.
`todense` can be troublesome when the matrix is large, and the result does not visualize sparsity.

I went ahead and added pydata/sparse to [matrepr](https://github.com/alugowski/matrepr).

You can see it in action with pydata/sparse matrices (1D, 2D, 3D) in this [Jupyter notebook](https://nbviewer.org/github/alugowski/matrepr/blob/main/doc/demo-pydata-sparse.ipynb).

For the Python REPL, there is a simple monkey-patch that replaces `__repr__()` and prints matrices in an interactive Python shell: `import matrepr.patch.sparse`

Is this something of interest to the pydata/sparse community?

Example of a random 2D COO matrix:
image

alugowski:
If available, use the [matrepr](https://github.com/alugowski/matrepr) package to render array values in string and HTML output.

Reverts to current behavior if matrepr is not installed.

See this [Jupyter notebook](https://nbviewer.org/github/alugowski/matrepr/blob/main/doc/demo-pydata-sparse.ipynb) for a demo of what 1D, 2D, and 3D sparse arrays look like when rendered by matrepr.

Closes #604

### DIFF
--- a/sparse/_compressed/compressed.py
+++ b/sparse/_compressed/compressed.py
@@ -344,9 +344,10 @@ class GCXS(SparseArray, NDArrayOperatorsMixin):
         return self.transpose()
 
     def __str__(self):
-        return "<GCXS: shape={}, dtype={}, nnz={}, fill_value={}, compressed_axes={}>".format(
+        summary = "<GCXS: shape={}, dtype={}, nnz={}, fill_value={}, compressed_axes={}>".format(
             self.shape, self.dtype, self.nnz, self.fill_value, self.compressed_axes
         )
+        return self._str_impl(summary)
 
     __repr__ = __str__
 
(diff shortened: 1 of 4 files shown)


hameerabbasi (approved):
Fairly minimal and self-contained. Thanks!

hameerabbasi:
I'll take a look at the CI later today.

alugowski:
> I'll take a look at the CI later today.

I see a few other recent PRs with the same fail.

The message is right, `.pre-commit-config.yaml` does not appear in the repo, though I see some PRs don't run that check at all.

alugowski:
The test fails were from the linter, I implemented its suggestions.

hameerabbasi (approved):
LGTM, thanks!

PR #606, opened 2023-09-14
DOC: Add array representations to sparse array construction examples

alugowski:
Reran the examples in [Construct Sparse Arrays](https://sparse.pydata.org/en/stable/construct.html) with `matrepr` installed to show the result of each snippet.

Previously only the first example showed the result, but only as a dense array. This change illustrates the sparsity of the sparse arrays. Some examples become easier to follow, such as the last example showing a "bad" identity matrix with explicit zeros and a fill value of 1. A visualization of the array reinforces the written description.

### DIFF
--- a/docs/construct.rst
+++ b/docs/construct.rst
@@ -20,13 +20,16 @@ matrix:
    ...           [0, 1, 2, 3, 4]]
    >>> data = [10, 20, 30, 40, 50]
    >>> s = sparse.COO(coords, data, shape=(5, 5))
-
-   >>> s.todense()
-   array([[10,  0,  0,  0,  0],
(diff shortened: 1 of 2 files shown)


alugowski:
The `pre-commit.ci` test fails with this message:

```
=====> .pre-commit-config.yaml is not a file
```

that all recent PRs appear to encounter.

hameerabbasi:
@alugowski This seems to need a rebase after #616 and #617.

alugowski:
> @alugowski This seems to need a rebase after #616 and #617. CI is also green again.

Done!

hameerabbasi:
Thanks for your contribution, @alugowski

PR #611, opened 2023-12-04
Fixing "GCXS matmul => slice leads to incorrect results"

### LINKED ISSUE #607: `GCXS` matmul => slice leads to incorrect results
**Describe the bug**
Performing matmul `@` between two `GCXS` arrays then slicing the output can lead to incorrect results, most commonly an array of all zeros.

**To Reproduce**
```python
import numpy as np
from sparse import GCXS
# Two matmul compatible arrays with no zeros
left = GCXS.from_numpy(np.ones((4, 16)))
right = GCXS.from_numpy(np.ones((16, 2)))
dotprod = left @ right
assert dotprod.nnz == dotprod.size  # result all nonzeros
dotprod_dense = dotprod.todense()
# Slice to get the first row of the result
print(dotprod_dense[0])
# > [16. 16.]  # correct result, reduced dimension was 16
print(dotprod[0].todense())
# > [0. 0.]  # sliced result is all zeros
assert all(dotprod[i].nnz == 0 for i in range(dotprod.shape[0]))  # Every row is empty !
```

**Expected behavior**
Slicing the sparse result should be consistent with slicing the dense result.

**System**
 - OS and version: Ubuntu 20
 - `sparse` version 0.14.0
 - NumPy version 1.24.3
 - Numba version 0.57.0

**Additional context**
I wonder if there is a link to #602 which also refers to slicing? The result array has an unsorted `indices` (`array([1, 0, 1, 0, 1, 0, 1, 0])` with indptr `array([0, 2, 4, 6, 8])`).

EuGig:
It fixes #607. The solution simply inverts the order of indices and of data before returning them in `_dot_csr_csr`.
(I did not write any test, since I'm not implementing a new feature. Let me know if I have to do that.)


hameerabbasi:
Could you perhaps add a regression test, I.e. a test that would trigger the bug before and now doesn't.

sparse/tests/test_dot.py
```diff
@@ -74,7 +74,6 @@ def test_tensordot(a_shape, b_shape, axes, a_format, b_format):
     # tests for return_type=np.ndarray
     sa_sb = sparse.tensordot(sa, sb, axes, return_type=np.ndarray)
 
-    assert_eq(a_b, sa_sb)
```
hameerabbasi, line 77: Why was this line removed?

EuGig: I thought it was redundant but it is not. My bad

hameerabbasi (dismissed):

hameerabbasi:
It seems like a lot of existing tests are failing -- Would it be possible for you to look into that?

EuGig:
Yes, of course, already on that.

hameerabbasi:
That's quite odd -- I'll look into it in depth this weekend, but certainly not normal.

hameerabbasi:
> Hi @hameerabbasi, is it normal to have failing tests without any modification? I got 178 tests failed after simply cloning the repo.

Did you do an editable pip install before running the tests? If not, that's normal.

EuGig:
Oh yeah sorry, I fixed that. Now tests are passing.

hameerabbasi:
> Actually, pre-commit.ci is failing. I don't know why!

That can be ignored for now.

hameerabbasi (approved):

hameerabbasi:
Thanks for the fix, @EuGig!

PR #616, opened 2024-01-02
Replace setup.py and versioneer with pyproject.toml and setuptools_scm.

hameerabbasi:


hameerabbasi:
Looks like `pytest-black` fails to respect ignores, but only on Windows for some reason. I'll make a follow-up PR to switch to pre-commit which is how Black/Ruff should be done anyway.

PR #617, opened 2024-01-02
Introduce pre-commit.

hameerabbasi:


hameerabbasi:
@mtsokol can we make sure that we aren't reducing the Array API coverage here? For example, I noticed `any`, `all`, `min`, `max`, `sum`, `prod` weren't tested before. Just making sure there aren't any others.

PR #619, opened 2024-01-04
API: Add set functions [Array API]

mtsokol:
Hi @hameerabbasi,

This PR adds four set functions for COO format described in the [Array API standard](https://data-apis.org/array-api/latest/API_specification/set_functions.html):

- ~~`unique_all`~~
- `unique_counts`
- ~~`unique_inverse`~~
- `unique_values`

### DIFF
--- a/sparse/_coo/common.py
+++ b/sparse/_coo/common.py
@@ -1059,6 +1059,98 @@ def clip(a, a_min=None, a_max=None, out=None):
     return a.clip(a_min, a_max)
 
 
+def unique_all(x, /):
+    """
+    Returns the unique elements of an input array x, the first occurring indices
+    for each unique element in x, the indices from the set of unique elements that
+    reconstruct x, and the corresponding counts for each unique element in x.
+    """
+    from .core import COO
+    if not isinstance(x, COO):
+        raise ValueError(f"Only COO arrays are supported but {type(x)} was passed.")
+
+    x = x.flatten()
+    values, index, inverse, counts = np.unique(
+        x.data, return_index=True, return_inverse=True, return_counts=True
+    )
+    index = x.coords.squeeze()[index]
+    if x.nnz < x.size:
+        # find the first occurence of the fill value
+        last_idx = -1
+        first_fill_value = x.coords.max() + 1 if x.coords.size > 0 else 0
+        for idx in np.nditer(x.coords, flags=["zerosize_ok"]):
+            if idx - last_idx > 1:
+                first_fill_value = last_idx + 1
+                break
+            else:
+                last_idx = idx
+
+        values = np.concatenate([[x.fill_value], values])
+        index = np.concatenate([[first_fill_value], index])
+        inverse = inverse + 1
+        counts = np.concatenate([[x.size - x.nnz], counts])
+
+    from .._dok import DOK
+    result_inverse = DOK(shape=x.size, dtype=np.intp, fill_value=np.intp(0))
+    result_inverse[x.coords.squeeze()] = inverse
+
+    return values, index, result_inverse, counts
+
+
+def unique_counts(x, /):
+    """
+    Returns the unique elements of an input array x and the corresponding
+    counts for each unique element in x.
+    """
+    from .core import COO
+    if not isinstance(x, COO):
+        raise ValueError(f"Only COO arrays are supported but {type(x)} was passed.")
+
+    x = x.flatten()
+    values, counts = np.unique(x.data, return_counts=True)
+    if x.nnz < x.size:
+        values = np.concatenate([[x.fill_value], values])
+        counts = np.concatenate([[x.size - x.nnz], counts])
+    return values, counts
+
+
+def unique_inverse(x, /):
+    """
+    Returns the unique elements of an input array x and the indices from
+    the set of unique elements that reconstruct x.
+    """
+    from .core import COO
+    if not isinstance(x, COO):
+        raise ValueError(f"Only COO arrays are supported but {type(x)} was passed.")
+
+    x = x.flatten()
+    values, inverse = np.unique(x.data, return_inverse=True)
+    if x.nnz < x.size:
+        values = np.concatenate([[x.fill_value], values])
+        inverse = inverse + 1
+
+    from .._dok import DOK
+    result_inverse = DOK(shape=x.size, dtype=np.intp, fill_value=np.intp(0))
+    result_inverse[x.coords.squeeze()] = inverse
+
+    return values, result_inverse
+
+
+def unique_values(x, /):
+    """
+    Returns the unique elements of an input array x.
+    """
+    from .core import COO
+    if not isinstance(x, COO):
+        raise ValueError(f"Only COO arrays are supported but {type(x)} was passed.")
+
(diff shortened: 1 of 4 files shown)


sparse/tests/test_coo.py
```diff
@@ -1745,3 +1745,52 @@ def test_squeeze_validation(self):
 
         with pytest.raises(ValueError, match="Specified axis `0` has a size greater than one: 3"):
             s_arr.squeeze(0)
+
+
+class TestUnique:
```
hameerabbasi, line 1750: I'd parametrize the test over fill-values as well.

sparse/_coo/common.py
```diff
@@ -1059,6 +1059,102 @@ def clip(a, a_min=None, a_max=None, out=None):
     return a.clip(a_min, a_max)
 
 
+def unique_all(x, /):
+    """
+    Returns the unique elements of an input array x, the first occurring indices
```
hameerabbasi, line 1064: The docstrings can certainly be more complete.

hameerabbasi (changes requested):
I'll give this a more in-depth review later, but here are some high-level suggestions.

sparse/_coo/common.py
mtsokol: Right, I added `Parameters` and `Returns` sections.

hameerabbasi: An examples section would be really nice too. 😄

hameerabbasi:
Okay, so I looked through the standard -- For `unique_inverse` and `unique_all`, it doesn't make sufficient sense to implement them, because there's "no better algorithm than dense", please let me know if I'm missing something. The other two, it definitely makes sense to implement.

mtsokol:
> Okay, so I looked through the standard -- For `unique_inverse` and `unique_all`, it doesn't make sufficient sense to implement them, because there's "no better algorithm than dense", please let me know if I'm missing something. The other two, it definitely makes sense to implement.

For `unique_all` I agree as `indices` component requires finding the first occurrence of the fill value, I can make it throw an exception that it isn't supported by COO.
For `unique_inverse` I construct `inverse` as a DOK and only plug non-fill values so it isn't dense algorithm I think - but is `inverse` even useful here?

sparse/tests/test_coo.py
mtsokol: Done!

sparse/_coo/common.py
mtsokol: Right, done!

mtsokol:
I removed `unique_all` and `unique_inverse` functions (updated PR's description).

sparse/_coo/common.py
```diff
@@ -1059,6 +1059,98 @@ def clip(a, a_min=None, a_max=None, out=None):
...
+        raise ValueError(f"Only COO arrays are supported but {type(x)} was passed.")
+
+    x = x.flatten()
+    values, counts = np.unique(x.data, return_counts=True)
+    if x.nnz < x.size:
+        values = np.concatenate([[x.fill_value], values])
+        counts = np.concatenate([[x.size - x.nnz], counts])
+        sorted_indices = np.argsort(values)
+        values[sorted_indices] = values.copy()
+        counts[sorted_indices] = counts.copy()
```
hameerabbasi, line 1111: Do we need `.copy()` here?

sparse/_coo/common.py
```diff
@@ -1059,6 +1059,98 @@ def clip(a, a_min=None, a_max=None, out=None):
...
+    --------
+    >>> import sparse
+    >>> x = sparse.COO.from_numpy([1, 0, 2, 1, 2, -3])
+    >>> sparse.unique_values(x)
+    array([-3, 0, 1, 2])
+    """
+    from .core import COO
+
+    if not isinstance(x, COO):
+        raise ValueError(f"Only COO arrays are supported but {type(x)} was passed.")
```
hameerabbasi, line 1145: I'd do `.asformat(COO)` here.

sparse/_coo/common.py
```diff
@@ -1059,6 +1059,98 @@ def clip(a, a_min=None, a_max=None, out=None):
...
+    --------
+    >>> import sparse
+    >>> x = sparse.COO.from_numpy([1, 0, 2, 1, 2, -3])
+    >>> sparse.unique_counts(x)
+    UniqueCountsResult(values=array([-3,  0,  1,  2]), counts=array([1, 1, 2, 2]))
+    """
+    from .core import COO
+
+    if not isinstance(x, COO):
+        raise ValueError(f"Only COO arrays are supported but {type(x)} was passed.")
```
hameerabbasi, line 1102: Same, `.asformat(COO)`

hameerabbasi (changes requested):
Small changes, otherwise LGTM.

sparse/_coo/common.py
mtsokol: Yes, because for `a = np.array([2,3])`:
```python
In [49]: a[[1,0]] = a

In [50]: a
Out[50]: array([2, 2])
```

when we need:
```python
In [45]: a[[1,0]] = a.copy()

In [46]: a
Out[46]: array([3, 2])
```

sparse/_coo/common.py
mtsokol: Done!

sparse/_coo/common.py
mtsokol: Done!

sparse/_coo/common.py
hameerabbasi: Sounds good to me! Thanks for the perspective.

sparse/_coo/common.py
```diff
@@ -1122,7 +1214,7 @@ def _arg_minmax_common(
     from .core import COO
 
     if not isinstance(x, COO):
```
hameerabbasi, line 1216: An additional check here for `SparseArray` and `spmatrix`, otherwise fail, would be nice too.

sparse/_coo/common.py
```diff
@@ -1059,6 +1059,98 @@ def clip(a, a_min=None, a_max=None, out=None):
...
+    Examples
+    --------
+    >>> import sparse
+    >>> x = sparse.COO.from_numpy([1, 0, 2, 1, 2, -3])
+    >>> sparse.unique_counts(x)
+    UniqueCountsResult(values=array([-3,  0,  1,  2]), counts=array([1, 1, 2, 2]))
+    """
+    from .core import COO
+
+    if not isinstance(x, COO):
```
hameerabbasi, line 1101: An additional check here for `SparseArray` and `spmatrix`, otherwise fail, would be nice too.

hameerabbasi (changes requested):
Last set of comments otherwise LGTM.

sparse/_coo/common.py
mtsokol: Sure, added `spmatrix` and `SparseArray` input checks.

sparse/_coo/common.py
mtsokol: Done!

hameerabbasi (approved):
Looks great! Thanks for your work on this.

Issue #589, opened 2023-04-17
`einsum` doesn't handle `dtype` and `optimize` kwargs


dcherian:
**Describe the bug**
I tried to use `np.einsum` on a combination of dask and sparse arrays. I get
```
TypeError: einsum() got an unexpected keyword argument 'dtype'
```

**To Reproduce**
```python
import dask
import dask.array as da

data = dask.array.ones((5,5))

wgts = sparse.COO(coords=(0,1,2,3,4), data=[1, 1, 0.5, 1, 1], shape=(5,))
np.einsum('ij,i-> j', data, wgts)
```

**Expected behavior**
The output should match the pure numpy version
```python
np.einsum('ij,i-> j', data.compute(), wgts.todense())
```
```
array([4.5, 4.5, 4.5, 4.5, 4.5])
```
**System**
 - OS and version:  macOS 13.3
 - `sparse` version 0.14.0
 - NumPy version (`np.__version__`): 1.23.5
 - Numba version (`numba.__version__`): 0.56.4

**Additional context**

hameerabbasi:
@mtsokol Again, would this be something you'd be interested in?

mtsokol:
Sure! I can take care of it.

PR #620, opened 2024-01-07
API: Add `kwargs` to `sparse.einsum`

mtsokol:
Related issue: #589

Hi @hameerabbasi,

I think the original error was caused due to the fact that `np.einsum` has kwargs which is missing in `sparse.einsum`. There is no Array API standard definition for `einsum` function so maybe we can just add `**kwargs` to `sparse.einsum` to make it compatible.

If there are plans for refactoring `sparse.einsum` in the future, then there will be an opportunity to introduce `dtype` parameter.

### DIFF
--- a/sparse/_common.py
+++ b/sparse/_common.py
@@ -1393,7 +1393,7 @@ def _einsum_single(lhs, rhs, operand):
     return to_output_format(COO(new_coords, new_data, shape=new_shape, has_duplicates=True))
 
 
-def einsum(*operands):
+def einsum(*operands, **kwargs):
     """
     Perform the equivalent of :obj:`numpy.einsum`.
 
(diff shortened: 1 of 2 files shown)


hameerabbasi:
While the `optimize` kwarg certainly isn't urgent as it doesn't change the result, I'd consider the `dtype` kwarg to be sufficiently important to coerce all the inputs to it before an operation.

mtsokol:
> While the `optimize` kwarg certainly isn't urgent as it doesn't change the result, I'd consider the `dtype` kwarg to be sufficiently important to coerce all the inputs to it before an operation.

Sure! I added `dtype` parameter and input operators are mapped to it with `astype`.

hameerabbasi:
Thanks, @mtsokol!

PR #621, opened 2024-01-08
Add more lints to Ruff for better code quality.

hameerabbasi:
This PR also adds lints that check for NumPy 2.0 compatibility.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       args: ["--autofix", "--no-ensure-ascii"]
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.1.9
+  rev: v0.1.11
   hooks:
     - id: ruff
       args: ["--fix"]
--- a/benchmarks/benchmark_coo.py
+++ b/benchmarks/benchmark_coo.py
@@ -5,9 +5,9 @@ import numpy as np
 
 class MatrixMultiplySuite:
     def setup(self):
-        np.random.seed(0)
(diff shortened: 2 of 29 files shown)


hameerabbasi:
@mtsokol Do you primarily work on Windows? If so, would it be possible for you to look into why this fails only on Windows?

mtsokol:
> @mtsokol Do you primarily work on Windows? If so, would it be possible for you to look into why this fails only on Windows?

I only have macos machine unfortunately.

mtsokol:
@hameerabbasi I have and idea what might be causing the failure. In `test_elemwise.py` (where tests are failing) there's a number of `.astype(np.int_)`. `np.int_` is an informal "default integer", which I think causes trouble here - on Windows it's 32bit but here for the sparse operation it stays as 64bit.

While working on NEP 52 there was even a discussion to remove `np.int_` and `np.uint` to enforce users to use sized aliases explicitly (e.g. `np.int64`) or c-like names (`np.long`). `np.int_` will get removed in a few releases: https://github.com/numpy/numpy/issues/24743

Maybe a solution would be to replace `.astype(np.int_)` in `test_elemwise.py` with `.astype(np.int64)` or `np.int32`?

hameerabbasi:
> Maybe a solution would be to replace `.astype(np.int_)` in `test_elemwise.py` with `.astype(np.int64)` or `np.int32`? (it's only bitwise and bitshift tests that are failing)

That worked, thanks! I see you requested a review, I'll wait for that before merging.

mtsokol (approved):
LGTM!

PR #625, opened 2024-01-11
NumPy 2.0 support

hameerabbasi:


rgommers:
That isn't bad at all, only a few idiomatic changes:)

hameerabbasi:
> That isn't bad at all, only a few idiomatic changes:)

There was a bunch of `np.random` stuff in #621 as well. So this isn't the only change.

mtsokol (approved):

PR #627, opened 2024-01-12
API: Add `sort` and `take` functions for COO format

mtsokol:
Hi @hameerabbasi,

This PR adds `sort` and `take` Array API functions for the COO format to the main namespace.
`sort` function, where looping happens, is numba jitted.


sparse/__init__.py
```diff
@@ -140,6 +140,8 @@
     nansum,
     result_type,
     roll,
+    sort,
+    take,
```
hameerabbasi, line 144: This should be in `docs/generated/sparse.rst`

sparse/_coo/common.py
```diff
@@ -1143,6 +1137,113 @@ def unique_values(x, /):
...
+    x = _validate_coo_input(x)
+
+    original_ndim = x.ndim
+    if x.ndim == 1:
+        x = x[None, :]
+        axis = -1
+
+    x = moveaxis(x, source=axis, destination=-1)
+    x_shape = x.shape
+    x = x.reshape((np.prod(x_shape[:-1]), x_shape[-1]))
```
hameerabbasi, line 1198: You can specify `-1` to say "put the rest of the shape here".

```suggestion
    x = x.reshape((-1, x_shape[-1]))
```

sparse/_coo/common.py
```diff
@@ -1152,11 +1253,52 @@ def unique_values(x, /):
...
+) -> None:
+    assert coords.shape[0] == 2
+    group_coords = coords[0, :]
+    sort_coords = coords[1, :]
+
+    result_indices = np.empty_like(sort_coords)
+    offset = 0  # tracks where the current group starts
+
+    # iterate through all groups and sort each one of them
+    for unique_val in np.unique(group_coords):
```
hameerabbasi, line 1275: `group_coords` is guaranteed to be sorted, I'd just have a `last_grp` and check against that. Something like 

```python
for i, grp in enumerate(group_coords):
    if grp != last_grp:
        last_grp = grp
        ...
    ...
```

sparse/_coo/common.py
```diff
@@ -1152,11 +1253,52 @@ def unique_values(x, /):
...
+        # .copy() required by numba, as `reshape` expects a continous array
+        group = np.argwhere(group_coords == unique_val).copy()
```
hameerabbasi, line 1277: I'd just find `first_idx` and `last_idx` here and then slice.

sparse/_coo/common.py
```diff
@@ -1152,11 +1253,52 @@ def unique_values(x, /):
...
+        group = np.reshape(group, -1)
+        group = np.atleast_1d(group)
```
hameerabbasi, line 1279: Not needed anymore after above modification.

sparse/_coo/common.py
```diff
@@ -1152,11 +1253,52 @@ def unique_values(x, /):
...
+                data[group] = data[group][::-1]
+
+        # SORT INDICES
+        fill_value_count = sort_axis_len - group.size
+        indices = np.arange(group.size)
+        # find a place where fill_value would be
+        for pos in range(group.size):
+            if (not descending and fill_value < data[group][pos]) or (descending and fill_value > data[group][pos]):
+                indices[pos:] += fill_value_count
+                break
```
hameerabbasi, line 1297: Just `argsort` the `values` and add the `first_idx` on top if needed.

hameerabbasi (changes requested):
Great work here! I feel the algorithm for `sort` could use some improvement. I've left a guide on what to do in the comments, but feel free to ask questions if needed.

sparse/_coo/common.py
mtsokol: I think that `argsort` isn't sufficient. In COO format `coords` **must aways be sorted**.
So values must be actually sorted to get indices in the ascending order. If (generally) `coords` aren't sorted it results in an undefined behavior: 
https://github.com/pydata/sparse/blob/82fb0d5d538e5071a017b5ecc0904914b2cedf72/sparse/_coo/core.py#L47-L50
WDYT?

hameerabbasi: `sorted=True` is an optimization to be applied when passing in sorted coords; it will internally sort them anyway if not.

mtsokol: Ah, then I need to call `COO` constructor here instead of sorting `coords` inplace (which is not even correct here as I think about it right now, as it modifies passed object instead of operating on a copy)

mtsokol: I have one more question:
I don't think that adding the `first_idx` on top if needed is sufficient.

Given `np.argsort` results in `[0,1,2,3]` indices where data is `[0,1,3,4]` it might be possible that we have `fill_value=2` and `first_idx` needs to be added to only specific subset of indices.

In the current implementation it's done simply by finding a specific place where fill values go and using increasing the rest by `indices[pos:] += fill_value_count`.

With `np.argsort` output that is not sorted I need to iterate through all values to check if it's before or after fill values and increase only then.

In the current scenario:
1. We sort values
2. We define indices just as `range(x)`
3. We find a place where `fill_values` go and increase indices by this offset
4. The resulting coords is ready for COO without sorting

In the proposed scenario:
1. We argsort values (obtaining indices)
2. We need to iterate through all values to find those indices that go past `fill_values` and increase by offset
3. The resulting coords still need to be sorted when constructing COO.

I would prefer to stick to the first option, WDYT?

sparse/__init__.py
mtsokol: Done!

sparse/_coo/common.py
mtsokol: Done!

sparse/_coo/common.py
mtsokol: Done!

sparse/_coo/common.py
mtsokol: Removed.

sparse/_coo/common.py
hameerabbasi: That sounds perfectly reasonable to me.

sparse/tests/test_coo.py
```diff
@@ -1790,6 +1790,25 @@ def test_expand_dims(axis):
...
+    ],
+)
+@pytest.mark.parametrize("fill_value", [-1, 0, 1, 3])
+@pytest.mark.parametrize("axis", [0, 1, -1])
+@pytest.mark.parametrize("descending", [False, True])
+def test_sort(arr, fill_value, axis, descending):
+    s_arr = sparse.COO.from_numpy(arr, fill_value)
+
+    result = sparse.sort(s_arr, axis=axis, descending=descending)
+    expected = -np.sort(-arr, axis=axis) if descending else np.sort(arr, axis=axis)
```
hameerabbasi, line 1807:
```suggestion
    expected = np.sort(arr, axis=axis, descending=descending)
```

sparse/tests/test_coo.py
```diff
@@ -1790,6 +1790,25 @@ def test_expand_dims(axis):
...
+    "arr",
+    [
+        np.array([[0, 0, 1, 5, 3, 0], [1, 0, 4, 0, 3, 0], [0, 1, 0, 1, 1, 0]], dtype=np.int64),
+        np.array([[[2, 0], [0, 5]], [[1, 0], [4, 0]], [[0, 1], [0, -1]]], dtype=np.float64),
+    ],
+)
+@pytest.mark.parametrize("fill_value", [-1, 0, 1, 3])
+@pytest.mark.parametrize("axis", [0, 1, -1])
+@pytest.mark.parametrize("descending", [False, True])
+def test_sort(arr, fill_value, axis, descending):
```
hameerabbasi, line 1803: I'd add a `test_sort_random` which checks random values and edge cases such as completely dense, `nnz=0`, and differemt fill-values.

hameerabbasi (changes requested):
A couple of comments to ensure better testing.

sparse/tests/test_coo.py
hameerabbasi: Are we adding it here because of Array API support?

sparse/tests/test_coo.py
mtsokol: Yes, right now I'm also working on `descending` for NumPy: https://github.com/numpy/numpy/pull/25437

sparse/tests/test_coo.py
mtsokol: Yes, right now I'm also working on descending for NumPy: https://github.com/numpy/numpy/pull/25437

sparse/tests/test_coo.py
mtsokol: Sure! Added completely dense case to `test_sort` and fully empty as a separate test (with parametrization).

mtsokol:
Ok, I think I addressed all comment - the PR is ready for another review.

hameerabbasi (approved):
LGTM, feel free to hit the button. 😄

mtsokol:
> LGTM, feel free to hit the button. 😄

Thanks! Unfortunately I see: "You’re not [authorized](https://docs.github.com/articles/about-protected-branches/) to merge this pull request."

hameerabbasi:
> > LGTM, feel free to hit the button. 😄
> 
> Thanks! Unfortunately I see: "You’re not [authorized](https://docs.github.com/articles/about-protected-branches/) to merge this pull request."

What about now?

mtsokol:
Merged! Thank you!

PR #629, opened 2024-01-15
API: Add `expand_dims` and `flip` functions for COO format

mtsokol:
Hi @hameerabbasi,

This PR adds `expand_dims` and `flip` Array API functions for the COO format to the main namespace.


sparse/__init__.py
```diff
@@ -116,6 +116,7 @@
     std,
     sum,
     tensordot,
+    var,
```
hameerabbasi, line 119: This should also be added to `docs/generated/sparse.rst`.

hameerabbasi (changes requested):
LGTM, just the one comment.

sparse/__init__.py
mtsokol: Done!

hameerabbasi (approved):
LGTM, thanks for awesome work!

PR #636, opened 2024-01-26
CI: Remove Python 3.8 and update miniconda

mtsokol:
Hi @hameerabbasi,

This PR introduces CI changes that we did in `finch-tensor`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,15 +6,15 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python: ['3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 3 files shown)


hameerabbasi:
I would also update `pyproject.toml` to reflect that we no longer support Python 3.8.

mtsokol:
> I would also update `pyproject.toml` to reflect that we no longer support Python 3.8.

Done!

hameerabbasi (approved):

hameerabbasi:
Thanks for the updates, @mtsokol!

hameerabbasi:
Seems like we need to change the required CI checks -- I'll d that later today.

PR #638, opened 2024-02-01
Fix CI Python always at 3.10.

hameerabbasi:


.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           activate-environment: sparse-dev
           allow-softlinks: true
           environment-file: ci/environment.yml
-          python-version: '3.10'
+          python-version: ${{ matrix.python }}
```
mtsokol, line 81: Does it mean that `docs` stage will run against three different python versions? Is it beneficial? (I think that e.g. NumPy runs docs only against one python version)

mtsokol: Hmm but `docs` job doesn't define `matrix` entry.

mtsokol: Where is the `matrix.python` coming from here?

hameerabbasi:
My bad, I thought this was during the `tests` job.

PR #646, opened 2024-03-12
API: Establish `pydata_backend`

mtsokol:
Replaces #645 

Hi @hameerabbasi,

Including the new backend in `pydata/sparse` ended up a bit convoluted, therefore I decided to break this task into smaller steps.

Here's the first step which moves existing implementation to `sparse/pydata_backend` directory with tests and adds context manager that tracks backend contextvar. The test suite has a lot of pydata specific calls so it will be a separate effort to make tests common for both backends. 

I applied all your comments from #645 here. 

The next PR will run some stub tests with finch to see that it can be installed and run in CI.


sparse/__init__.py
```diff
@@ -1,323 +1,56 @@
...
+        return backend_module
+
+
+def __getattr__(attr):
+    if attr == "pydata_backend":
+        import sparse.pydata_backend as backend_module
+
+        return backend_module
+    if attr == "finch_backend":
+        import sparse.finch_backend as backend_module
```
hameerabbasi, line 52: Shouldn't these refer to `backend_var`?

sparse/pydata_backend/__init__.py
```diff
@@ -0,0 +1,322 @@
...
+    "unique_counts",
+    "unique_values",
+    "var",
+    "vecdot",
+    "where",
+    "zeros",
+    "zeros_like",
+]
+
+__array_api_version__ = "2022.12"
```
hameerabbasi, line 322: Maybe move this to the top level?

sparse/__init__.py
```diff
@@ -1,323 +1,56 @@
...
-    "vecdot",
-    "where",
-    "zeros",
-    "zeros_like",
-]
 
-__array_api_version__ = "2022.12"
+class BackendType(Enum):
+    pydata = "pydata"
+    finch = "finch"
```
hameerabbasi, line 10: I'd do the make these UpperCamelCase.

hameerabbasi (changes requested):
Thanks for the great work here, @mtsokol! A couple of comments.

sparse/__init__.py
hameerabbasi: Ah, I see what you're doing. I'd just import these at the top so they work out of the box:

```python
from . import pydata_backend, finch_backend
```

mtsokol: So the reason we have these lines is a special case when a user does e.g. `sparse.pydata_backend`. For anything else, e.g. `sparse.permute_dims` the `return getattr(...` line is used.

I can't import them at the top because these backends must be imported lazily. Importing `finch_backend` when Finch.jl is not installed will fail.

sparse/__init__.py
mtsokol: Done!

sparse/pydata_backend/__init__.py
mtsokol: Done!

sparse/tests/test_backends.py
```diff
@@ -0,0 +1,17 @@
+import contextlib
+
+import sparse
+
+import pytest
+
+import numpy as np
+
+
+@pytest.mark.parametrize("backend", [sparse.BackendType.pydata, sparse.BackendType.finch])
```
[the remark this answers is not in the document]
mtsokol: Sure! Added a global fixture.

sparse/__init__.py
hameerabbasi: Ah, that's true.

sparse/__init__.py
```diff
@@ -1,323 +1,58 @@
...
-    "zeros",
-    "zeros_like",
-]
-
 __array_api_version__ = "2022.12"
+
+
+class BackendType(Enum):
+    pydata = "PyData"
+    finch = "Finch"
```
hameerabbasi, line 12: The variable names as well, please.

sparse/tests/conftest.py
```diff
@@ -1,8 +1,9 @@
-import pytest
+import sparse
 
+import pytest
 
-@pytest.fixture(scope="session")
-def rng():
-    from sparse._utils import default_rng
 
-    return default_rng
```
hameerabbasi, line 8: Probably shouldn't be deleted, I guess it's used in the tests?

hameerabbasi (changes requested):
Small changes.

hameerabbasi:
We also need a rebase.

sparse/tests/conftest.py
mtsokol: This is displayed that way only in GitHub. `sparse/tests/conftest.py` was moved to `sparse/pydata_backend/tests/conftest.py`, and then I added a new file, `sparse/tests/conftest.py`.

In the diff it shows it as if the original file was changed, but you can still find `def rng():` in `sparse/pydata_backend`.

hameerabbasi: I see, in that case I'd merge the two files here... There should be only one `tests` directory at the top level.

hameerabbasi: For transparency: We decided to have backend-specific tests into their own directories, moving things out as they encompassed multiple backends.

rgommers:
Quick question: since `pydata_backend` and `finch_backend` are (I assume) private and are only supposed to be used via context manager backend selection, wouldn't it be better to prepend underscores to the directory names?

mtsokol:
Sure! I think they can be prepended with an underscore. There will always be backend-specific members, like GCXS only present in `pydata_backend`, but importing it is hidden by backend selection mechanism (env-var or context manager).

PR #651, opened 2024-03-21
Enable benchmarks

mtsokol:


hameerabbasi (changes requested):
A couple of suggestions. Did you get it working?

mtsokol:
> Did you get it working?

@hameerabbasi Yes! Here are results: https://github.com/pydata/sparse/actions/runs/8374665448/job/22930405042?pr=651#step:6:15

.github/workflows/ci.yml
```diff
@@ -91,6 +91,37 @@ jobs:
...
+          environment-file: ci/environment.yml
+          python-version: '3.9'
+          miniforge-version: latest
+      - name: Install asv
+        run: |
+          pip install asv
+          asv machine --yes
+      - name: Run benchmarks
+        run: |
+          asv run
```
hameerabbasi, line 124:
```suggestion
          asv run --quick
```

hameerabbasi (changes requested):
Another suggestion to cut down on CI time as benchmarks expand.

.github/workflows/ci.yml
mtsokol: Done!

asv.conf.json
```diff
@@ -1,11 +1,20 @@
...
+  "matrix": {
+    "env": {},
+    "env_nobuild": {},
+    "req": {}
+  },
   "project": "sparse",
   "project_url": "https://sparse.pydata.org/",
   "repo": ".",
-  "results_dir": ".asv/results",
+  "uninstall_command": [],
```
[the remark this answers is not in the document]
mtsokol: Done!

asv.conf.json
```diff
@@ -1,11 +1,20 @@
...
+    "HEAD"
+  ],
+  "build_command": [],
   "environment_type": "conda",
-  "html_dir": ".asv/html",
+  "install_command": [
+    "pip install ."
+  ],
+  "matrix": {
+    "env": {},
```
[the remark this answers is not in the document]
mtsokol: Actually we don't need `matrix` entry. Dependencies are installed in `pip install .`.

asv.conf.json
```diff
@@ -1,11 +1,25 @@
...
-  "dvcs": "git",
-  "env_dir": ".asv/env",
+  "branches": [
+    "HEAD"
+  ],
+  "build_command": [],
   "environment_type": "conda",
-  "html_dir": ".asv/html",
+  "install_command": [
+    "pip install ."
```
hameerabbasi, line 8:
```suggestion
    "pip install --no-deps ."
```

hameerabbasi (approved):

.github/workflows/ci.yml
mtsokol: @hameerabbasi Just to be sure - this option means that benchmarks in the PR aren't meaningful and they're only meant to make sure benchmarks can be run, right?

If someone actually wants to run benchmarks they need to do it locally.

hameerabbasi: Correct, because the machine we'd get for benchmarks would be inconsistent anyway, so comparisons are useless here.

Even if it was consistent it's inside a VM on top of a system where other stuff is running, so there would be a lot of noise across CI runs.

hameerabbasi (approved):

PR #656, opened 2024-03-26
Support Python 3.12 and drop 3.9.

hameerabbasi:


mtsokol (approved):

mtsokol:
Hmm I see that some CI jobs are hanging.

hameerabbasi:
They're required jobs; and they need to be changed due to the Python version change.

PR #657, opened 2024-03-26
Update Finch backend

mtsokol:
Hi @hameerabbasi,

This small PR updates Finch backend with what we currently have in the latest `finch-tensor` release.

### DIFF
--- a/sparse/finch_backend/__init__.py
+++ b/sparse/finch_backend/__init__.py
@@ -3,11 +3,68 @@ try:
 except ModuleNotFoundError as e:
     raise ImportError("Finch not installed. Run `pip install sparse[finch]` to enable Finch backend") from e
 
-from finch import Tensor, astype, permute_dims
+from finch import (
+    add,
+    astype,
+    bool,
+    compiled,
+    complex64,
+    complex128,
+    compute,
+    divide,
+    float16,
+    float32,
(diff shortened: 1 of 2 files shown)


sparse/finch_backend/__init__.py
```diff
@@ -3,11 +3,93 @@
...
+    Element,
+    Pattern,
+    RepeatRLE,
+    SparseByteMap,
+    SparseCOO,
+    SparseHash,
+    SparseList,
+    SparseVBL,
+    Storage,
+    Tensor,
```
hameerabbasi, line 18: I would keep the API as Finch-agnostic as possible.

hameerabbasi (changes requested):

sparse/finch_backend/__init__.py
mtsokol: But to define e.g. `Storage(Dense(...))` we would need to `import finch` separately. I think for now we can keep it in the `finch_backend` namespace, WDYT?

mtsokol: @hameerabbasi Done! I removed Finch-specific objects from the import list.

hameerabbasi (approved):
LGTM, thanks @mtsokol

PR #658, opened 2024-03-28
API: Update `asarray` function

mtsokol:


sparse/pydata_backend/_compressed/compressed.py
```diff
@@ -857,14 +868,19 @@ class CSR(_Compressed2d):
     Sparse supports 2-D CSR.
     """
 
-    def __init__(self, arg, shape=None, prune=False, fill_value=0):
-        super().__init__(arg, shape=shape, compressed_axes=(0,), fill_value=fill_value)
+    def __init__(self, arg, shape=None, compressed_axes=(0,), prune=False, fill_value=0):
+        assert compressed_axes == (0,), "CSR only accepts rows as compressed axis"
```
hameerabbasi, line 872: Let's replace this by `ValueError`.

sparse/pydata_backend/_compressed/compressed.py
```diff
@@ -882,14 +898,19 @@ class CSC(_Compressed2d):
     Sparse supports 2-D CSC.
     """
 
-    def __init__(self, arg, shape=None, prune=False, fill_value=0):
-        super().__init__(arg, shape=shape, compressed_axes=(1,), fill_value=fill_value)
+    def __init__(self, arg, shape=None, compressed_axes=(1,), prune=False, fill_value=0):
+        assert compressed_axes == (1,), "CSR only accepts columns as compressed axis"
```
hameerabbasi, line 902: Same here.

hameerabbasi (changes requested):
Small changes.

sparse/pydata_backend/_compressed/compressed.py
```diff
@@ -847,6 +848,16 @@ def __str__(self):
     def ndim(self) -> int:
         return 2
 
+    @classmethod
+    @abstractmethod
+    def get_compressed_axes(cls):
+        pass
```
hameerabbasi, line 854: Do we need this method (and the ones in the classes below)?

sparse/pydata_backend/_compressed/compressed.py
mtsokol: Done!

sparse/pydata_backend/_compressed/compressed.py
mtsokol: Done!

sparse/pydata_backend/_compressed/compressed.py
mtsokol: I think yes. `from_coo` that is called in `from_numpy` needs to have `compressed_axes`. If it's `None` then it selects one based on best compression ratio. So there must be this piece of information in `CSC` and `CSR` classes, namely which `compressed_axes` should be used.

hameerabbasi: There's already the `compressed_axes` attribute/property for each instance.

mtsokol: But it's an instance attribute - available once object is created. Here we want to make sure that `compressed_axes` is exactly as the class require it (therefore we need a class attribute here for `CSC` and `CSR`, to make sure we have correct `compressed_axes` on creation).

mtsokol: Alternatively we could have copied `def from_coo` for `CSR`/`CSC` that just hardcode `compressed_axes` but I think it isn't clean approach. I would stick to class attribute that says what `compressed_axes` is.

hameerabbasi: In that case, let's use a class attribute rather than a class method.

mtsokol: Sure - done!

hameerabbasi (approved):

PR #660, opened 2024-04-08
BUG: Fix CSR/CSC matmul

mtsokol:
Hi @hameerabbasi,

I noticed that matmul for CSR/CSC in `sparse` is broken:
```python
import sparse
import scipy.sparse as sps

arr_csr = sps.random_array((10, 10), format='csr')
sparse_csr = sparse._compressed.CSR.from_scipy_sparse(arr_csr)

sparse_csr @ sparse_csr
```
throws a `ValueError`. This PR fixes `transpose` for `_compressed` 2D formats.


hameerabbasi:
Thanks for the fix, @mtsokol!

PR #668, opened 2024-04-26
CI: Add `array_api_tests` CI job

mtsokol:
Hi @hameerabbasi,

This PR adds `array_api_tests` CI job that runs `array_api_tests/test_signatures.py` test suite against Finch backend. 

[`array-api-skips.txt`](https://github.com/pydata/sparse/compare/array-api-test-ci?expand=1#diff-8179f4caec7bd03cc4b0d22c2cd5588280875d851df625a54859ebd6ff98d23f) file contains all currently unimplemented functions.


hameerabbasi:
Is there a way to ensure that if we add new functionality, we HAVE to remove the corresponding lines from the skips file? Basically treat them as xfail and use this: [scrict xfails](https://docs.pytest.org/en/latest/how-to/skipping.html#strict-parameter)

mtsokol:
I don't think `array-api-tests` supports it. Skips file just defines which functions are missing. [We did it this way in NumPy](https://github.com/numpy/numpy/pull/25167) and we successfully got to 99% compatibility after a few months. One needs to remember to remove specific lines from this file.

hameerabbasi (approved):

PR #669, opened 2024-04-29
[Finch] API: More creation functions

mtsokol:
Hi @hameerabbasi,

This PR adds new creation functions that we just added to `finch-tensor`.

With `zeros` implemented, `array-api-tests` suite is able to also test `Tensor` object namespace, therefore I added a few skips for the `test_array_method_signature` test, that are rather low priority (such as `__dlpack__` or `__int__`)


hameerabbasi:
Thanks @mtsokol! We seem to be missing the bitwise logical functions you added.

mtsokol:
Done!

PR #673, opened 2024-05-07
Add `pagerank` example

mtsokol:
Comment: https://github.com/Quansight-Labs/darpa-taco-sparse-mgmt/issues/3#issuecomment-2089964771

Hi @hameerabbasi,

Here I share initial pagerank implementation (following [scipy_pagerank](https://networkx.org/documentation/stable/_modules/networkx/algorithms/link_analysis/pagerank_alg.html#pagerank)).

A few comments:
- It's slower than scipy implementation.
- `spdiag` is still missing in `finch-tensor`, therefore I use SciPy diag to create it.


examples/pagerank_example.py
```diff
@@ -0,0 +1,79 @@
...
+    alpha = sparse.asarray(alpha)
+    nodelist = list(G)
+    A = nx.to_scipy_sparse_array(G, dtype=float, format="csc")
+    A = sparse.asarray(A)
+    S = sparse.sum(A, axis=1)
+    S = sparse.where(sparse.asarray(0.0) != S, sparse.asarray(1.0) / S, S)
+
+    # TODO: spdiags https://github.com/willow-ahrens/Finch.jl/issues/499
+    Q = sparse.asarray(sp.csc_array(sp.spdiags(S.todense(), 0, *A.shape)))
+    A = Q @ A
```
hameerabbasi, line 26: IIUC This is equivalent to https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.diagonal.html#diagonal, which is [supported by the Numba backend](https://github.com/pydata/sparse/blob/400d7f0b52042ae21c49a7e30aea19b3d1198e47/sparse/numba_backend/_coo/common.py#L804). Let's use that instead.

```suggestion
    Q = sparse.diagonal(S)
    A *= Q
```

hameerabbasi: For Finch, we can add support for it later, as the example only needs to work on Numba for now.

hameerabbasi: You might need `Q[None, :]` to match the results, but otherwise this is equivalent.

mtsokol: But the `pagerank` example is run either with Finch or Numba backend. when we run with Finch backend (and I think it should be runnable with Finch) then we can't execute one line with Numba.

mtsokol: I would prefer to keep these examples Finch and Numba ready, and update implementation once both backends support something new.

hameerabbasi: > I would prefer to keep these examples Finch and Numba ready, and update implementation once both backends support something new.

According to the deliverable, we just need a script that will demonstrate a speedup against the "old" code. If you would prefer, we can always keep this PR open until `xp.diagonal` is supported by Finch.

hameerabbasi:
@mtsokol Is this still relevant?

mtsokol:
> @mtsokol Is this still relevant?

Let's leave it open for now. For power iteration we have HITS example, but once we have `diagonal` this will be a fine addition to the examples docs.

PR #678, opened 2024-05-15
Update Finch backend

mtsokol:
Hi @hameerabbasi,

This PR updates Finch backend to the latest (unreleased yet) `finch-tensor` version.
I also enabled full Array API test suite, on my machine, with unreleased `finch-tensor` the result of running it is:
```
============================ 149 failed, 602 passed, 180 skipped, 225 warnings in 1365.91s (0:22:45) =============================
```
So about 60% of tests are passing.


hameerabbasi (approved):

ci/array-api-skips.txt
```diff
@@ -1,104 +1 @@
...
-array_api_tests/test_signatures.py::test_array_method_signature[__dlpack__]
-array_api_tests/test_signatures.py::test_array_method_signature[__dlpack_device__]
-array_api_tests/test_signatures.py::test_array_method_signature[__float__]
-array_api_tests/test_signatures.py::test_array_method_signature[__index__]
-array_api_tests/test_signatures.py::test_array_method_signature[__int__]
-array_api_tests/test_signatures.py::test_array_method_signature[__setitem__]
-
-# missing 'stream' from signature
-array_api_tests/test_signatures.py::test_array_method_signature[to_device]
+array_api_tests/test_searching_functions.py::test_nonzero_zerodim_error
```
hameerabbasi, line 1: Why not also have this in the xfails file?

mtsokol: It's a bit tricky: 
```
array_api_tests/test_searching_functions.py::test_nonzero_zerodim_error
```
now passes as it just expects an exception to happen. But in `xfails` file we have:
```
array_api_tests/test_searching_functions.py::test_nonzero
```
as `nonzero` still has some discrepancies. This line expects XFAIL from both `test_nonzero` and `test_nonzero_zerodim_error` (because `test_nonzero` is a prefix of `test_nonzero_zerodim_error`!) but `test_nonzero_zerodim_error` passes, causing unexpected XPASS failure.

It's caused by the fact that one test's name is exactly a prefix of another one, where one passes and other doesn't - I would say bad luck.

Therefore let's keep `test_nonzero_zerodim_error` in skips until we have `test_nonzero` passing. After that we can remove skips file.

hameerabbasi (approved):

ci/array-api-skips.txt
mtsokol: Thankfully that's the only test (I think) where the name is exactly a prefix of another test's name 😅 

Prefixes are useful in `skips`/`xfails` files to include a whole group of tests/test files.

PR #683, opened 2024-05-21
Add tests for `scipy.sparse.csgraph` and `scipy.sparse.linalg`

mtsokol:
Hi @hameerabbasi,

Here's a PR that contains:
- Moved `test_unary[sign((x_i is -0 or x_i == +0)) -> 0]` to skips as this test is now flaky (it correctly failed in the PR but I see in the main branch it succeeded causing XPASS, so I guess `--derandomize` didn't get rid of all randomness)
- `test_scipy_sparse_csgraph_dispatch` with `scipy.sparse.csgraph` tests
- commented out a few linalg tests that require https://github.com/scipy/scipy/pull/20759 (and a new scipy release)
- `test_scikit_learn_dispatch`, skipped for now as it requires https://github.com/scikit-learn/scikit-learn/pull/29031

### DIFF
--- a/sparse/tests/test_backends.py
+++ b/sparse/tests/test_backends.py
@@ -93,7 +93,7 @@ def test_scipy_inv(backend, format, order):
 
 
 @pytest.mark.skip(reason="https://github.com/scipy/scipy/pull/20759")
-@pytest.mark.parametrize("format, order", [("coo", "C")])
+@pytest.mark.parametrize("format, order", [("csc", "F"), ("csr", "C"), ("coo", "F"), ("coo", "C")])
 def test_scipy_norm(backend, format, order):
     x = np.eye(10, order=order) * 2
     x_pydata = sparse.asarray(x, format=format)
@@ -104,7 +104,7 @@ def test_scipy_norm(backend, format, order):
 
 
 @pytest.mark.skip(reason="https://github.com/scipy/scipy/pull/20759")
-@pytest.mark.parametrize("format, order", [("coo", "C")])
+@pytest.mark.parametrize("format, order", [("csc", "F"), ("csr", "C"), ("coo", "F"), ("coo", "C")])
 def test_scipy_lsqr(backend, format, order):
     x = np.eye(10, order=order) * 2
     y = np.ones((10, 1), order=order)
@@ -116,7 +116,7 @@ def test_scipy_lsqr(backend, format, order):
 
 
 @pytest.mark.skip(reason="https://github.com/scipy/scipy/pull/20759")
-@pytest.mark.parametrize("format, order", [("coo", "C")])
+@pytest.mark.parametrize("format, order", [("csc", "F"), ("csr", "C"), ("coo", "F"), ("coo", "C")])
 def test_scipy_eigs(backend, format, order):
     x = np.eye(10, order=order) * 2
     x_pydata = sparse.asarray(x, format=format)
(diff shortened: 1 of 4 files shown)


sparse/tests/test_backends.py
```diff
@@ -69,19 +70,116 @@ def test_asarray(backend, format, order):
...
 
 @pytest.mark.parametrize("format, order", [("csc", "F"), ("csr", "C"), ("coo", "F"), ("coo", "C")])
-def test_scipy_sparse_dispatch(backend, format, order):
+def test_scipy_sparse_linalg_dispatch(backend, format, order):
     x = np.eye(10, order=order) * 2
     y = np.ones((10, 1), order=order)
+    x_pydata = sparse.asarray(x, format=format)
+    y_pydata = sparse.asarray(y, format="coo")
+    # x_sp = sps.coo_matrix(x)
+    # y_sp = sps.coo_matrix(y)
```
hameerabbasi, line 79: Let's remove commented code blocks.

sparse/tests/test_backends.py
```diff
@@ -69,19 +70,116 @@ def test_asarray(backend, format, order):
...
+    # lsqr
+    # actual_x, _ = splin.lsqr(x_pydata, y)[:2]
+    # expected_x, _ = sp.linalg.lstsq(x, y)[:2]
+    # assert_almost_equal(actual_x, expected_x.ravel())
+
+    # NOTE: https://github.com/scipy/scipy/pull/20759
+    # eigs
+    # actual_vals, _ = splin.eigs(x_pydata, k=3)
+    # expected_vals, _ = splin.eigs(x_sp, k=3)
+    # assert_almost_equal(actual_vals, expected_vals)
```
hameerabbasi, line 107: Let's parametrize the test over functions, and mark these as skips

sparse/tests/test_backends.py
```diff
@@ -69,19 +70,116 @@ def test_asarray(backend, format, order):
...
+
+
+@pytest.mark.parametrize("matrix_fn, format, order", [(sps.csc_matrix, "csc", "F")])
+def test_scipy_sparse_csgraph_dispatch(backend, matrix_fn, format, order):
```
hameerabbasi, line 111: Let's also parametrize this over functions.

hameerabbasi (changes requested):
Thanks for the work here, @mtsokol! I've left some comments, mostly that the tests should be parametrized over function.

You can use a module-level PyTest fixture if you don't want to create the arrays over and over again.

sparse/tests/test_backends.py
mtsokol: Hmm I'm not sure if it makes sense this way. There are slight differences between each functions (some of them require custom parameters, have different number of `assert_equal`, or require calling `todense()`). Parametrizing over functions with all these information would produce rather obscure code in my opinion.

I think it should be rather as one test (as right now) or separate unit test for each function. WDYT?

hameerabbasi: Hmm, that sounds like a good idea -- let's make them separate tests so we know what fails.

sparse/tests/test_backends.py
mtsokol: Done!

sparse/tests/test_backends.py
mtsokol: Done!

sparse/tests/test_backends.py
mtsokol: Done!

hameerabbasi (approved):

hameerabbasi:
Thanks for working on this @mtsokol!

PR #684, opened 2024-05-22
Upgrade `finch-tensor`

mtsokol:


examples/sddmm_example.py
```diff
@@ -6,7 +6,7 @@
 import scipy.sparse as sps
 
 LEN = 10000
-DENSITY = 0.0001
+DENSITY = 0.00001
```
hameerabbasi, line 9: Why do we need this change?

hameerabbasi (commented):
A question.

examples/sddmm_example.py
mtsokol: This example runs a bit faster now - SDDMM takes the longest time out of all examples. It also shows Finch speedup even better.

hameerabbasi (approved):

mtsokol:
Can we also make a new alpha release?

hameerabbasi:
Well, we don't need one for this do we? It's bumping the lower bound on a dependency, but one could just do `pip install -U finch-tensor`.

mtsokol:
> Well, we don't need one for this do we? It's bumping the lower bound on a dependency, but one could just do `pip install -U finch-tensor`.

Last release was on 13.05, we still don't have https://github.com/pydata/sparse/pull/678 released in another alpha. I think we could have a new one.

hameerabbasi:
> Last release was on 13.05, we still don't have #678 API changes released in another alpha. I think we could have a new one.

Done! https://pypi.org/project/sparse/0.16.0a7/

mtsokol:
Thank you!

PR #689, opened 2024-05-29
Support direct Numba imports

mtsokol:
Hi @hameerabbasi,

I noticed that downstream libraries use direct imports for some Numba objects, e.g.: `from sparse import COO` (https://github.com/search?q=%22from+sparse+import%22+language%3APython&type=code). Since we will be releasing version `0.16.0` in some time, we need to support this kind of import to avoid disruption to users.

### DIFF
--- a/sparse/__init__.py
+++ b/sparse/__init__.py
@@ -43,6 +43,16 @@ class Backend:
             raise ValueError(f"Invalid backend identifier: {backend}")
         return backend_module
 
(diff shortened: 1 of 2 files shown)


hameerabbasi:
For this pattern, we would probably need something else that dispatches appropriately.

mtsokol:
Replaced by https://github.com/pydata/sparse/pull/695

PR #691, opened 2024-05-30
Expand Array API coverage for Numba backend

hameerabbasi:


sparse/numba_backend/_common.py
```diff
@@ -2066,6 +2081,7 @@ def _support_numpy(func):
     we want to flag it and dispatch to NumPy.
     """
 
+    @wraps(func)
```
hameerabbasi, line 2084: This change makes the functions wrapped by `_support_numpy` show the right signature, and is an easy win.

sparse/numba_backend/_common.py
```diff
@@ -2136,6 +2152,19 @@ def astype(x, dtype, /, *, copy=True):
...
+
+    Parameters
+    ----------
+    x : SparseArray
+    axis : int or tuple[int, ...], optional
+        The singleton axes to remove. By default all singleton axes are removed.
+
+    Returns
+    -------
+    SparseArray
```
mtsokol, line 2165:
```suggestion
    output : SparseArray
```

sparse/numba_backend/_common.py
```diff
@@ -2136,6 +2152,19 @@ def astype(x, dtype, /, *, copy=True):
 
 @_support_numpy
 def squeeze(x, /, axis=None):
+    """Remove singleton dimensions from array.
+
+    Parameters
+    ----------
+    x : SparseArray
```
mtsokol, line 2159: I think you need to add one line that describes `x`

mtsokol (commented):
LGTM, only docs are failing with one warning. I'm not sure if that's the fix but here are two suggestions.

sparse/numba_backend/_common.py
hameerabbasi: Done in https://github.com/pydata/sparse/pull/691/commits/1cf01c9a95c1095c48eb65bf1e39ba6e8ce9a3ab

sparse/numba_backend/_common.py
hameerabbasi: Done in https://github.com/pydata/sparse/pull/691/commits/1cf01c9a95c1095c48eb65bf1e39ba6e8ce9a3ab.

.github/workflows/ci.yml
```diff
@@ -173,7 +173,7 @@ jobs:
         SPARSE_BACKEND: ${{ matrix.backend }}
       run: |
         cd ${GITHUB_WORKSPACE}/array-api-tests
-        pytest array_api_tests -v -c pytest.ini -n 4 --ci --max-examples=2 --derandomize --disable-deadline -o xfail_strict=True --xfails-file ${GITHUB_WORKSPACE}/ci/${{ matrix.backend }}-array-api-xfails.txt --skips-file ${GITHUB_WORKSPACE}/ci/${{ matrix.backend }}-array-api-skips.txt
```
mtsokol, line 176: Why do we drop `--ci`?

sparse/numba_backend/tests/test_coo.py
```diff
@@ -1898,7 +1904,7 @@ def data_rvs(size):
     s1 = sparse.random(shape1, density=density, data_rvs=data_rvs)
     s2 = sparse.random(shape2, density=density, data_rvs=data_rvs)
 
-    axis = rng.integers(max(s1.ndim, s2.ndim))
+    axis = rng.integers(min(s1.ndim, s2.ndim))
```
mtsokol, line 1907: I think I would prefer to make `axis` a test parameter (`"shape1, shape2, axis"`) instead of a random integer (how do we know whether `pytest.raises(ValueError, match="Shapes must match along")` is run for every test case?).

hameerabbasi: Done in https://github.com/pydata/sparse/pull/691/commits/a9c01db04424b137c8167316e56391d8c6dcb759. I also fixed a few bugs relating to the array API suite, but triggering them requires https://github.com/data-apis/array-api-tests/pull/272.

mtsokol (approved):

PR #694, opened 2024-06-06
Upgrade `finch-tensor`

mtsokol:
Hi @hameerabbasi,

Once https://github.com/willow-ahrens/finch-tensor/pull/45 is merged and released this PR can go in. 
I also moved `pytest-xdist` to the test dependencies, WDYT?

### DIFF
--- a/benchmarks/benchmark_backends.py
+++ b/benchmarks/benchmark_backends.py
@@ -24,10 +24,10 @@ class Tensordot:
         if sparse.BackendType.Finch == BACKEND:
             import finch
 
-            self.s1 = self.s1.to_device(
+            self.s1 = self.s1.to_storage(
                 finch.Storage(finch.Dense(finch.SparseList(finch.Element(0.0))), order=self.s1.get_order())
             )
-            self.s2 = self.s2.to_device(
+            self.s2 = self.s2.to_storage(
                 finch.Storage(
                     finch.Dense(finch.SparseList(finch.SparseList(finch.Element(0.0)))),
                     order=self.s2.get_order(),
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -163,7 +163,7 @@ jobs:
         python-version: '3.11'
     - name: Install build and test dependencies from PyPI
       run: |
-        python -m pip install pytest-xdist -r array-api-tests/requirements.txt
+        python -m pip install -r array-api-tests/requirements.txt
     - name: Build and install Sparse
       run: |
         python -m pip install '.[finch]'
```
hameerabbasi, line 169: > I also moved pytest-xdist to the test dependencies, WDYT?

That won't work, unless we do this too.

```suggestion
        python -m pip install '.[finch,tests]'
```

hameerabbasi (commented):
Let's also add `-n $(nproc)` or whatever to each CI job.

.github/workflows/ci.yml
mtsokol: Ah, right. On the second thought let's drop these changes as they're not relevant to `finch-tensor` update and leave it for "light CI refactor" PR.

hameerabbasi: That sounds like a good idea to keep changes isolated.

hameerabbasi (approved):

hameerabbasi (dismissed):

mtsokol:
Maybe we can cut a new alpha release after merging it?

hameerabbasi:
> Maybe we can cut a new alpha release after merging it?

Done: https://pypi.org/project/sparse/0.16.0a8/

PR #697, opened 2024-06-10
Add elemwise examples

mtsokol:
Hi @hameerabbasi,

This PR adds elemwise examples that got removed from `asv`.

### DIFF
--- a/examples/elemwise_example.py
+++ b/examples/elemwise_example.py
@@ -1,0 +1,72 @@
+import importlib
+import operator
+import os
+
+import sparse
+
+from utils import benchmark
+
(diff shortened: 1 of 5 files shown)


mtsokol:
Hmm it looks like macos tests started to fail with:
```
 INTERNALERROR> PermissionError: [Errno 1] Operation not permitted
```
that you mentioned earlier. I think it's strictly related to some dependency version as this PR introduces changes to `examples/` dir only (independent of macos tests) and in the main branch those tests passed three hours ago. Weird.

hameerabbasi (approved):

PR #699, opened 2024-06-11
Refactor CI jobs.

hameerabbasi:


hameerabbasi:
I think for the same reason as #697 we can ignore the macOS job.

mtsokol (approved):

PR #704, opened 2024-06-13
Fix 1D return dtype in argmax/argmin

mtsokol:
Hi @hameerabbasi,

This PR fixes `argmin`/`argmax` when input is 1D and result is a fill value. To reproduce:

```python
sparse.argmax(sparse.asarray(np.arange(1), dtype=np.uint8)).dtype  # defaulted to `float64` as `np.array([]).dtype` is `float64`. Now it's `intp`
```

### DIFF
--- a/sparse/numba_backend/_coo/common.py
+++ b/sparse/numba_backend/_coo/common.py
@@ -1456,7 +1456,7 @@ def _compute_minmax_args(
             if not found:
                 result_data.append(current_coord + 1)
 
-    return (result_indices, result_data)
+    return (result_indices, np.array(result_data, dtype=np.intp))
 
 
 def _arg_minmax_common(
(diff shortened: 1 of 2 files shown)


sparse/numba_backend/_coo/common.py
```diff
@@ -1456,7 +1456,7 @@ def _compute_minmax_args(
             if not found:
                 result_data.append(current_coord + 1)
 
-    return (result_indices, result_data)
+    return (result_indices, np.array(result_data, dtype=np.intp))
```
hameerabbasi, line 1459: Let's avoid a copy if we can?

```suggestion
    return (result_indices, np.asarray(result_data, dtype=np.intp))
```

hameerabbasi (changes requested):
Thanks, @mtsokol. Just one comment and this LGTM!

sparse/numba_backend/_coo/common.py
mtsokol: With the current implementation there's no way to avoid a copy.
The `result_data` that we pass to `np.array` is defined as:
```python
result_data = []
```
and then individual coords are added with `result_data.append(...)` in a loop.
So `result_data` can't be a view of another array. It's a Python list that is passed to NumPy (always a copy is made for Python lists), therefore `np.array` is fine here.

hameerabbasi (approved):

PR #707, opened 2024-06-19
BUG: Skip `test_where` test

mtsokol:


hameerabbasi:
It seems like it's not just one test that hangs -- when (# of workers) tests hang in parallel, then we run into the issue.

mtsokol:
Hmm right now it "hangs" on:
```
array_api_tests/test_searching_functions.py::test_nonzero XFAIL (--x...) [ 44%] 
```
with one worker only: https://github.com/pydata/sparse/actions/runs/9578134709/job/26407734342?pr=707

hameerabbasi:
Can we use something like https://github.com/marketplace/actions/debugging-with-ssh to get a stacktrace?

mtsokol:
Hmm it looks like I managed to connect as in the logs I see:
```
Connected Client(s): ...and new entries show up as I try to ssh
```
but on my side when connecting I get `Connection closed by 37.16.25.99 port 22`. You can also try to ssh into it as I added your username to the access section.

hameerabbasi:
> Hmm it looks like I managed to connect as in the logs I see:
> 
> ```
> Connected Client(s): ...and new entries show up as I try to ssh
> ```
> 
> https://github.com/pydata/sparse/actions/runs/9579045837/job/26410620454?pr=707
> 
> but on my side when connecting I get `Connection closed by 37.16.25.99 port 22`. You can also try to ssh into it as I added your username to the access section.

Do you know what IP to connect to? Seems like it's a generic domain which can't possibly be it.

hameerabbasi:
CI Succeeded! 🥳 Let's add back the parallel workers and also remove the debugging code?

hameerabbasi (approved):

PR #708, opened 2024-06-21
Re-word backend selection warning.

hameerabbasi:


sparse/__init__.py
```diff
@@ -14,8 +14,12 @@ class BackendType(Enum):
 
 _ENV_VAR_NAME = "SPARSE_BACKEND"
 
-if _ENV_VAR_NAME in os.environ:
-    warnings.warn("Selectable backends feature in `sparse` might change in the future.", FutureWarning, stacklevel=1)
+if os.environ.get(_ENV_VAR_NAME, "") != "":
```
hameerabbasi, line 17: This was so that setting an empty string (which is common in bash scripts/CI) won't trip up the warning.

mtsokol (approved):

PR #716, opened 2024-07-08
MNT: Add CI mode to examples

mtsokol:
This PR adds CI mode to examples reducing `examples` job execution time (each benchmark returns after precompilation stage).

### DIFF
--- a/ci/test_examples.sh
+++ b/ci/test_examples.sh
@@ -1,3 +1,3 @@
 for example in $(find ./examples/ -iname *.py); do
-  python $example
+  CI_MODE=True python $example
 done
--- a/ci/test_examples.sh
+++ b/ci/test_examples.sh
@@ -1,3 +1,3 @@
 for example in $(find ./examples/ -iname *.py); do
-  CI_MODE=True python $example
+  CI_MODE=1 python $example
 done
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -3,17 +3,28 @@ import time
 from collections.abc import Callable, Iterable
 from typing import Any
 
-CI_MODE = os.getenv("CI_MODE", default=False)
+CI_MODE = bool(int(os.getenv("CI_MODE", default="0")))
 
 
-def benchmark(func: Callable, args: Iterable[Any], info: str, iters: int):
+def benchmark(
+    func: Callable,
+    args: Iterable[Any],
+    info: str,
+    iters: int,
+) -> object:
+    # Compile
+    result = func(*args)
+
     if CI_MODE:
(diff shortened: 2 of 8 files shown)


examples/utils.py
```diff
@@ -1,9 +1,16 @@
+import os
 import time
 from collections.abc import Callable, Iterable
 from typing import Any
 
+CI_MODE = os.getenv("CI_MODE", default=False)
```
hameerabbasi, line 6:
```suggestion
CI_MODE = bool(int(os.getenv("CI_MODE", default="0")))
```

ci/test_examples.sh
```diff
@@ -1,3 +1,3 @@
 for example in $(find ./examples/ -iname *.py); do
-  python $example
+  CI_MODE=True python $example
```
hameerabbasi, line 2:
```suggestion
  CI_MODE=1 python $example
```

examples/utils.py
```diff
@@ -1,9 +1,16 @@
...
+
 
 def benchmark(func: Callable, args: Iterable[Any], info: str, iters: int):
+    if CI_MODE:
```
hameerabbasi, line 10: We should have a warmup here.

hameerabbasi (changes requested):
Thanks for the PR, @mtsokol.

examples/utils.py
mtsokol: Warmup, or a precompilation stage, happens in each example file before `benchmark` is called. It isn't placed here as we also want the result of that call for `assert_allclose` statements.

ci/test_examples.sh
mtsokol: Is there a difference compared to the notebook job that we merged?

https://github.com/pydata/sparse/blob/0e95f8b5c6e98f73bfb2c5faba71c4f90abc17af/ci/test_notebooks.sh#L1

We use `True` there. Or do we want to also change it here?

hameerabbasi: I'd say also change it there. If there's repetition we should factor it out.

examples/utils.py
hameerabbasi: We could return from this function which is the result of the warmup call.

mtsokol: Done!

ci/test_examples.sh
mtsokol: Done!

examples/utils.py
mtsokol: Done!

hameerabbasi (approved):
Thanks for the work here, @mtsokol!

PR #719, opened 2024-07-22
ci: Adding release drafter

DeaMariaLeon:
The release drafter will classify new PRs according to their title (if the title is correctly written). 
The PRs titles can be manually corrected and the drafter will classify them (even if they are closed). 
Please notice that once the auto-label is set, the release drafter doesn't change it.

I **think** that there is no way to test these files before merging. The reason being that these files are supposed to be added directly to the repository on GitHub, _not through a PR_. (But I tested them on my own "toy" project and I also added them on the Narwhals project). 

The label names/classifications were **heavily**  inspired (copied) from polars' repo. I can change them if needed.


hameerabbasi:
Can we force people to have proper PR titles?

DeaMariaLeon:
I need to do find out how to do that.

hameerabbasi (approved):

hameerabbasi (changes requested):
Following the discussion on Slack, let's update `CONTRIBUTING.md` to include instructions for how to properly write PR titles.

hameerabbasi:
Looks like my cat likes adding unnecessary labels. 😅

DeaMariaLeon:
Added to `contributing.md`. What do you think?

DeaMariaLeon:
I added instructions on release-procedure.md.
Added also this to `.github.release-drafter.yml` so the PRs opened by the pre-commit are ignored by the drafter:
```
- label: 'skip changelog'
    title:
      - '/^\[pre-commit.ci\]/'
 ```
If there are more PR that need to be ignored, we need to make sure they are skipped.

hameerabbasi:
> Maybe we can add a note (of the PR title) on the PR template as well.

Let's leave this to a follow-up and get this in.

hameerabbasi (approved):

Issue #720, opened 2024-07-22
Add "dependabot" to the ci


DeaMariaLeon:
Going through GitHub documentation, there is a suggestion to add a "dependabot". 

If we set it up, it sends a PR when an update to actions is needed. Something to consider.. ? 

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

hameerabbasi:
That sounds like a great idea to me.

PR #723, opened 2024-07-25
ci: Adding dependabot

DeaMariaLeon:
Addresses [issue](https://github.com/pydata/sparse/issues/720). Many other `package-ecosystems` can be added. 

Should the link to the GitHub documentation be added to the file? (it's very long). 

This "dependabot" is also implemented on projects like pandas and Polars.


hameerabbasi (approved):

hameerabbasi:
@DeaMariaLeon Thanks! This is in.

PR #733, opened 2024-08-05
ci: WIP Benchmarks 

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [ ] 🧪 Test
- [x] 🛠️ Other

## Related issues

- Related issue #
- Closes #

## Checklist

- [ ] Tests added
- [ ] Documented the changes

***

WIP to test how to use codspeed


benchmarks/test_benchmark_coo.py
```diff
@@ -3,16 +3,16 @@
...
-        rng = np.random.default_rng(0)
-        self.x = sparse.random((100, 100), density=0.01, random_state=rng)
-        self.y = sparse.random((100, 100), density=0.01, random_state=rng)
+def test_MatrixMultiplySuite(benchmark):
+    rng = np.random.default_rng(seed=42)
+    x = sparse.random((3, 3), density=0.01, random_state=rng)
+    y = sparse.random((3, 3), density=0.01, random_state=rng)
 
-        self.x @ self.y  # Numba compilation
+    # self.x @ self.y  # Numba compilation
```
hameerabbasi, line 11: This should be uncommented -- The first time an operation is run after an import; some compilation happens. This gets rid of that overhead when benchmarking

benchmarks/test_benchmark_coo.py
```diff
@@ -3,16 +3,16 @@
...
 
-class MatrixMultiplySuite:
-    def setup(self):
-        rng = np.random.default_rng(0)
-        self.x = sparse.random((100, 100), density=0.01, random_state=rng)
-        self.y = sparse.random((100, 100), density=0.01, random_state=rng)
+def test_MatrixMultiplySuite(benchmark):
+    rng = np.random.default_rng(seed=42)
+    x = sparse.random((3, 3), density=0.01, random_state=rng)
+    y = sparse.random((3, 3), density=0.01, random_state=rng)
```
hameerabbasi, line 9: 3 is way too small: We'd like to have something statiscally significant. This is so small most of the measurements will be noise or overhead of the Python call, rather than actual number crunching.

benchmarks/test_benchmark_coo.py
```diff
@@ -3,16 +3,16 @@
 import numpy as np
 
 
-class MatrixMultiplySuite:
-    def setup(self):
-        rng = np.random.default_rng(0)
-        self.x = sparse.random((100, 100), density=0.01, random_state=rng)
-        self.y = sparse.random((100, 100), density=0.01, random_state=rng)
+def test_MatrixMultiplySuite(benchmark):
```
hameerabbasi, line 6:
```suggestion
def test_matmul(benchmark):
```

hameerabbasi (changes requested):
Comments so far.

DeaMariaLeon:
I was only testing if codspeed would read the test itself.

benchmarks/test_benchmark_coo.py
DeaMariaLeon: What should it be? I took 3 from `finch_example.py`

DeaMariaLeon:
@hameerabbasi I changed with your comments. 
I think this needs to be merged. From codspeed: 
_"You will start to see performance impacts in the reports once the benchmarks are run from your default branch."_

No matter how much I edit the files, there is only one benchmark result on codspeed's server. Just for the first time it detected the new benchmark.

hameerabbasi (approved):

PR #741, opened 2024-08-09
test: Adapt benchmarks to use `codspeed`

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

- Related issue #
- Closes #

## Checklist

- [x] Tests added
- [ ] Documented the changes

***

Adapted benchmarks from original file `benchmark_coo.py` that was used to benchmark with `asv`. Now benchmarking with `codspeed`.


benchmarks/test_benchmark_coo.py
```diff
@@ -11,6 +11,23 @@
 SEED = 42
```
hameerabbasi, line 11: I'd import this from a central place, so it can be changed just once.

benchmarks/test_benchmark_coo.py
```diff
@@ -36,3 +53,83 @@ def test_elemwise(benchmark, f, elemwise_args):
...
+def indexing_args(request):
+    side = request.param
+    rng = np.random.default_rng(seed=SEED)
+    if side**side >= 2**26:
+        pytest.skip()
+
+    x = sparse.random((side, side, side), density=0.01, random_state=rng)
+
+    # Numba compilation
+    x[5]
```
hameerabbasi, line 89: This should be done inside the benchmark, not the fixture. You can also re-use fixtures across tests/benchmarks.

benchmarks/test_benchmark_coo.py
```diff
@@ -36,3 +53,83 @@ def test_elemwise(benchmark, f, elemwise_args):
...
+
+    return x, side
```
hameerabbasi, line 91: Returning `x` here is enough, `side` can come from `x.shape`.

benchmarks/test_benchmark_coo.py
```diff
@@ -36,3 +53,83 @@ def test_elemwise(benchmark, f, elemwise_args):
...
+
+
+def time_index_scalar(benchmark, indexing_args):
+    x, side = indexing_args
+
+    @benchmark
+    def bench():
+        x[side / 2, side / 2, side / 2]
```
hameerabbasi, line 99: We need floor division here (`//` instead of `/`).

hameerabbasi (changes requested):
A couple of comments. Nice work overall!

benchmarks/test_benchmark_coo.py
DeaMariaLeon: Thanks for the feedback. Here:
>You can also re-use fixtures across tests/benchmarks.

I'm reusing `indexing_args`.. so, I'm not sure of what you mean.

benchmarks/test_benchmark_coo.py
DeaMariaLeon: True!

benchmarks/test_benchmark_coo.py
DeaMariaLeon: My bad

benchmarks/test_benchmark_coo.py
DeaMariaLeon: Ok

benchmarks/test_benchmark_coo.py
```diff
@@ -36,3 +52,91 @@ def test_elemwise(benchmark, f, elemwise_args):
...
+        pytest.skip()
+
+    return sparse.random((side, side, side), density=DENSITY, random_state=rng)
+
+
+def test_index_scalar(benchmark, indexing_args):
+    x = indexing_args
+    side = x.shape([0])
+
+    x[5]  # Numba compilation
```
hameerabbasi, line 92: I'd do exactly the same thing inside and outside the benchmark -- to make sure the same function is compiled.

benchmarks/test_benchmark_coo.py
```diff
@@ -36,3 +52,91 @@ def test_elemwise(benchmark, f, elemwise_args):
...
+    @benchmark
+    def bench():
+        x[side // 2, side // 2, side // 2]
+
+
+def test_index_slice(benchmark, indexing_args):
+    x = indexing_args
+    side = x.shape([0])
+
+    x[5]  # Numba compilation
```
hameerabbasi, line 103: Same here.

benchmarks/test_benchmark_coo.py
```diff
@@ -36,3 +52,91 @@ def test_elemwise(benchmark, f, elemwise_args):
...
+    @benchmark
+    def bench():
+        x[: side // 2]
+
+
+def test_index_slice2(benchmark, indexing_args):
+    x = indexing_args
+    side = x.shape([0])
+
+    x[5]  # Numba compilation
```
hameerabbasi, line 114: And here.

benchmarks/test_benchmark_coo.py
```diff
@@ -36,3 +52,91 @@ def test_elemwise(benchmark, f, elemwise_args):
...
+    @benchmark
+    def bench():
+        x[: side // 2, : side // 2]
+
+
+def test_index_slice3(benchmark, indexing_args):
+    x = indexing_args
+    side = x.shape([0])
+
+    x[5]  # Numba compilation
```
hameerabbasi, line 125: And here.

hameerabbasi (changes requested):
Small changes -- otherwise LGTM.

benchmarks/test_benchmark_coo.py
DeaMariaLeon: Could you clarify please?
I though you had said to remove this from the benchmark. 
BTW there are other issues. I'm working on that.

hameerabbasi: Below, inside the benchmark, you are doing 
```suggestion
    x[side // 2, side // 2, side // 2]  # Numba compilation
```

So you should do the same here.

DeaMariaLeon:
@hameerabbasi  Regarding:
>SEED = 42
>I'd import this from a central place, so it can be changed just once.

It's not working the way I did it on this commit. But locally I did something else: I put them inside a `conftest.py`. Which it seems to work, except that I noticed that SEED and DENSITY vary on other files. Like `elemwise_example.py`, `matmul_example.py`, etc.

hameerabbasi:
For seed, let's unify it. For density, we can leave it constant in each file where it's constant; and change it where it should be changed.

DeaMariaLeon:
>For seed, let's unify it.

What does this mean? set just one value for all the tests, on all the files? Set the value on `conftest.py`?

DeaMariaLeon:
There is this error on the tests, but why would that fail with what I did..? 

```
FAILED sparse/numba_backend/tests/test_dot.py::test_matmul_nan_warnings[a1-b1] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
= 1 failed, 5892 passed, 14 skipped, 94 xfailed, 63 xpassed, 2259 warnings in 169.96s (0:02:49) =
Error: Process completed with exit code 1.
```

benchmarks/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # Benchmark
+    print(info)
+    start = time.time()
+    for _ in range(iters):
+        func(*args)
+    elapsed = time.time() - start
+    print(f"Took {elapsed / iters} s.\n")
+
+    return result
```
hameerabbasi, line 30: We can get rid of these lines.

hameerabbasi (changes requested):
One final change and we should be good to go. The changes discussed out-of-band can take place on this PR or a follow-up, as you wish.

benchmarks/utils.py
DeaMariaLeon: Done

DeaMariaLeon:
I'll do the out-of-band changes on another PR if that's fine with you then.

hameerabbasi (approved):

PR #742, opened 2024-08-12
depr, chore: Enforce deprecations and clean up warnings.

hameerabbasi:

- [x] 💾 Refactor
- [ ] 🪄 Feature
- [ ] 🐞 Bug Fix
- [x] 🔧 Optimization
- [ ] 📚 Documentation
- [ ] 🧪 Test
- [ ] 🛠️ Other

## Checklist

- [x] Tests added
- [x] Documented the changes

***

This PR cleans up the deprecations introduced previously, and it makes sure no warnings are triggered during the tests.

There is still some Numba code using reflected lists (which are deprecated); that needs to be refactored.

### DIFF
--- a/examples/sparse_finch.ipynb
+++ b/examples/sparse_finch.ipynb
@@ -24,7 +24,7 @@
     "import os\n",
     "\n",
     "os.environ[\"SPARSE_BACKEND\"] = \"Finch\"\n",
-    "CI_MODE = os.getenv(\"CI_MODE\", default=False)"
+    "CI_MODE = bool(int(os.getenv(\"CI_MODE\", default=\"0\")))"
    ]
   },
   {
@@ -42,9 +42,7 @@
     "\n",
     "import numpy as np\n",
     "import scipy.sparse as sps\n",
-    "import scipy.sparse.linalg as splin\n",
-    "\n",
-    "assert sparse.BackendType.Finch == sparse.BACKEND"
+    "import scipy.sparse.linalg as splin"
    ]
   },
   {
--- a/sparse/numba_backend/_common.py
+++ b/sparse/numba_backend/_common.py
@@ -10,7 +10,7 @@ from numba import literal_unroll
 
 import numpy as np
 
-from ._coo.common import asCOO
+from ._coo import as_coo
 from ._sparse_array import SparseArray
 from ._utils import (
     _zero_of_dtype,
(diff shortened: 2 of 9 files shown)


examples/sparse_finch.ipynb
```diff
@@ -42,9 +42,7 @@
     "\n",
     "import numpy as np\n",
     "import scipy.sparse as sps\n",
-    "import scipy.sparse.linalg as splin\n",
-    "\n",
-    "assert sparse.BackendType.Finch == sparse.BACKEND"
```
mtsokol, line 47: Why do we want to remove that assertion?

sparse/numba_backend/_common.py
```diff
@@ -189,7 +192,12 @@ def tensordot(a, b, axes=2, *, return_type=None):
     oldb = [bs[axis] for axis in notin]
 
     if builtins.any(dim == 0 for dim in chain(newshape_a, newshape_b)):
-        res = asCOO(np.empty(olda + oldb), check=False)
+        from sparse import COO
+
+        dt = (np.empty((), dtype=a.dtype) + np.empty((), dtype=b.dtype)).dtype
```
mtsokol, line 197: Alternatively, this can be done with:
```python
dt = np.result_type(a.dtype, b.dtype)
```

sparse/numba_backend/_coo/core.py
```diff
@@ -1157,52 +1153,6 @@ def squeeze(self, axis=None):
             fill_value=self.fill_value,
         )
 
-    def resize(self, *args, refcheck=True, coords_dtype=np.intp):
```
mtsokol, line 1160: Why do we remove `resize` method?

examples/sparse_finch.ipynb
hameerabbasi: Because changing back ends is a private API, and using private API in an example would encourage its use.

sparse/numba_backend/_coo/core.py
hameerabbasi: It was deprecated for a long time, and doesn't make any sense in the context of sparse arrays.

mtsokol: Thanks for an explanation!

examples/sparse_finch.ipynb
mtsokol: Sure!

sparse/numba_backend/_common.py
hameerabbasi:
```suggestion
        dt = np.result_type(a.dtype, b.dtype)
```

PR #757, opened 2024-08-24
feat, perf: Refactor the PoC to support multiple dtypes

hameerabbasi:

- [x] 💾 Refactor
- [ ] 🪄 Feature
- [ ] 🐞 Bug Fix
- [x] 🔧 Optimization
- [ ] 📚 Documentation
- [ ] 🧪 Test
- [ ] 🛠️ Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

***

- **Refactor to allow arbitrary dtypes.**
- **Do refcounting instead of copies.**
- **Actually test multiple dtypes.**

### DIFF
--- a/sparse/mlir_backend/_common.py
+++ b/sparse/mlir_backend/_common.py
@@ -1,0 +1,9 @@
+import abc
+
+from mlir import ir
+
+
+class MlirType(abc.ABC):
+    @classmethod
+    @abc.abstractmethod
+    def get_mlir_type(cls) -> ir.Type: ...
--- a/sparse/mlir_backend/_common.py
+++ b/sparse/mlir_backend/_common.py
@@ -1,4 +1,5 @@
 import abc
+import functools
 
 from mlir import ir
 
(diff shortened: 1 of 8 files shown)


hameerabbasi:
xref scipy/scipy#21442

hameerabbasi:
Blocked due to scipy/scipy#21442, workaround in #757.

mtsokol:
Right, so the addition operation fails for integer dtype inputs as we use signed/unsigned ones, where apparently `arith.addi` only expects signless integers.

hameerabbasi:
> Right, so the addition operation fails for integer dtype inputs as we use signed/unsigned ones, where apparently `arith.addi` only expects signless integers.

Looks like we need to add casts, would you have time to do that?

mtsokol:
Looks that it's failing for `numpy<2` due to missing `isdtype`. We can either require `>2.0` or check it with `isinstance`?

mtsokol (approved):
LGTM! @hameerabbasi so now we don't differentiate between signed and unsigned ints in mlir backend, right?

hameerabbasi:
> LGTM! @hameerabbasi so now we don't differentiate between signed and unsigned ints in mlir backend, right?

There are a few ops where we need to, but we need to use different ops in those cases. For example, [`arith.minsi`](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithminsi-arithminsiop) for signed and [`arith.minui`](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithminui-arithminuiop) for unsigned.

PR #761, opened 2024-08-29
test: Converting tensordot benchmark to run with CodSpeed

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

- Related issue #
- Closes #

## Checklist

- [x] Tests added
- [ ] Documented the changes

***

The sides of the tensors are small, but when using 200, 500, 1000 (please see lines 17 - 19) the tests took a long time locally. More than 25 min.  But I don't know for how much longer because my machine never finished the tests (so I Ctrl-break to stop). Please let me know if you have an idea of the sizes I should use.


hameerabbasi (changes requested):
LGTM besides the one comment; thanks @DeaMariaLeon!

hameerabbasi:
Seems like an out of memory error again... I would reduce the scope of the PyTest fixtures to `test`, or maybe `module` at best.

benchmarks/test_tensordot.py
```diff
@@ -14,7 +14,11 @@ def get_sides_ids(param):
     return f"{m=}-{n=}-{p=}-{q=}"
 
 
-@pytest.fixture(params=itertools.product([10, 20, 50], [10, 20, 50], [10, 20, 50], [10, 20, 50]), ids=get_sides_ids)
+@pytest.fixture(
+    params=itertools.product([10, 20, 50], [10, 20, 50], [10, 20, 50], [10, 20, 50]),
```
hameerabbasi, line 18: Uhh, we're doing `3 ** 4 = 81` permutations here. We might need to put this into a list, shuffle it, and pick the first 10 maybe, but the shuffle needs to be deterministic otherwise we have benchmarks disappearing and appearing.

PR #767, opened 2024-09-03
ci: Add doctests to CI again.

hameerabbasi:

- [ ] 💾 Refactor
- [ ] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

- Related issue #
- Closes #

## Checklist

- [x] Tests added
- [x] Documented the changes

***


hameerabbasi:
@willow-ahrens Doctest issue.

willow-ahrens:
using numba from the numba channel (with a statically linked llvmlite) fixes the segfault issue on mac. It appears that the segfault issue is mac-only at this point, as linux supports some special symbol resolution thing that fixed this problem.

Proposed solution: I would like to leave a warning to users on startup in the event that they have tried to load this particular config of packages. I think we could warn in pydata/sparse whenever we see (conda in use) AND (macos) AND (numba was not installed from numba channel)

As a side note: it is odd to me that we are loading both numba and finch at the same time. we might reduce some of this headache by switching between backends as a build option, rather than a runtime option.

willow-ahrens:
I cannot get a reproducer for this in docker, though I have tried. If anyone can get a reproducer for this that would be great, then we can put it in an issue where we describe the problem. If you ssh into the CI runner, even doing "import sparse.finch_backend" will segfault, so we don't need all the pytest stuff to reproduce.

willow-ahrens:
I've put a comment in to pyjuliapackage that we experience an issue with parallel installs, and suggested a fix here: https://github.com/JuliaPy/pyjuliapkg/issues/19

.github/workflows/ci.yml
```diff
@@ -126,6 +131,9 @@ jobs:
         SPARSE_BACKEND: ${{ matrix.backend }}
       run: |
         cd ${GITHUB_WORKSPACE}/array-api-tests
+        if [ "${SPARSE_BACKEND}" = "Finch" ]; then
+            python -c 'import finch'
+        fi
```
mtsokol, line 136: Why do we need these lines?

willow-ahrens:
also, note this comment regarding installation of juliacall and numba on conda: https://github.com/JuliaPy/PythonCall.jl/issues/215#issuecomment-2540274772

mtsokol (approved):
LGTM! Just one question.

mtsokol:
And please remember about a commit squash!

.github/workflows/ci.yml
willow-ahrens: thats the parallel julia install problem: https://github.com/JuliaPy/pyjuliapkg/issues/19

PR #772, opened 2024-09-10
test: working on benchmarks with `Finch` as backend

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

- Related issue #
- Closes #

## Checklist

- [ ] Code follows style guide
- [ ] Tests added
- [ ] Documented the changes

***

Only added `Finch` on `config.py` - (locally I see the names of the backend. But I'm not sure if this renders fine on CodSpeed).
Also, per Hameer instructions I need to see which tests fail with `Finch`. 
I would like to see if the change on `codspeed.yml` works like expected. 
Finally, I have only added the line `if hasattr(sparse, "compiled"): f = sparse.compiled(f)` to the first two benchmarks to make sure I correctly understood.

### DIFF
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -1,10 +1,25 @@
 import pytest
-
+import os
+import sparse
 
 @pytest.fixture
 def seed(scope="session"):
     return 42
 
+
+def get_backend_id(param):
+    backend = param
+    return f"{backend=}"
+
+@pytest.fixture(params=[sparse._BACKEND], autouse=True, ids=get_backend_id)
+def backend(request):
+
+    return request.param
+    
+
+@pytest.fixture
+def min_size(scope="session"):
+    return 100
 
 @pytest.fixture
 def max_size(scope="session"):
--- a/benchmarks/test_benchmark_coo.py
+++ b/benchmarks/test_benchmark_coo.py
@@ -13,17 +13,21 @@ DENSITY = 0.01
 def format_id(format):
     return f"{format=}"
 
-
 @pytest.mark.parametrize("format", ["coo", "gcxs"])
-def test_matmul(benchmark, sides, format, seed, max_size, ids=format_id):
+def test_matmul(benchmark, sides, seed, format, backend, min_size, max_size, ids=format_id):
+    #if backend == sparse._BackendType.Finch:
+    #    pytest.skip()
+    
     m, n, p = sides
-
-    if m * n >= max_size or n * p >= max_size:
+  
+    if m * n >= max_size or n * p >= max_size or m * n <= min_size or n * p <= min_size:
         pytest.skip()
-
+    
     rng = np.random.default_rng(seed=seed)
     x = sparse.random((m, n), density=DENSITY, format=format, random_state=rng)
     y = sparse.random((n, p), density=DENSITY, format=format, random_state=rng)
+    
+    if hasattr(sparse, "compiled"): operator.matmul = sparse.compiled(operator.matmul)
 
     x @ y  # Numba compilation
 
@@ -52,6 +56,9 @@ def elemwise_args(request, seed, max_size):
 @pytest.mark.parametrize("f", [operator.add, operator.mul])
 def test_elemwise(benchmark, f, elemwise_args):
     x, y = elemwise_args
+
+    if hasattr(sparse, "compiled"): f = sparse.compiled(f)
+    
     f(x, y)
 
     @benchmark
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -25,4 +25,9 @@ jobs:
       - name: Run benchmarks
         uses: CodSpeedHQ/action@v3
         with:
-          run: pytest benchmarks/ --codspeed
+          run: SPARSE_BACKEND=Numba pytest benchmarks/ --codspeed
+
+      - name: Run benchmarks
+        uses: CodSpeedHQ/action@v3
+        with:
+          run: SPARSE_BACKEND=Finch pytest benchmarks/ --codspeed


benchmarks/conftest.py
```diff
@@ -6,6 +8,21 @@ def seed(scope="session"):
     return 42
 
 
+def get_backend_id(param):
+    backend = param
+    return f"{backend=}"
+
+
+@pytest.fixture(params=[sparse._BACKEND], autouse=True, ids=get_backend_id)
```
hameerabbasi, line 16:
```suggestion
@pytest.fixture(params=[sparse._BACKEND.value], autouse=True, ids=get_backend_id)
```
So we get a simple string.

benchmarks/test_benchmark_coo.py
```diff
@@ -15,16 +15,22 @@ def format_id(format):
 
 
 @pytest.mark.parametrize("format", ["coo", "gcxs"])
-def test_matmul(benchmark, sides, format, seed, max_size, ids=format_id):
+def test_matmul(benchmark, sides, seed, format, backend, min_size, max_size, ids=format_id):
+    # if backend == sparse._BackendType.Finch:
+    #    pytest.skip()
```
hameerabbasi, line 20: Let's remove commented code.

benchmarks/test_benchmark_coo.py
```diff
@@ -52,6 +58,10 @@ def elemwise_args(request, seed, max_size):
 @pytest.mark.parametrize("f", [operator.add, operator.mul])
 def test_elemwise(benchmark, f, elemwise_args):
     x, y = elemwise_args
+
+    if hasattr(sparse, "compiled"):
+        f = sparse.compiled(f)
```
hameerabbasi, line 63: We need this in every benchmark.

hameerabbasi (changes requested):
Great work @DeaMariaLeon! A few suggestions.

benchmarks/conftest.py
DeaMariaLeon: Done.

benchmarks/test_benchmark_coo.py
DeaMariaLeon: Done.

benchmarks/test_benchmark_coo.py
DeaMariaLeon: Done.

benchmarks/test_benchmark_coo.py
```diff
@@ -15,16 +15,19 @@ def format_id(format):
...
+def test_matmul(benchmark, sides, seed, format, backend, min_size, max_size, ids=format_id):
     m, n, p = sides
 
-    if m * n >= max_size or n * p >= max_size:
+    if m * n >= max_size or n * p >= max_size or m * n <= min_size or n * p <= min_size:
         pytest.skip()
 
     rng = np.random.default_rng(seed=seed)
     x = sparse.random((m, n), density=DENSITY, format=format, random_state=rng)
     y = sparse.random((n, p), density=DENSITY, format=format, random_state=rng)
```
DeaMariaLeon, line 27: All the tests on file `test_benchmark_coo.py` were meant to fail with `Finch`. 
`Finch` `sparse.random` (used to build  `x` & `y`, line 25 and 26) is different, because it doesn't have the `format` argument that is used with `Numba`.

hameerabbasi: @mtsokol How hard would it be to add this to `finch`?

benchmarks/test_benchmark_coo.py
```diff
@@ -15,16 +15,19 @@ def format_id(format):
...
 
+    if hasattr(sparse, "compiled"):
+        operator.matmul = sparse.compiled(operator.matmul)
```
hameerabbasi, line 29:
```suggestion
    f = operator.matmul
    if hasattr(sparse, "compiled"):
        f = sparse.compiled(f)
```

benchmarks/test_benchmark_coo.py
```diff
@@ -15,16 +15,19 @@ def format_id(format):
...
+
     x @ y  # Numba compilation
```
hameerabbasi, line 31:
```suggestion
    f(x, y)  # Compilation
```

benchmarks/test_benchmark_coo.py
hameerabbasi:
```suggestion
    def f(x, y):
        return x @ y
    if hasattr(sparse, "compiled"):
        f = sparse.compiled(f)
```

benchmarks/test_benchmark_coo.py
hameerabbasi: You can probably use `sparse.asarray(..., format=<format>)`

benchmarks/test_benchmark_coo.py
```diff
@@ -15,16 +15,19 @@ def format_id(format):
 
 
 @pytest.mark.parametrize("format", ["coo", "gcxs"])
-def test_matmul(benchmark, sides, format, seed, max_size, ids=format_id):
+def test_matmul(benchmark, sides, seed, format, backend, min_size, max_size, ids=format_id):
     m, n, p = sides
 
-    if m * n >= max_size or n * p >= max_size:
+    if m * n >= max_size or n * p >= max_size or m * n <= min_size or n * p <= min_size:
```
hameerabbasi, line 21:
```suggestion
    if m * n >= max_size or n * p >= max_size or m * n * DENSITY <= min_size or n * p * DENSITY <= min_size:
```

hameerabbasi:
`backend` should be the first argument in all tests.

benchmarks/test_benchmark_coo.py
willow-ahrens: Do we just need a format=nothing default and then a reformatting line inside finch.random?

hameerabbasi: That'd suffice -- if the format isn't what the `format` arg demands, reformat.

willow-ahrens: do we have a way of specifying format currently?

PR #773, opened 2024-09-12
MNT: Move constructors to a direct style

mtsokol:
Hi @hameerabbasi,

This PR moves MLIR constructors to the direct style (with: `sparse-assembler{direct-out=true}`).
I think it's more concise with this approach - `_constructors.py` LOC reduced by 50%, to 250.

Now all Tensors created from NumPy arrays or SciPy sparse arrays are owned and freed by the python side. 

The ones that are results of MLIR ops (like `add` right now) are allocated by MLIR and `owns_memory` in Tensor class is set to `False`. 

I haven't figured out yet how to correctly free MLIR allocated memrefs, (tried with `libc.free`) as I think there's still a memory leak after `sparse.add(tensor_a, tensor_b)`.

### DIFF
--- a/sparse/mlir_backend/_constructors.py
+++ b/sparse/mlir_backend/_constructors.py
@@ -118,7 +118,8 @@ def get_coo_class(values_dtype: type[DType], index_dtype: type[DType]) -> type:
         _index_dtype = index_dtype
 
         @classmethod
-        def from_sps(cls, arr: sps.csr_array) -> "Coo":
+        def from_sps(cls, arr: sps.coo_array) -> "Coo":
+            assert arr.has_canonical_format, "COO must have canonical format"
             np_pos = np.array([0, arr.size], dtype=index_dtype.np_dtype)
             np_coords = np.stack(arr.coords, axis=1, dtype=index_dtype.np_dtype)
             pos = numpy_to_ranked_memref(np_pos)
@@ -131,7 +132,7 @@ def get_coo_class(values_dtype: type[DType], index_dtype: type[DType]) -> type:
 
             return coo_instance
 
-        def to_sps(self, shape: tuple[int, ...]) -> sps.csr_array:
+        def to_sps(self, shape: tuple[int, ...]) -> sps.coo_array:
             pos = ranked_memref_to_numpy(self.pos)
             coords = ranked_memref_to_numpy(self.coords)[pos[0] : pos[1]]
             data = ranked_memref_to_numpy(self.data)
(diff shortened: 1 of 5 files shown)


sparse/mlir_backend/_constructors.py
```diff
@@ -468,35 +204,50 @@ def _is_numpy_obj(x) -> bool:
...
+################
+
+
+class Tensor:
+    def __init__(self, obj, shape=None) -> None:
+        self.shape = shape if shape is not None else obj.shape
+        self.values_dtype = asdtype(obj.dtype)
+
+        if _is_scipy_sparse_obj(obj):
+            self.owns_memory = True
```
hameerabbasi, line 222:
```suggestion
            self.owns_memory = False
```

sparse/mlir_backend/_constructors.py
```diff
@@ -468,35 +204,50 @@ def _is_numpy_obj(x) -> bool:
...
+                self.obj = self.format_class.from_sps(obj)
+            elif obj.format == "coo":
+                index_dtype = asdtype(obj.coords[0].dtype)
+                self.format_class = get_coo_class(self.values_dtype, index_dtype)
+                self.obj = self.format_class.from_sps(obj)
+            else:
+                raise Exception(f"{obj.format} SciPy format not supported.")
+
+        elif _is_numpy_obj(obj):
+            self.owns_memory = True
```
hameerabbasi, line 236:
```suggestion
            self.owns_memory = False
```

sparse/mlir_backend/_constructors.py
```diff
@@ -468,35 +204,50 @@ def _is_numpy_obj(x) -> bool:
...
+            self.format_class = get_dense_class(self.values_dtype, obj.ndim)
+            self.obj = self.format_class.from_sps(obj)
+
+        elif _is_mlir_obj(obj):
+            self.owns_memory = False
```
hameerabbasi, line 241:
```suggestion
            self.owns_memory = True
```

hameerabbasi (changes requested):
Thanks @mtsokol! LGTM!

1. Let's make all members of `Tensor` private.
2. Let's store a list of `free`-able `memref` `str`s. So, for example, in `to_sps` let's store `["data"]`  for `dense`, then we will need to call `libc.free` on `data.allocated`. This will apply when we take ownership from MLIR only.
3. We need to store a reference to NumPy/SciPy objects when we create `Tensor`s from them, as we were doing before. The pattern for `_hold_self_ref_in_ret` will apply still in this case.

sparse/mlir_backend/_constructors.py
mtsokol: Done!

sparse/mlir_backend/_constructors.py
mtsokol: Done!

sparse/mlir_backend/_constructors.py
mtsokol: Done!

mtsokol:
@hameerabbasi I applied all your review comments.

sparse/mlir_backend/_constructors.py
```diff
@@ -468,35 +218,58 @@ def _is_numpy_obj(x) -> bool:
...
+        elif _is_mlir_obj(obj):
+            self._owns_memory = True
+            self._format_class = type(obj)
+            self._obj = obj
+
+            def finalizer(ptrs):
+                for ptr in ptrs:
+                    freeme(ptr)
+
+            weakref.finalize(self, finalizer, obj.get__fields_())
```
hameerabbasi, line 264: This can be moved to `__del__`.

hameerabbasi (changes requested):
A simplification to the memory logic: We can do most of this in `__del__`, check for `self._owns_memory` and then call `libc.free` on the constituent `memref`s if so.

sparse/mlir_backend/_constructors.py
```diff
@@ -1,463 +1,213 @@
...
-        ptr: ctypes.c_void_p,
-        shape: list[int],
-        values_dtype: type[DType],
-        index_dtype: type[DType],
-    ) -> np.ndarray:
-        class Dense(ctypes.Structure):
-            _fields_ = [
-                ("data", rt.make_nd_memref_descriptor(1, values_dtype.to_ctype())),
-                ("data_len", np.ctypeslib.c_intp),
+def freeme(obj: ctypes.Structure) -> None:
```
hameerabbasi, line 42: Maybe rename this to `free_memref`?

sparse/mlir_backend/_constructors.py
mtsokol: Done!

sparse/mlir_backend/_constructors.py
mtsokol: Done!

PR #775, opened 2024-09-16
ENH: CSC and CSF formats for MLIR backend

mtsokol:
Hi @hameerabbasi,

This PR adds support for CSC and CSF formats.


hameerabbasi (approved):
Cool! Just one more step to unify csx/csf: Rank support. Just take a rank and an order, and generate the class.

Doesn't need to be in this PR though.

PR #779, opened 2024-09-24
ENH: `__getitem__` logic for MLIR backend

mtsokol:
Hi @hameerabbasi,

This PR adds `__getitem__` logic so that `tensor[:, :, ...]` can be run. The current version preserves rank (and format).

For now unfortunately it's blocked by https://discourse.llvm.org/t/illegal-operation-when-slicing-csr-csc-coo-tensor/81404 and I'm not sure if SparseTensor dialect fully supports slices.

An interesting case is for example `tensor[:, :]` which just returns `tensor` but our ownership mechanism sees it as MLIR allocated object, where in the meantime it's still SciPy/NumPy that was passed in. I think the mechanism requires a tweak where calling MLIR ops (reshape, slices, elemwise) should also tell if it's MLIR allocated (thus requires a `free`) or just a reference to what was passed (SciPy/NumPy managed arrays).


hameerabbasi (commented):
It seems the only test added here is skipped, due to a bug in LLVM. Let's keep this PR open until it adds new working functionality.

PR #781, opened 2024-09-27
ENH: Add counting triangles example

mtsokol:
Hi @hameerabbasi,

Here's a short example that runs counting triangles with Finch, SciPy, and NetworkX. Two comments:
- The Finch implementation is the slowest right now, it's only run in eager mode as lazy mode is blocked by https://github.com/willow-ahrens/Finch.jl/issues/615
- I see that examples in `examples/` directory aren't executed in the CI job anymore, is it intended? I think they should still be executed.

### DIFF
--- a/examples/triangles_example.py
+++ b/examples/triangles_example.py
@@ -1,0 +1,53 @@
+import importlib
+import os
+
+import sparse
+
+import networkx as nx
+from utils import benchmark
+
+import numpy as np
+
+ITERS = 3
+
+
+if __name__ == "__main__":
+    print("Counting Triangles Example:\n")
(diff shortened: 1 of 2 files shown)


hameerabbasi:
> I see that examples in `examples/` directory aren't executed in the CI job anymore, is it intended? I think they should still be executed.

I believe that benchmarks should be moved to CodSpeed benchmarks, see #772, but examples with small datasets should be executed without measuring timings.

mtsokol:
Ok, then I guess this PR is ready anyway.

hameerabbasi (approved):

PR #782, opened 2024-09-30
ENH: Implement `broadcast_to` function

mtsokol:
Hi @hameerabbasi,

This PR adds [`broadcast_to`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.broadcast_to.html) Array API function.

It it only broadcasts missing dimensions, for `1`-size dimension it can't broadcast as it's a limitation of `linalg.broadcast` in MLIR.

### DIFF
--- a/sparse/mlir_backend/__init__.py
+++ b/sparse/mlir_backend/__init__.py
@@ -16,11 +16,13 @@ from ._dtypes import (
 )
 from ._ops import (
     add,
+    broadcast_to,
     reshape,
 )
 
 __all__ = [
     "add",
+    "broadcast_to",
     "asarray",
     "asdtype",
     "reshape",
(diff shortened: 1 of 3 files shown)


hameerabbasi:
I'm hesitant about merging this as it contains lots of heuristics to figure out the output format -- not to mention there should be a "broadcast level" that's the equivalent of a "zero stride", or perhaps we can ignore the broadcast indices on the Python side.

hameerabbasi:
> What do you mean by a "broadcast level"?

Something that represents a zero-stride, or an ignored index.

mtsokol:
Can you give an Array API `broadcast_to(...)` example call that uses ignored index?

mtsokol:
Decided that for now `broadcast_to` is a copying operation and Tensor views should be included on the MLIR or Finch Dialect side.

hameerabbasi (approved):
Thanks, @mtsokol!

hameerabbasi:
Force-merged due to the CI hiccup. Thanks, @mtsokol!

PR #784, opened 2024-10-01
ENH: Support 3D COO and cross-rank `reshape`

mtsokol:
Hi @hameerabbasi,

This PR:
- Enables COO of rank higher than 2.
- Improves `reshape` function so that it allows reshaping with 1-size dimensions across different ranks, e.g. `(8, 1, 2)` into `(8, 2)`.

For `reshaping` across different ranks we need some kind of logic for selecting new sparse format. I implemented the simplest option in `_infer_format_class`, but I guess we can come up with smarter logic.

### DIFF
--- a/sparse/mlir_backend/_common.py
+++ b/sparse/mlir_backend/_common.py
@@ -12,6 +12,10 @@ class MlirType(abc.ABC):
     def get_mlir_type(cls) -> ir.Type: ...
 
 
+class RefableList(list):
+    pass
+
+
 def fn_cache(f, maxsize: int | None = None):
     return functools.wraps(f)(functools.lru_cache(maxsize=maxsize)(f))
 
--- a/sparse/mlir_backend/_common.py
+++ b/sparse/mlir_backend/_common.py
@@ -2,6 +2,7 @@ import abc
 import ctypes
 import functools
 import weakref
+from dataclasses import dataclass
 
 from mlir import ir
 
@@ -12,8 +13,18 @@ class MlirType(abc.ABC):
     def get_mlir_type(cls) -> ir.Type: ...
 
 
-class RefableList(list):
-    pass
+@dataclass
+class PackedArgumentTuple:
+    contents: tuple
+
+    def __getitem__(self, index):
+        return self.contents[index]
+
+    def __iter__(self):
+        yield from self.contents
+
+    def __len__(self):
+        return len(self.contents)
 
 
 def fn_cache(f, maxsize: int | None = None):
--- a/sparse/mlir_backend/_constructors.py
+++ b/sparse/mlir_backend/_constructors.py
@@ -8,7 +8,7 @@ from mlir.dialects import sparse_tensor
 import numpy as np
 import scipy.sparse as sps
 
-from ._common import _hold_self_ref_in_ret, _take_owneship, fn_cache
+from ._common import RefableList, _hold_self_ref_in_ret, _take_owneship, fn_cache
 from ._core import ctx, libc
 from ._dtypes import DType, asdtype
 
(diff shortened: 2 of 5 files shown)


sparse/mlir_backend/_constructors.py
```diff
@@ -118,26 +118,31 @@ class Coo(ctypes.Structure):
         _index_dtype = index_dtype
 
         @classmethod
-        def from_sps(cls, arr: sps.coo_array) -> "Coo":
-            assert arr.has_canonical_format, "COO must have canonical format"
-            np_pos = np.array([0, arr.size], dtype=index_dtype.np_dtype)
-            np_coords = np.stack(arr.coords, axis=1, dtype=index_dtype.np_dtype)
+        def from_sps(cls, arr: sps.coo_array | np.ndarray) -> "Coo":
```
hameerabbasi, line 121:
```suggestion
        def from_sps(cls, arr: sps.coo_array | tuple[np.ndarray, np.ndarray, np.ndarray]) -> "Coo":
```

sparse/mlir_backend/_constructors.py
```diff
@@ -118,26 +118,31 @@ class Coo(ctypes.Structure):
...
+            if isinstance(arr, sps.coo_array):
+                assert arr.has_canonical_format, "COO must have canonical format"
```
hameerabbasi, line 123: Let's convert this to a `if-raise`

sparse/mlir_backend/_constructors.py
```diff
@@ -118,26 +118,31 @@ class Coo(ctypes.Structure):
...
+                np_pos = np.array([0, arr.size], dtype=index_dtype.np_dtype)
+                np_coords = np.stack(arr.coords, axis=1, dtype=index_dtype.np_dtype)
+                np_data = arr.data
+            else:
+                assert len(arr) == 3, "COO must be comprised of three arrays"
```
hameerabbasi, line 128: Same here, `if-raise`

sparse/mlir_backend/_constructors.py
```diff
@@ -159,8 +164,13 @@ def get_tensor_definition(cls, shape: tuple[int, ...]) -> ir.RankedTensorType:
...
                 )
-                levels = (compressed_lvl, sparse_tensor.LevelFormat.singleton)
-                ordering = ir.AffineMap.get_permutation([0, 1])
+                mid_singleton_lvls = [
+                    sparse_tensor.EncodingAttr.build_level_type(
+                        sparse_tensor.LevelFormat.singleton, [sparse_tensor.LevelProperty.non_unique]
+                    )
+                ] * (len(shape) - 2)
+                levels = (compressed_lvl, *mid_singleton_lvls, sparse_tensor.LevelFormat.singleton)
+                ordering = ir.AffineMap.get_permutation([*range(len(shape))])
```
hameerabbasi, line 173: Nice logic!

sparse/mlir_backend/_common.py
```diff
@@ -12,6 +12,10 @@ class MlirType(abc.ABC):
     def get_mlir_type(cls) -> ir.Type: ...
 
 
+class RefableList(list):
```
hameerabbasi, line 15: Rename to `PackedArgumentTuple(tuple)`?

hameerabbasi (changes requested):
Some changes -- it'd be nice to give CSF this same treatment, not for this PR though.

sparse/mlir_backend/_common.py
mtsokol: Unfortunately we can't subclass `tuple` here. This class is used in `weakref.finalize(...)` and the `weakref` docs say:

> Several built-in types such as [list](https://docs.python.org/3/library/stdtypes.html#list) and [dict](https://docs.python.org/3/library/stdtypes.html#dict) do not directly support weak references but can add support through subclassing:
> ```python
> class Dict(dict): 
>     pass
> obj = Dict(red=1, green=2, blue=3)   # this object is weak referenceable
> ```
> CPython implementation detail: Other built-in types such as [tuple](https://docs.python.org/3/library/stdtypes.html#tuple) and [int](https://docs.python.org/3/library/functions.html#int) do not support weak references even when subclassed.

So it must be a `list` subclass. Would you still like to change its name?

hameerabbasi: That sounds okay, but let's `s/List/Tuple` instead.

sparse/mlir_backend/_constructors.py
hameerabbasi: Or maybe `Iterable[np.ndarray]`?

sparse/mlir_backend/_common.py
mtsokol: What do you mean by `s/Tuple/List`? Subclass `List` from `typing`?

hameerabbasi: I was initially thinking of renaming, but how about composition (using `tuple`) instead of inheritance? It will solve the `weakref` issue; and be a bit more future proof. If needed, we can define `__getitem__`, `__len__` and `__iter__` for ease of usage. And let's use a `dataclass`.

sparse/mlir_backend/_constructors.py
mtsokol: Done!

sparse/mlir_backend/_constructors.py
mtsokol: Done!

sparse/mlir_backend/_constructors.py
mtsokol: Done!

sparse/mlir_backend/_common.py
mtsokol: Sure! I moved to a `@dataclass` here, and implemented functions required for the iteration. I also renamed it to `PackedArgumentTuple`.

hameerabbasi (approved):

hameerabbasi:
Nice work!

PR #787, opened 2024-10-03
ENH: Update MLIR backend to LLVM 20.dev

mtsokol:
Hi @hameerabbasi,

This PR updates MLIR backend to current LLVM 20.dev (so `main` branch) **experimentally**:

- I ran it locally against latest LLVM version.
- Moved COO format to SoA convention [link](https://discourse.llvm.org/t/passmanager-fails-on-simple-coo-addition-example/81247/2?u=mtsokol).
- Added `format` attribute to Tensor class.
- Updated `tensor.empty` call [link](https://github.com/llvm/llvm-project/pull/110656).

As you can see it fixes a bunch of skips in the test suite: [sparse/mlir_backend/tests/test_simple.py](https://github.com/pydata/sparse/compare/llvm-nightly-test?expand=1#diff-69c60a2345c3e317169e806992fbea9202647744064406623584ea8e96cad54c)

- Dense+Dense and COO+COO now works.
- Reshaping Dense and COO formats also works.

It's thanks to changes already present in `main` branch, added after `19.x` branched, and:
- https://github.com/llvm/llvm-project/pull/108615
- https://github.com/llvm/llvm-project/pull/109135
- https://github.com/llvm/llvm-project/pull/110656
- https://github.com/llvm/llvm-project/pull/110882

Maybe one thing that we can consider is to have a `nightly` label for MLIR Python bindings feedstock some time in the future? Otherwise, I guess `20.x` will branch in ~6months?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,30 +2,30 @@ defaults:
   run:
     shell: bash -leo pipefail {0}
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+# concurrency:
+#   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+#   cancel-in-progress: true
 
 jobs:
   test:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python: ['3.10', '3.11', '3.12']
+        python: ['3.10'] # , '3.11', '3.12'
         pip_opts: ['']
         numba_boundscheck: [0]
-        include:
-          - os: macos-latest
-            python: '3.10'
-          - os: windows-latest
-            python: '3.10'
-          - os: ubuntu-latest
-            python: '3.10'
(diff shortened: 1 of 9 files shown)


hameerabbasi:
I think the best option is backporting these PRs as they are blockers and real issues.

mtsokol:
I asked about it here: https://github.com/llvm/llvm-project/pull/109135#issuecomment-2391063246.

But reshaping and adding Dense formats will stay broken until first 20 RC as it happens to be fixed in the current `main`.

hameerabbasi:
In that case we will need to build LLVM in CI (which is quite complicated) and it absolutely needs to be cached.

mtsokol:
> In that case we will need to build LLVM in CI (which is quite complicated) and it absolutely needs to be cached.

Right, in https://github.com/pydata/sparse/pull/799 I added caching with `ccache` and LLVM builds in 10min instead of 2h.

hameerabbasi:
Closing in favour of #799.

PR #791, opened 2024-10-14
ENH: Add `sparse_vector` constructor

mtsokol:
Hi @hameerabbasi,

I started porting [Array API vecdot function](https://data-apis.org/array-api/latest/API_specification/generated/array_api.vecdot.html) to Finch MLIR backend. So far it's 1D only so here's a SparseVector format that is needed for it.


hameerabbasi (approved):
Thanks, I'm working on a more defined way of putting together levels and ops in a lazy fashion.

Issue #797, opened 2024-10-22
Add SOA level property and change COO format once in LLVM


hameerabbasi:
Once we update to LLVM 20/nightly builds we will change COO format to SoA (Structure of Arrays) calling convention, as it fixes some issues for us [link](https://discourse.llvm.org/t/passmanager-fails-on-simple-coo-addition-example/81247/2?u=mtsokol).

```mlir
#COO = #sparse_tensor.encoding<{
    map = (i, j) -> (i : compressed(nonunique), j : singleton(soa)), posWidth = 64, crdWidth = 64
}>
#COO_3D = #sparse_tensor.encoding<{
    map = (i, j, k) -> (i : compressed(nonunique), j : singleton(soa, nonunique), k : singleton(soa)), posWidth = 64, crdWidth = 64
}>
#COO_4D = #sparse_tensor.encoding<{
    map = (i, j, k, l) -> (i : compressed(nonunique), j : singleton(soa, nonunique), k : singleton(soa, nonunique), l : singleton(soa)), posWidth = 64, crdWidth = 64
}>
```

Then instead of:
```python
fields.append((f"indices_{compressed_counter}", get_nd_memref_descr(2, idx_dtype)))
```
we would have:
```python
fields.append((f"indices_coo_dim_0_{compressed_counter}", get_nd_memref_descr(1, idx_dtype)))
fields.append((f"indices_coo_dim_1_{compressed_counter}", get_nd_memref_descr(1, idx_dtype)))
coo_dim_counter = 2
...
if LevelFormat.Singleton == level.format:
    fields.append((f"indices_coo_dim_{coo_dim_counter}_{compressed_counter}", get_nd_memref_descr(1, idx_dtype)))
    coo_dim_counter += 1
```

Does it make sense?

But this will be addressed separately once we can build with nightlies.

_Originally posted by @mtsokol in https://github.com/pydata/sparse/pull/792#discussion_r1810682069_

PR #799, opened 2024-10-23
ENH: Update MLIR backend to LLVM 20.dev

mtsokol:
This PR replaces https://github.com/pydata/sparse/pull/787 and fixes https://github.com/pydata/sparse/issues/797

---

This PR updates MLIR backend to current LLVM 20.dev (so `main` branch):

- I ran it locally against latest LLVM version.
- Moved COO format to SoA convention [link](https://discourse.llvm.org/t/passmanager-fails-on-simple-coo-addition-example/81247/2?u=mtsokol).
- Updated `tensor.empty` call [link](https://github.com/llvm/llvm-project/pull/110656).

As one can see it fixes a bunch of skips in the test suite: [sparse/mlir_backend/tests/test_simple.py](https://github.com/pydata/sparse/compare/llvm-nightly-test?expand=1#diff-69c60a2345c3e317169e806992fbea9202647744064406623584ea8e96cad54c)

- Dense+Dense and COO+COO now works.
- Reshaping Dense and COO formats also works.

It's thanks to changes already present in `main` branch, added after `19.x` branched, and:
- https://github.com/llvm/llvm-project/pull/108615
- https://github.com/llvm/llvm-project/pull/109135
- https://github.com/llvm/llvm-project/pull/110656
- https://github.com/llvm/llvm-project/pull/110882

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,10 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
-      - uses: mamba-org/setup-micromamba@v2
+      - uses: mamba-org/setup-micromamba@v1.9.0
         with:
+          # NOTE: https://github.com/mamba-org/setup-micromamba/issues/227
+          micromamba-version: '1.5.10-0'
           environment-file: ci/environment.yml
           init-shell: >-
             bash
--- a/sparse/mlir_backend/_conversions.py
+++ b/sparse/mlir_backend/_conversions.py
@@ -94,13 +94,17 @@ def _from_scipy(arr: ScipySparseArray, copy: bool | None = None) -> Array:
         case "coo":
             if copy is not None and not copy:
                 raise RuntimeError(f"`scipy.sparse.{type(arr.__name__)}` cannot be zero-copy converted.")
-            coords = np.stack([arr.row, arr.col], axis=1)
+            row, col = arr.row, arr.col
+            if row.dtype != col.dtype:
+                raise RuntimeError(f"`row` and `col` dtypes must be the same: {row.dtype} != {col.dtype}.")
             pos = np.array([0, arr.nnz], dtype=np.int64)
             pos_width = pos.dtype.itemsize * 8
-            crd_width = coords.dtype.itemsize * 8
+            crd_width = row.dtype.itemsize * 8
             data = arr.data
             if copy:
-                data = arr.data.copy()
+                data = data.copy()
+                row = row.copy()
+                col = col.copy()
 
             level_props = LevelProperties(0)
             if not arr.has_canonical_format:
@@ -109,7 +113,7 @@ def _from_scipy(arr: ScipySparseArray, copy: bool | None = None) -> Array:
             coo_format = get_storage_format(
                 levels=(
                     Level(LevelFormat.Compressed, level_props | LevelProperties.NonUnique),
-                    Level(LevelFormat.Singleton, level_props),
+                    Level(LevelFormat.Singleton, level_props | LevelProperties.SOA),
                 ),
                 order=(0, 1),
                 pos_width=pos_width,
@@ -117,7 +121,7 @@ def _from_scipy(arr: ScipySparseArray, copy: bool | None = None) -> Array:
                 dtype=arr.dtype,
             )
 
-            return from_constituent_arrays(format=coo_format, arrays=(pos, coords, data), shape=arr.shape)
+            return from_constituent_arrays(format=coo_format, arrays=(pos, row, col, data), shape=arr.shape)
         case _:
             raise NotImplementedError(f"No conversion implemented for `scipy.sparse.{type(arr.__name__)}`.")
 
@@ -133,8 +137,8 @@ def to_scipy(arr: Array) -> ScipySparseArray:
                 return sps.csr_array((data, indices, indptr), shape=arr.shape)
             return sps.csc_array((data, indices, indptr), shape=arr.shape)
         case (Level(LevelFormat.Compressed, _), Level(LevelFormat.Singleton, _)):
-            _, coords, data = arr.get_constituent_arrays()
-            return sps.coo_array((data, (coords[:, 0], coords[:, 1])), shape=arr.shape)
+            _, row, col, data = arr.get_constituent_arrays()
+            return sps.coo_array((data, (row, col)), shape=arr.shape)
         case _:
             raise RuntimeError(f"No conversion implemented for `{storage_format=}`.")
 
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -11,20 +11,20 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python: ['3.10', '3.11', '3.12']
+        python: ['3.10']  # , '3.11', '3.12'
```
mtsokol, line 14: `Finch-mlir` CI build needs to support other oses, archs, and Python versions.

mtsokol:
We still need more wheels for `finch-milr` package on PyPI but the PR itself can be already reviewed.

sparse/mlir_backend/_core.py
```diff
@@ -2,14 +2,23 @@
...
 
+finch_lib_path = f"{sys.prefix}/lib/python3.{sys.version_info.minor}/site-packages/lib"
+
+ld_library_path = os.environ.get("LD_LIBRARY_PATH")
+ld_library_path = f"{finch_lib_path}:{ld_library_path}" if ld_library_path is None else finch_lib_path
+os.environ["LD_LIBRARY_PATH"] = ld_library_path
+
 MLIR_C_RUNNER_UTILS = ctypes.util.find_library("mlir_c_runner_utils")
+if os.name == "posix":
+    MLIR_C_RUNNER_UTILS = f"{finch_lib_path}/{MLIR_C_RUNNER_UTILS}"
```
hameerabbasi, line 21: This needs something for Windows.

sparse/mlir_backend/levels.py
```diff
@@ -108,15 +109,28 @@ def _get_ctypes_type(self, *, owns_memory=False):
...
+                        )
                     else:
                         fields.append((f"indices_{compressed_counter}", get_nd_memref_descr(1, idx_dtype)))
 
+                if LevelFormat.Singleton == level.format:
+                    singleton_counter += 1
+                    fields.append(
+                        (f"indices_{compressed_counter}_coords_{singleton_counter}", get_nd_memref_descr(1, idx_dtype))
+                    )
+
```
hameerabbasi, line 133: This should probably handle SOA and without SOA separately.

mtsokol: In my opinion we should only support SoA singleton format:
1. Non-SoA singleton looks to be buggy for basic operations [link](https://discourse.llvm.org/t/passmanager-fails-on-simple-coo-addition-example/81247/2?u=mtsokol)
2. Mixed singleton levels aren't allowed: https://github.com/llvm/llvm-project/blob/8d38fbf2f027c72332c8ba03ff0ff0f83b4dcf02/mlir/lib/Dialect/SparseTensor/IR/SparseTensorDialect.cpp#L811

What would be the benefit of supporting non-SoA singleton levels separately?

.github/workflows/ci.yml
mtsokol: Done! All OSes and archs that we need right now for this PR work.

sparse/mlir_backend/_conversions.py
```diff
@@ -94,13 +94,15 @@ def _from_scipy(arr: ScipySparseArray, copy: bool | None = None) -> Array:
         case "coo":
             if copy is not None and not copy:
                 raise RuntimeError(f"`scipy.sparse.{type(arr.__name__)}` cannot be zero-copy converted.")
-            coords = np.stack([arr.row, arr.col], axis=1)
+            row, col = arr.row, arr.col
```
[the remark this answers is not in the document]
mtsokol: Done!

sparse/mlir_backend/levels.py
hameerabbasi: We'd be able to support the current COO format only for non-SoA.

mtsokol: What do you mean by that? Can you give some example?
With this PR we can support COO format (also input object from `scipy.sparse.coo_array`) where MLIR-backend implementation uses SoA representation.

hameerabbasi: The existing Numba `COO` format uses the non-SoA format, and we pretty much have to support this to be backwards compatible.

.github/workflows/ci.yml
```diff
@@ -34,8 +34,10 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
-      - uses: mamba-org/setup-micromamba@v2
```
hameerabbasi, line 37: We can use v2 here right?

hameerabbasi (approved):
Just one comment plus CI/Windows and should be good to go.

.github/workflows/ci.yml
mtsokol: IMO we can't: `v2` failed with: https://github.com/pydata/sparse/actions/runs/11685373919/job/32538684569#step:3:460 which is explained in https://github.com/mamba-org/setup-micromamba/issues/227 (also linked it in the file). Versions used here (for now until v2 is fixed) solve it.

mtsokol: I don't know why an issue reported in September appeared in our CI only now but I will go ahead and merge it with `v1.9`.

mtsokol: I will go ahead and merge it now. I'll open a separate PR with `v2` to track when it's fixed upstream.

mtsokol: Here's a PR that reproduces it: https://github.com/pydata/sparse/pull/805

sparse/mlir_backend/levels.py
mtsokol: I think I'm still missing the point here. Numba is a separate backend that supports only 1D and 2D COO arrays. MLIR backend supports >=2D COO arrays.
What do you mean by backward compatibility here? Can you give an example where backward compatibility breaks here? If a user passes `scipy.sparse.coo_array` object to `sparse.asarray` function it it will work for any backend regardless of an internal representation (SoA or non-SoA).

hameerabbasi: So, `sparse.COO` has a constructor that takes `coords` and a `.coords` attribute. The attribute is a 2D NumPy array.

Ideally, I'd like to keep it a 2D `np.ndarray` as otherwise I'm not 100% sure how much will break.

We could do this with `np.stack`, but that would incur a performance penalty.

mtsokol: I think for accessing `coords`, `np.stack` could be an option. My caveat for non-SoA in MLIR backend is that it already has some issues with basic operations: https://discourse.llvm.org/t/passmanager-fails-on-simple-coo-addition-example/81247

PR #800, opened 2024-10-24
Add back removed ops

hameerabbasi:
- **Fix signed/unsigned integer naming mix-up.**
- **Add mechanism for detecting output format.**

### DIFF
--- a/sparse/mlir_backend/levels.py
+++ b/sparse/mlir_backend/levels.py
@@ -195,3 +195,56 @@ def _get_storage_format(
         crd_width=crd_width,
         dtype=dtype,
     )
+
+
+def _is_sparse_level(lvl: Level | LevelFormat, /) -> bool:
+    if isinstance(lvl, Level):
+        lvl = lvl.format
+    return LevelFormat.Dense != lvl
+
+
+def _count_sparse_levels(format: StorageFormat) -> int:
+    return sum(_is_sparse_level(lvl) for lvl in format.levels)
+
+
+def _determine_levels(*formats: StorageFormat, dtype: DType, union: bool, out_ndim: int | None = None) -> StorageFormat:
+    if len(formats) == 0:
+        if out_ndim is None:
+            out_ndim = 0
+        return get_storage_format(
+            levels=(Level(LevelFormat.Dense),) * out_ndim,
+            order="C",
+            pos_width=64,
+            crd_width=64,
+            dtype=dtype,
+        )
+
+    if out_ndim is None:
+        out_ndim = max(fmt.rank for fmt in formats)
+
+    n_sparse = 0
+    pos_width = 0
+    crd_width = 0
+    op = max if union else min
+    order = ()
+    for fmt in formats:
+        n_sparse = op(n_sparse, _count_sparse_levels(fmt))
+        pos_width = max(pos_width, fmt.pos_width)
+        crd_width = max(crd_width, fmt.crd_width)
+        if order != "C":
+            if fmt.order[: len(order)] == order:
+                order = fmt.order
+            elif order[: len(fmt.order)] != fmt.order:
+                order = "C"
+
+    if out_ndim < n_sparse:
+        n_sparse = out_ndim
+
+    levels = (Level(LevelFormat.Dense),) * (out_ndim - n_sparse) + (Level(LevelFormat.Compressed),) * n_sparse
+    return get_storage_format(
+        levels=levels,
+        order=order,
+        pos_width=pos_width,
+        crd_width=crd_width,
+        dtype=dtype,
+    )
--- a/sparse/mlir_backend/_dtypes.py
+++ b/sparse/mlir_backend/_dtypes.py
@@ -76,10 +76,10 @@ class UnsignedIntegerDType(IntegerDType):
         return np.dtype(getattr(np, f"uint{self.bit_width}"))
 
 
-int8 = UnsignedIntegerDType(bit_width=8)
-int16 = UnsignedIntegerDType(bit_width=16)
-int32 = UnsignedIntegerDType(bit_width=32)
-int64 = UnsignedIntegerDType(bit_width=64)
+uint8 = UnsignedIntegerDType(bit_width=8)
+uint16 = UnsignedIntegerDType(bit_width=16)
+uint32 = UnsignedIntegerDType(bit_width=32)
+uint64 = UnsignedIntegerDType(bit_width=64)
 
 
 @dataclasses.dataclass(eq=True, frozen=True, kw_only=True)
@@ -89,10 +89,10 @@ class SignedIntegerDType(IntegerDType):
         return np.dtype(getattr(np, f"int{self.bit_width}"))
 
 
-uint8 = SignedIntegerDType(bit_width=8)
-uint16 = SignedIntegerDType(bit_width=16)
-uint32 = SignedIntegerDType(bit_width=32)
-uint64 = SignedIntegerDType(bit_width=64)
+int8 = SignedIntegerDType(bit_width=8)
+int16 = SignedIntegerDType(bit_width=16)
+int32 = SignedIntegerDType(bit_width=32)
+int64 = SignedIntegerDType(bit_width=64)
 
 
 intp: SignedIntegerDType = locals()[f"int{_PTR_WIDTH}"]


sparse/mlir_backend/_common.py
```diff
@@ -52,3 +53,13 @@ def finalizer(ptr):
...
+
+
+def as_shape(x) -> tuple[int]:
+    if not isinstance(x, Iterable):
+        x = (x,)
+
+    if not all(isinstance(xi, int) for xi in x):
+        raise TypeError("Shape must be an `int` or tuple of `int`s.")
+
+    return tuple(int(xi) for xi in x)
```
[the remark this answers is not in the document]
hameerabbasi: Because, for example, NumPy integers inherit from `int` (so the `isinstance` check will return `True`); but we need exactly `int` here.

sparse/mlir_backend/_ops.py
```diff
@@ -42,8 +44,13 @@ def add(a, b):
...
                     [out],
-                    ir.ArrayAttr.get([ir.AffineMapAttr.get(p) for p in (ordering,) * 3]),
-                    ir.ArrayAttr.get([ir.Attribute.parse("#linalg.iterator_type<parallel>")] * rank),
+                    ir.ArrayAttr.get(
+                        [
+                            ir.AffineMapAttr.get(ir.AffineMap.get_minor_identity(max_rank, t.rank))
+                            for t in (a_tensor_type, b_tensor_type, out_tensor_type)
+                        ]
+                    ),
+                    ir.ArrayAttr.get([ir.Attribute.parse("#linalg.iterator_type<parallel>")] * out_tensor_type.rank),
```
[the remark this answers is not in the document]
hameerabbasi: Done in https://github.com/pydata/sparse/pull/800/commits/4e0fe38716fb31b7b2b50dd14313baa8428f19b9.

sparse/mlir_backend/_ops.py
```diff
@@ -123,26 +130,46 @@ def broadcast_to(in_tensor):
...
 
 def add(x1: Array, x2: Array) -> Array:
-    ret_storage_format = x1.format
+    ret_storage_format = _determine_format(x1.format, x2.format, dtype=x1.dtype, union=True)
     ret_storage = ret_storage_format._get_ctypes_type(owns_memory=True)()
-    out_tensor_type = ret_storage_format._get_mlir_type(shape=x1.shape)
+    out_tensor_type = ret_storage_format._get_mlir_type(shape=np.broadcast_shapes(x1.shape, x2.shape))
 
-    # TODO: Decide what will be the output tensor_type
+    # TODO: Determine output format via autoscheduler
```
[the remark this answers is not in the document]
hameerabbasi: Done in https://github.com/pydata/sparse/pull/800/commits/4e0fe38716fb31b7b2b50dd14313baa8428f19b9.

sparse/mlir_backend/levels.py
```diff
@@ -195,3 +195,66 @@ def _get_storage_format(
         crd_width=crd_width,
         dtype=dtype,
     )
+
+
+def _is_sparse_level(lvl: Level | LevelFormat, /) -> bool:
+    if isinstance(lvl, Level):
+        lvl = lvl.format
+    return LevelFormat.Dense != lvl
```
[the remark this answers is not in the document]
hameerabbasi: Done in https://github.com/pydata/sparse/pull/800/commits/4e0fe38716fb31b7b2b50dd14313baa8428f19b9. I used an `assert` as this is private API. In a public API I would use `if ... raise`.

sparse/mlir_backend/levels.py
```diff
@@ -195,3 +195,66 @@ def _get_storage_format(
...
+
+def _count_sparse_levels(format: StorageFormat) -> int:
+    return sum(_is_sparse_level(lvl) for lvl in format.levels)
+
+
+def _count_dense_levels(format: StorageFormat) -> int:
+    return sum(not _is_sparse_level(lvl) for lvl in format.levels)
+
+
+def _determine_format(*formats: StorageFormat, dtype: DType, union: bool, out_ndim: int | None = None) -> StorageFormat:
```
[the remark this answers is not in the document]
hameerabbasi: Done in https://github.com/pydata/sparse/pull/800/commits/4e0fe38716fb31b7b2b50dd14313baa8428f19b9.

hameerabbasi:
> As #799 is now merged let's rebase it and also include COO in the `determine_format` function. I think that it could be `COO+COO=COO` and `CSR+CSR=CSR` etc.

Right now `COO+COO=DCSR` (as both levels are sparse for COO, and I'd like to avoid special cases as much as possible). However; I added an `asformat` function to convert between formats if we need to.

sparse/mlir_backend/_common.py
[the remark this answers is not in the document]
mtsokol: For NumPy scalars I get:
```python
In [10]: isinstance(3, int)
Out[10]: True

In [11]: isinstance(np.int64(3), int)
Out[11]: False
```

pixi.toml
```diff
@@ -55,17 +55,20 @@ finch-tensor = ">=0.1.31"
...
 [feature.finch.target.osx-arm64.activation.env]
+SPARSE_BACKEND = "Finch"
 PYTHONFAULTHANDLER = "${HOME}/faulthandler.log"
 
 [feature.mlir.dependencies]
 scipy = ">=0.19"
-mlir-python-bindings = "19.*"
+
+[feature.mlir.target.osx-arm64.pypi-dependencies]
+finch-mlir = ">=0.0.2"
```
mtsokol, line 65: Nitpick: I think we already want to use `finch-mlir` for all OSes/architectures (doesn't need to be in this PR, I'm not using pixi).

sparse/mlir_backend/_ops.py
```diff
@@ -125,26 +132,88 @@ def broadcast_to(in_tensor):
...
+
+        with ir.InsertionPoint(module.body):
+
+            @func.FuncOp.from_py_func(in_tensor_type)
+            def convert(in_tensor):
+                return sparse_tensor.convert(out_tensor_type, in_tensor)
+
+            convert.func_op.attributes["llvm.emit_c_interface"] = ir.UnitAttr.get()
+            if DEBUG:
+                (CWD / "broadcast_to_module.mlir").write_text(str(module))
```
mtsokol, line 154: I think it should be `"convert_module.mlir"`.

sparse/mlir_backend/_ops.py
```diff
@@ -125,26 +132,88 @@ def broadcast_to(in_tensor):
...
+            pm.run(module.operation)
+            if DEBUG:
+                (CWD / "broadcast_to_module_opt.mlir").write_text(str(module))
```
mtsokol, line 157: I think it should be `"convert_module.mlir"`.

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -297,3 +299,97 @@ def test_copy():
...
+        arrs = get_example_csf_arrays(dtype)
+        csf_tensor = sparse.from_constituent_arrays(format=csf_format, arrays=arrs, shape=shape)
+
+        result = sparse.reshape(csf_tensor, shape=new_shape)
+        for actual, expected in zip(result.get_constituent_arrays(), expected_arrs, strict=True):
+            np.testing.assert_array_equal(actual, expected)
+
+    # DENSE
+    # NOTE: dense reshape is probably broken in MLIR in 19.x branch
+    # dense = np.arange(math.prod(SHAPE), dtype=dtype).reshape(SHAPE)
```
mtsokol, line 395: As were are using our wheels (and .20dev LLVM version) we should uncomment and test this case.

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -297,3 +299,97 @@ def test_copy():
...
     np.testing.assert_array_equal(sparse.to_numpy(arr_sp3), arr_np_copy)
+
+
+@parametrize_dtypes
+@pytest.mark.parametrize(
+    "format",
+    [
+        "csr",
+        pytest.param("csc", marks=pytest.mark.xfail(reason="https://github.com/llvm/llvm-project/pull/109641")),
+        pytest.param("coo", marks=pytest.mark.xfail(reason="https://github.com/llvm/llvm-project/pull/109135")),
```
mtsokol, line 310: This one should be passing - `LevelProperty.soa` is available in our wheels.

sparse/mlir_backend/_ops.py
```diff
@@ -125,26 +132,88 @@ def broadcast_to(in_tensor):
...
         "add",
         ctypes.pointer(ctypes.pointer(ret_storage)),
         *x1._to_module_arg(),
         *x2._to_module_arg(),
     )
-    return Array(storage=ret_storage, shape=out_tensor_type.shape)
+    return Array(storage=ret_storage, shape=tuple(out_tensor_type.shape))
+
+
+def asformat(x: Array, /, format: StorageFormat) -> Array:
```
mtsokol, line 183: I would like to have (even simple) `asformat` tests in `test_simple` in a separate `def test_asformat()`, as for me `asformat` logic isn't that trivial. I would just test `COO, CSR`, `Dense, CSF`, etc. with different `union` arg.

sparse/mlir_backend/levels.py
```diff
@@ -209,3 +209,79 @@ def _get_storage_format(
...
+    """Determines the output format from a group of input formats.
+
+    1. Counts the sparse levels for `union=True`, and dense ones for `union=False`.
+    2. Gets the max number of counted levels for each format.
+    3. Constructs a format with the same number of counted levels.
+       Sparse levels are replaced with `LevelFormat.Compressed`.
```
mtsokol, line 235: I would expand `3.` a bit: 
```
    3. Constructs a format with rank of `out_ndim` (max rank of inputs is taken if it's `None`).
       If `union=False` counted levels is the number of sparse levels, otherwise dense.
       Sparse levels are replaced with `LevelFormat.Compressed`.
```

mtsokol (changes requested):
The progress looks good! One more round of comments.

pixi.toml
hameerabbasi: There seems to be a bug in the `pixi` solver that prevents the installation on Linux.

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -42,8 +42,13 @@ def assert_csx_equal(
...
     actual.sum_duplicates()
 
-    np.testing.assert_array_equal(expected.indptr, actual.indptr)
-    np.testing.assert_array_equal(expected.indices, actual.indices)
+    if expected.format != "coo":
+        np.testing.assert_array_equal(expected.indptr, actual.indptr)
+        np.testing.assert_array_equal(expected.indices, actual.indices)
+    else:
+        np.testing.assert_array_equal(expected.row, actual.col)
+        np.testing.assert_array_equal(expected.row, actual.col)
```
mtsokol, line 50: I think you mean:
```python
        np.testing.assert_array_equal(expected.col, actual.col)
        np.testing.assert_array_equal(expected.row, actual.row)
```
right?

mtsokol (changes requested):
I think it's the last round of review - I think `test_reshape` should be refined to cover most of reasonable scenarios.

And let's push another commit instead of force push.

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -297,3 +304,139 @@ def test_copy():
...
+@pytest.mark.parametrize(
+    ("shape", "new_shape"),
+    [
+        ((100, 50), (25, 200)),
+        ((100, 50), (10, 500, 1)),
+        ((80, 1), (8, 10)),
+        ((80, 1), (80,)),
+    ],
+)
+def test_reshape(rng, dtype, format, shape, new_shape):
```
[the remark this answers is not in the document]
hameerabbasi: I've converted to an equivalent dense format and compared via NumPy here.

sparse/mlir_backend/tests/test_simple.py
hameerabbasi: Done!

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -297,3 +304,139 @@ def test_copy():
...
+        assert actual_np.dtype == expected.dtype
+        np.testing.assert_equal(actual_np, expected)
+
+
+@pytest.mark.parametrize(
+    "src_fmt", ["csr", "csc", pytest.param("coo", marks=pytest.mark.skip(reason="TODO: Report MLIR issue"))]
+)
+@pytest.mark.parametrize(
+    "dst_fmt", ["csr", "csc", pytest.param("coo", marks=pytest.mark.skip(reason="TODO: Report MLIR issue"))]
+)
```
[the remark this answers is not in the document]
hameerabbasi: I've linked to it.

sparse/mlir_backend/tests/test_simple.py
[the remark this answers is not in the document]
hameerabbasi: > Also, no tests in `COO` format are run, but they should (reshaping COO results in DCSR) - I think it should be COO?

This is deliberate -- As mentioned we *never* want to guarantee the output format. If a user needs a specific format they should request it.

mtsokol (approved):
LGTM!

hameerabbasi:
Thanks for the thorough review, @mtsokol!

PR #805, opened 2024-11-07
BLD: Revert micromamba version to v2

mtsokol:
This PR reverts CI change introduced in #799.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,10 +34,8 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
-      - uses: mamba-org/setup-micromamba@v1.9.0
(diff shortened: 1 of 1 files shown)


hameerabbasi (approved):

mtsokol:
I think we should merge it once windows job passes, so when https://github.com/mamba-org/setup-micromamba/issues/227 is fixed. WDYT?

hameerabbasi:
> I think we should merge it once windows job passes, so when https://github.com/mamba-org/setup-micromamba/issues/227 is fixed. WDYT?

SGTM, can't merge it until CI passes anyway. So my recommendation would be to re-run the CI and then merge after a fix.

mtsokol:
I think we can merge it - the bug was solved in mamba (failing job is unrelated I think)

hameerabbasi:
Thanks, @mtsokol!

PR #812, opened 2024-11-19
update finch-tensor-python to fix tests

willow-ahrens:
This should fix https://github.com/finch-tensor/Finch.jl/issues/644


willow-ahrens:
@hameerabbasi it looks like this fixes the CI regression for Finch. The failure is an unrelated issue that the MLIR backend is wrong on Ubuntu.

hameerabbasi (approved):

hameerabbasi:
Thanks for the quick fix, @willow-ahrens!

PR #814, opened 2024-11-25
MAINT: Re-enable COO tests after fix

mtsokol:
Hi @hameerabbasi,

As described in https://github.com/llvm/llvm-project/issues/116012#issuecomment-2497561848 this PR re-enables COO tests and makes (temporarily) COO instantiation copy-full.

Addition ops with COO still fail though.


hameerabbasi (approved):
Thanks for the deep dive here, @mtsokol

hameerabbasi:
Thanks for the deep dive here, @mtsokol!

hameerabbasi:
Seems like this doesn't fix `test_add` and `test_add_dense_sparse`, I just tried removing the `xfail`s and it crapped out.

PR #819, opened 2024-11-28
Introduce format factories

hameerabbasi:

- [x] 💾 Refactor
- [x] 🪄 Feature
- [x] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

***

Introduces format factories to simplify the building of formats without having to fudge with levels.

### DIFF
--- a/sparse/mlir_backend/formats.py
+++ b/sparse/mlir_backend/formats.py
@@ -23,7 +23,7 @@ from ._dtypes import DType, asdtype
 
 _CAMEL_TO_SNAKE = [re.compile("(.)([A-Z][a-z]+)"), re.compile("([a-z0-9])([A-Z])")]
 
-__all__ = ["LevelProperties", "LevelFormat", "StorageFormat", "Level", "get_storage_format"]
+__all__ = ["LevelProperties", "LevelFormat", "ConcreteFormat", "Level", "get_concrete_format"]
 
 
 def _camel_to_snake(name: str) -> str:
@@ -61,7 +61,7 @@ class Level:
 
 
 @dataclasses.dataclass(eq=True, frozen=True, kw_only=True)
-class StorageFormat:
+class ConcreteFormat:
     levels: tuple[Level, ...]
     order: tuple[int, ...]
     pos_width: int
@@ -151,7 +151,7 @@ class StorageFormat:
                     _hold_ref(arr, self)
                 return arrays
 
-            def get_storage_format(self) -> StorageFormat:
+            def get_storage_format(self) -> ConcreteFormat:
                 return storage_format
 
             @classmethod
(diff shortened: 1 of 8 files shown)


hameerabbasi:
Welp, turns out the `sparse_vector` complex issue is consistent after this PR.

And the Array API tests fail because of some change in Hypothesis randomness, I assume. I'll create a separate PR with a fix for that bug.

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -503,6 +450,8 @@ def test_reshape_dense(dtype):
 @pytest.mark.parametrize("src_fmt", ["csr", "csc", "coo"])
 @pytest.mark.parametrize("dst_fmt", ["csr", "csc", "coo"])
 def test_asformat(rng, src_fmt, dst_fmt):
+    if "coo" in {src_fmt, dst_fmt}:
+        pytest.xfail(reason="https://github.com/llvm/llvm-project/issues/116012")
```
hameerabbasi, line 454: This started failing again despite the copy.

mtsokol:
Thanks for the PR! I will review it once CI is green (let's skip what we need to skip for now).

hameerabbasi:
I plan to merge-commit this one since it has a few things mixed in; so I rewrote the history to be cleaner.

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -323,13 +300,9 @@ def test_coo_3d_format(dtype):
...
-    format = sparse.levels.get_storage_format(
-        levels=(sparse.levels.Level(sparse.levels.LevelFormat.Compressed),),
-        order="C",
-        pos_width=64,
-        crd_width=64,
-        dtype=sparse.asdtype(dtype),
-    )
+    if sparse.asdtype(dtype) in {sparse.complex64, sparse.complex128}:
+        pytest.xfail("Heisenbug")
+    format = sparse.formats.Coo().with_ndim(1).with_dtype(dtype).build()
```
mtsokol, line 305: It's confusing for me that SparseVector format is being constructed from `Coo` to be honest. As an end-user I would go for a `Csf().with_ndim(1)` but this one is dense, right?

How about disallowing `Csf` and `Coo` with 1-dimension? And have `Dense` (arbitrary dimensionality) and separate `SparseVector` (only 1-dimension)?

sparse/mlir_backend/tests/test_simple.py
```diff
@@ -323,13 +300,9 @@ def test_coo_3d_format(dtype):
...
 def test_sparse_vector_format(dtype):
-    format = sparse.levels.get_storage_format(
-        levels=(sparse.levels.Level(sparse.levels.LevelFormat.Compressed),),
-        order="C",
-        pos_width=64,
-        crd_width=64,
-        dtype=sparse.asdtype(dtype),
-    )
+    if sparse.asdtype(dtype) in {sparse.complex64, sparse.complex128}:
+        pytest.xfail("Heisenbug")
```
mtsokol, line 304: Let's give a one sentence description why we skip it (because in assertion first elements have incorrect values on python 3.10 IIRC). We can also `xfail` with `if` for the Python version where it fails.

sparse/mlir_backend/formats.py
```diff
@@ -61,7 +61,7 @@ def build(self):
 
 
 @dataclasses.dataclass(eq=True, frozen=True, kw_only=True)
-class StorageFormat:
+class ConcreteFormat:
```
mtsokol, line 64: `ConcreteFormat` suggests that there's  `AbstractFormat` somewhere. Didn't `StorageFormat` fit here? Why is our format concrete now?

mtsokol (changes requested):
A few comments and it can be merged from my side!

sparse/mlir_backend/tests/test_simple.py
hameerabbasi: Well, `Coo` is actually 1 compressed + n singleton and `Csf` is 1 dense + n compressed.

So by that logic -- `Csf().with_ndim(1)` would be dense, while `Coo().with_ndim(1)` is actually what we want. So this statement is actually quite consistent. Would you be happy with documenting this behaviour?

sparse/mlir_backend/tests/test_simple.py
hameerabbasi: This is actually quite random on the Python version, but deterministic on previous code paths + data somehow. I don't fully know the reason. I'll add a description.

sparse/mlir_backend/formats.py
hameerabbasi: A format factory is what most users would think of as a format (just specify the format, ndim and dtype, and be ready to go), whereas `StorageFormat` is more concrete: It includes pos_width and crd_width as well as very precise levels and order. Additionally, it is what one would allocate as an output.

mtsokol (approved):

PR #835, opened 2025-01-25
feat: Greek letters in einsum

jwjeffr:

- [ ] 💾 Refactor
- [X] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [ ] 🧪 Test
- [ ] 🛠️ Other

## Related issues

- Related issue #
- Closes #

## Checklist

- [ ] Code follows style guide
- [X] Tests added
- [ ] Documented the changes

***

Added Greek letters to valid einsum symbols, and a test that ensures all einsum symbols get parsed properly.

Just a QOL change, makes it easier to translate some tensor contractions that might use Greek letters by convention

### DIFF
--- a/sparse/numba_backend/tests/test_einsum.py
+++ b/sparse/numba_backend/tests/test_einsum.py
@@ -82,7 +82,7 @@ einsum_cases = [
 ]
 
 
-@pytest.mark.parametrize("symbols", sparse.numba_backend._common._EINSUM_SYMBOLS_SET)
+@pytest.mark.parametrize("symbols", sparse.numba_backend._common._EINSUM_SYMBOLS)
 def test_symbols(symbols):
     arr = sparse.random(shape=10, density=1.0)
 
--- a/sparse/numba_backend/tests/test_einsum.py
+++ b/sparse/numba_backend/tests/test_einsum.py
@@ -84,7 +84,6 @@ einsum_cases = [
 
 @pytest.mark.parametrize("symbols", sparse.numba_backend._common._EINSUM_SYMBOLS_SET)
 def test_symbols(symbols):
-
     arr = sparse.random(shape=10, density=1.0)
 
     # ensure we can use any of the defined symbols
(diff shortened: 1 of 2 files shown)


hameerabbasi:
Hi! Thanks for the PR! We usually like to avoid adding new API/features, instead trying to perfect what we have already.

One question I have is: Do other array implementations support this feature?

sparse/numba_backend/tests/test_einsum.py
```diff
@@ -82,6 +82,15 @@
 ]
 
 
+@pytest.mark.parametrize("symbols", sparse.numba_backend._common._EINSUM_SYMBOLS_SET)
```
hameerabbasi, line 85:
```suggestion
@pytest.mark.parametrize("symbols", sparse.numba_backend._common._EINSUM_SYMBOLS)
```

This should get rid of the test failures

hameerabbasi: For context, `set` is unordered in Python (which can be confusing as `dict` IS ordered). So each runner ends up with a different order of the same tests, which apparently trips up `pytest-xdist` (the framework we use to parallelize tests).

jwjeffr:
I haven't been able to find other array libraries that have this kind of support. [einops](https://pypi.org/project/einops/) supports Greek indices (among lots of other things), but it also supports `sparse >= 0.15` as is, so mixing the two would give me access to Greek indices.

No worries if this isn't the kind of feature you normally add - this was just an easy enough change that I figured it would be worth making a PR about :)

hameerabbasi:
It's good to know your use-case can be easily supported by something like the einops package. This feature isn't well-suited for this library, but I hope you won't be discouraged and continue contributing. I'm available on Discord if you'd like to chat, usually in [the following channel.](https://discord.com/channels/786703927705862175/1301155724646289420)

PR #848, opened 2025-03-12
Add HITS use case

mtsokol:
This PR adds [HITS](https://en.wikipedia.org/wiki/HITS_algorithm) algorithm use case, found in [graphblas-algorithms](https://github.com/python-graphblas/graphblas-algorithms).


hameerabbasi (commented):
General suggestion

examples/hits_example.py
```diff
@@ -0,0 +1,82 @@
...
+
+def converged(xprev: Array, x: Array, N: int, tol: float) -> bool:
+    err = xp.sum(xp.abs(x - xprev))
+    return err < xp.asarray(N * tol)
+
+
+class Graph:
+    def __init__(self, N: int, A: Array):
+        assert A.ndim == 2 and all(dim == N for dim in A.shape)
+        self.N = N
```
hameerabbasi, line 27: Maybe it makes sense to infer `N` from `A` instead of an `assert` to make sure it's consistent? Or maybe we can directly pass an array in and use its `.shape` property?

mtsokol: Right, we don't need that - done!

examples/hits_example.py
```diff
@@ -0,0 +1,81 @@
...
+        return xp.asarray([]), xp.asarray([])
+
+    h = xp.full((N, 1), 1.0 / N)
+    A = xp.asarray(G.A)
+
+    for _ in range(max_iter):
+        hprev = h
+        a = hprev.mT @ A
+        h = A @ a.mT
+        h = h / xp.linalg.vector_norm(h, ord=xp.inf)
```
hameerabbasi, line 52:
```suggestion
        h = h / xp.max(h, axis=...)
```

Why obfuscate? 😉 `xp.max` is much simpler than `xp.linalg.vector_norm(..., ord=xp.inf)`.

examples/hits_example.py
```diff
@@ -0,0 +1,81 @@
...
+        if converged(hprev, h, N, tol):
+            break
+        # alternatively these lines can be compiled
+        # h, a, conv = kernel(h, A, N, tol)
+    else:
+        raise Exception("Didn't converge")
+
+    if normalized:
+        h = h / xp.linalg.vector_norm(h, ord=1)
+        a = a / xp.linalg.vector_norm(a, ord=1)
```
hameerabbasi, line 62: Again,

```suggestion
        h = h / xp.max(sum(abs(h), axis=...))
        a = a / xp.max(sum(abs(a), axis=...))
```

hameerabbasi (changes requested):
With this we don't need `linalg` anymore. ;)

examples/hits_example.py
mtsokol: In this use case I wanted to show broader use of Array API, and show that `linalg` is also a standardized namespace, so it serves showcase purposes.
But I can change it to `xp.max`/`xp.sum` if you think it suits presentation better.

hameerabbasi: I'd only use the linalg namespace when we have things that can't be expressed as functions in the main namespace.

hameerabbasi (approved):

PR #849, opened 2025-03-13
Fix CSC/CSR `.mT` fill-value result

mtsokol:
Hi @hameerabbasi,

It looks like transposing CSC/CSR drops fill-value. Here's a fix.


hameerabbasi (dismissed):
Thanks, @mtsokol!

mtsokol:
> Looks like it's just one test -- that can be bundled and we can merge-commit this PR instead of squash.

Done!

hameerabbasi (approved):

PR #857, opened 2025-04-01
MAINT: Use asterisk import for `finch-tensor`

mtsokol:


hameerabbasi (approved):
Ahh, I always love removing lots of lines! Thanks, @mtsokol.

PR #862, opened 2025-04-09
Addressing array api failure for sort() function

prady0t:
Removes :

```
FAILED ../array-api-tests/array_api_tests/test_sorting_functions.py::test_sort - AssertionError: out.shape=(), but should be (1,) [sort()]
```
While running array-api-tests.


hameerabbasi (approved):
Thanks for the fix, @prady0t!

Issue #739, opened 2024-08-07
Add `codespell` to pre-commit


DeaMariaLeon:
Maybe consider checking spelling?: [conbench.github.io/conbench/index.html](https://github.com/codespell-project/codespell)

PR #866, opened 2025-04-24
Added codespell to pre-commit

prady0t:
Fixes: #739


.pre-commit-config.yaml
```diff
@@ -36,4 +36,4 @@ repos:
   hooks:
     - id: codespell
       files: ^.*\.(py|md|yml)$
-      args: ["--write-changes", "--ignore-words=FILE"]
```
hameerabbasi: There should probably be an ignore-words.txt file. 😉

PR #943, opened 2026-06-18
Array API 2025.12 support

### LINKED ISSUE #940: array API: add `broadcast_shapes`
https://data-apis.org/array-api/draft/API_specification/generated/array_api.broadcast_shapes.html

https://data-apis.org/array-api/draft/changelog.html#v2025-12

@prady0t might you be interested in working on this?

hameerabbasi:

- [ ] 💾 Refactor
- [x] 🪄 Feature
- [x] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

- Closes #940

## Checklist

- [x] Tests added
- [x] Documented the changes

***


hameerabbasi:
@prady0t `__ifloordiv__` fails for some weird reason I can't figure out; all other tests (including elemwise) succeed. Maybe an issue with the test suite or a latent bug?

prady0t:
I think the hypothesis issue has now been resolved. After a lot of effort, I found out that the `asarray` implementation in sparse does not take `copy` into account. See the latest commit I pushed. 

Also see this example:

```py
import numpy as np
from sparse.numba_backend import COO, asarray

x = COO.from_numpy(np.array(6.0))
y = asarray(x, copy=True)
a = asarray(x, copy=False)
z = np.asarray(x.todense(), copy=True)
print(id(x), id(y), id(z), id(a))
print(x is a)
print(x is y)
print(x is z)
```

The output is:


```
5727776080 5727776080 5770313616 5727776080
True
True
False
```

Which means `x`, `y`, and `a` are the same object, and it does not affect the object if set to `True` or `False`, whereas `copy` should produce different objects, just like `z`

After this commit, the output is:

```
6072479056 6072479376 6073926064 6072479056
True
False
False
```

Which is the correct behaviour.

hameerabbasi:
Well done tracking that down! Can we add a test to ensure that `asarray` with `copy=True` copies for all formats?

prady0t:
>  Can we add a test to ensure that asarray with copy=True copies for all formats?

Sure. In this PR itself?

hameerabbasi:
> Sure. In this PR itself?

That'd be nice.

prady0t:
`DOK` had no `copy()`; added it in the latest commit(inspired by GCXS copy).

prady0t (dismissed):

hameerabbasi:
Still need to remove some `xfail`s. I'll do that later today.

lucascolley:
> I found out that the `asarray` implementation in sparse does not take `copy` into account.

I wonder why we didn't catch this earlier (and more easily) via the array-api-tests suite? Is there something we should add over there?

hameerabbasi:
> I wonder why we didn't catch this earlier (and more easily) via the array-api-tests suite? Is there something we should add over there?

You probably do; but by default the arrays are also immutable (don't implement setitem) in PyData/Sparse; which is why the test probably got skipped.

PR #945, opened 2026-06-25
fix: treat NaN values as distinct in unique_values and unique_counts

### LINKED ISSUE #855: Bug: `unique_values`/`unique_counts`: NaNs should not be treated as equal
### sparse version checks

- [x] I checked that this issue has not been reported before [list of issues](https://github.com/pydata/sparse/issues).

- [x] I have confirmed this bug exists on the latest version of sparse.

- [x] I have confirmed this bug exists on the main branch of sparse.

### Describe the bug

The Array API specification [states](https://data-apis.org/array-api/latest/API_specification/generated/array_api.unique_values.html#unique-values) for `unique_*` functions:

> As nan values compare as False, nan values should be considered distinct.

In other words, it dictates behaviour equal to `np.unique(..., equal_nan=False)`.

`sparse.unique_*` functions however treat NaNs as equal.

### Steps or code to reproduce the bug

```python
>>> from math import nan

>>> import array_api_strict as xp
>>> xp.unique_values(xp.asarray([math.nan, math.nan]))
Array([nan, nan], dtype=array_api_strict.float64)

>>> import sparse
>>> sparse.unique_values(sparse.asarray([math.nan, math.nan]))
array([nan])
```

### Expected results

.

### Actual results

.

### Please describe your system.

sparse 0.16.0b4

### Relevant log output

```shell

```

gaoflow:
## Summary

Fixes #855.

The [Array API specification](https://data-apis.org/array-api/latest/API_specification/generated/array_api.unique_values.html) states:

> As nan values compare as False, nan values should be considered distinct.

Both `unique_values` and `unique_counts` were using `np.unique` (which defaults to `equal_nan=True`), causing multiple NaN values to be collapsed into a single entry:

```python
>>> import sparse
>>> sparse.unique_values(sparse.asarray([float('nan'), float('nan')]))
array([nan])   # wrong: should be [nan, nan]
```

## Changes

- Pass `equal_nan=False` to `np.unique` so NaN values in stored data are kept distinct.
- Handle the case where `fill_value` is NaN: instead of prepending a single fill-value entry, append one distinct NaN per un-stored slot.

## After fix

```python
>>> sparse.unique_values(sparse.asarray([float('nan'), float('nan')]))
array([nan, nan])   # correct

>>> sparse.unique_counts(sparse.asarray([float('nan'), float('nan'), float('nan')]))
UniqueCountsResult(values=array([nan, nan, nan]), counts=array([1, 1, 1]))   # correct
```

Three new tests covering:
1. NaN values in stored data treated as distinct by `unique_values`
2. NaN fill-value slots each producing a distinct NaN entry
3. NaN values counted individually by `unique_counts`

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/sparse/numba_backend/_coo/common.py
+++ b/sparse/numba_backend/_coo/common.py
@@ -1211,13 +1211,19 @@ def unique_counts(x, /):
     x = _validate_coo_input(x)
 
     x = x.flatten()
-    values, counts = np.unique(x.data, return_counts=True)
-    if x.nnz < x.size:
-        values = np.concatenate([[x.fill_value], values])
-        counts = np.concatenate([[x.size - x.nnz], counts])
-        sorted_indices = np.argsort(values)
-        values[sorted_indices] = values.copy()
-        counts[sorted_indices] = counts.copy()
+    values, counts = np.unique(x.data, return_counts=True, equal_nan=False)
+    fill_count = x.size - x.nnz
+    if fill_count > 0:
+        if np.isnan(x.fill_value):
+            # Per the Array API spec, NaNs compare as False, so each NaN is distinct.
+            values = np.concatenate([values, np.full(fill_count, np.nan)])
+            counts = np.concatenate([counts, np.ones(fill_count, dtype=counts.dtype)])
+        else:
+            values = np.concatenate([[x.fill_value], values])
+            counts = np.concatenate([[fill_count], counts])
+            sorted_indices = np.argsort(values)
+            values[sorted_indices] = values.copy()
+            counts[sorted_indices] = counts.copy()
 
     return UniqueCountsResult(values, counts)
 
@@ -1252,9 +1258,14 @@ def unique_values(x, /):
     x = _validate_coo_input(x)
 
     x = x.flatten()
-    values = np.unique(x.data)
-    if x.nnz < x.size:
-        values = np.sort(np.concatenate([[x.fill_value], values]))
+    values = np.unique(x.data, equal_nan=False)
+    fill_count = x.size - x.nnz
+    if fill_count > 0:
+        if np.isnan(x.fill_value):
+            # Per the Array API spec, NaNs compare as False, so each NaN is distinct.
+            values = np.concatenate([values, np.full(fill_count, np.nan)])
+        else:
+            values = np.sort(np.concatenate([[x.fill_value], values]))
     return values
 
 
--- a/sparse/numba_backend/_coo/common.py
+++ b/sparse/numba_backend/_coo/common.py
@@ -1216,7 +1216,7 @@ def unique_counts(x, /):
     if fill_count > 0:
         if np.isnan(x.fill_value):
             # Per the Array API spec, NaNs compare as False, so each NaN is distinct.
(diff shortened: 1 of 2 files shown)


sparse/numba_backend/_coo/common.py
```diff
@@ -1211,13 +1211,19 @@ def unique_counts(x, /):
...
-        counts = np.concatenate([[x.size - x.nnz], counts])
-        sorted_indices = np.argsort(values)
-        values[sorted_indices] = values.copy()
-        counts[sorted_indices] = counts.copy()
+    values, counts = np.unique(x.data, return_counts=True, equal_nan=False)
+    fill_count = x.size - x.nnz
+    if fill_count > 0:
+        if np.isnan(x.fill_value):
+            # Per the Array API spec, NaNs compare as False, so each NaN is distinct.
+            values = np.concatenate([values, np.full(fill_count, np.nan)])
```
hameerabbasi:
```suggestion
            values = np.concatenate([values, np.full(fill_count, x.fill_value)])
```

To preserve the exact "kind" of NaN that gets forwarded.

sparse/numba_backend/_coo/common.py
```diff
@@ -1252,9 +1258,14 @@ def unique_values(x, /):
...
     x = x.flatten()
-    values = np.unique(x.data)
-    if x.nnz < x.size:
-        values = np.sort(np.concatenate([[x.fill_value], values]))
+    values = np.unique(x.data, equal_nan=False)
+    fill_count = x.size - x.nnz
+    if fill_count > 0:
+        if np.isnan(x.fill_value):
+            # Per the Array API spec, NaNs compare as False, so each NaN is distinct.
+            values = np.concatenate([values, np.full(fill_count, np.nan)])
```
hameerabbasi:
```suggestion
            values = np.concatenate([values, np.full(fill_count, x.fill_value)])
```

hameerabbasi (changes requested):
Mostly LGTM; I've left two minor comments.

hameerabbasi (dismissed):

hameerabbasi:
A test for the missing lines of coverage would be nice.

binggao1230:
Pushed `325a0ad` to add coverage for the missing `unique_counts` + `fill_value=nan` branch.

Local verification:

```bash
uv run --with-editable . --extra tests python -m pytest sparse/numba_backend/tests/test_coo.py::TestUnique -q
python -m ruff check sparse/numba_backend/_coo/common.py sparse/numba_backend/tests/test_coo.py
python -m ruff format --check sparse/numba_backend/_coo/common.py sparse/numba_backend/tests/test_coo.py
```

I also looked at the current `xp-tests-finch` failure. It reports a broad set of Finch array-api failures (211 failed, including missing `copysign`, `hypot`, `maximum`, `minimum`, `count_nonzero`, `sort`, etc.), which does not look specific to this `unique_values` / `unique_counts` change.

hameerabbasi (approved):

hameerabbasi:
Thanks for the contribution, @gaoflow.

PR #957, opened 2026-09-14
fix: avoid dense column selectors in GCXS basic slicing

### LINKED ISSUE #853: Bug: GCXS slicing can take inordinate amount of time / crash kernel
### sparse version checks

- [x] I checked that this issue has not been reported before [list of issues](https://github.com/pydata/sparse/issues).

- [x] I have confirmed this bug exists on the latest version of sparse.

- [x] I have confirmed this bug exists on the main branch of sparse.

### Describe the bug

I have noticed that under some circumstances, slicing a GCXS array will take an unexpecteda amount of time, or crash the kernel after a runtime of >45 seconds.

### Steps or code to reproduce the bug

```
a = sparse.COO(
  [[1, 100, 215, 66],[5, 101, 242, 11],[3, 5, 1, 11],[13, 1, 3, 1],[55, 1, 6, 8]],
  [5, 10, 2, 1],
  shape=(255, 255, 255, 255, 255)
)
b = a.asformat('gcxs')
b[1, :, :, :, :]
```

### Expected results

A slice returned within a short amount of time, provided the sparsity of the matrix.

`b[:, 1, :, :, :]`, `b[:, :, 1, :, :]` and `b[:, :, :, 1, :]` return after a fraction of a second.
`b[: :, :, :, 1]` for some reason takes ~4 seconds on my machine.

### Actual results

Kernel crash after >45 seconds of runtime.

### Please describe your system.

1. OS and version: Ubuntu 22.04, latest update
2. sparse version: '0.16.0b4'
3. NumPy version: '2.1.3'
4. Numba version: '0.61.0'

### Relevant log output

```shell
The Kernel crashed while executing code in the current cell or a previous cell. 
Please review the code in the cell(s) to identify a possible cause of the failure. 
Click here for more info. 
View Jupyter log for further details.



19:39:49.903 [info] Restarted f3b136f3-350b-49cc-90d0-e337e991f066
19:46:21.391 [error] Disposing session as kernel process died ExitCode: undefined, Reason: 

(this is the full, unabridged log output from the crashed session)
```

Doribelove:

- [ ] 💾 Refactor
- [ ] 🪄 Feature
- [x] 🐞 Bug Fix
- [x] 🔧 Optimization
- [ ] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

Closes #853. This addresses the large Cartesian allocation also identified in numba/numba#10040; it does not attempt to resolve every platform-specific compilation error discussed there.

## Checklist

- [x] Tests added
- [x] Documented the changes

Slicing a GCXS array currently expands the selected uncompressed coordinates into a Cartesian column index, even when the selected rows contain only a few stored entries. In #853, four entries in a `(255, 255, 255, 255, 255)` array lead `x[1]` to request `255**4` int64 column indices: about 33.8 GB for that array alone.

This change filters stored columns for broad basic column selections, decodes their coordinates, and maps matches to the sliced shape. Negative steps sort the output indices and data together; empty columns return immediately. Narrow selections retain the existing search path when their column count is bounded by the largest selected row's stored-entry count. Advanced column indices retain the existing Cartesian path, including repeated indices. Compressed-row selection and output reconstruction retain their existing algorithms.

The new column-filtering temporaries scale with the stored entries in the selected rows, including repeated rows. This is not a claim that every GCXS slicing allocation scales only with `nnz`: row selection and the resulting `indptr` still depend on the compressed shape. No public API or fill-value behavior changes.

Regression tests cover the original huge shape with an allocation guard that fails safely on the old implementation, as well as stepped/reversed slices, multiple compressed axes, integer axes, empty results, nonzero/NaN fill values, index dtypes, steps beyond the machine integer range, CSR/CSC, and repeated advanced indices. Added benchmarks keep four stored entries while increasing the logical dimensions; their sizes are bounded so the same benchmarks can safely run against the old implementation.

### Validation

Base: `bddecfc2574ea725bd32824a24fb9a547c0a1056`. Local Linux x86_64, Python 3.12.14, NumPy 2.5.3, Numba 0.67.0, SciPy 1.18.1.

- New regression file on the original implementation: **2 failed, 58 passed**; both failures are allocation guards before unsafe column expansion. The revised regression file now has **74 passing cases**, including the review follow-up below.
- Full Numba-backend package tests and doctests (`pytest --pyargs sparse --doctest-modules -n 2 -o addopts=` with coverage): **6,283 passed, 15 skipped, 98 xfailed, 59 xpassed**, two existing warnings. Untouched baseline: 6,209 passed, the same 15 skips and two warnings. Existing float16 XFAIL/XPASS counts vary with random inputs; a separate 120-pair comparison with fixed seeds produced identical before/after results.
- Nine added benchmark cases pass under ordinary pytest; CodSpeed instrumentation was not run locally. Repository-pinned pre-commit hooks on all three changed files, `git diff --check`, pass. The initial wheel build passed.
- Finch/Julia, MLIR, and the full platform/backend CI matrix were not run locally.

Local measurements after JIT warmup reduced the four-entry `48**5` example from 52.30 ms to 0.158 ms. The original `255**5` case completed correctly with a 256 MiB additional address-space allowance after warmup; the unsafe original-size baseline was not executed. Twelve ordinary 2D CSR/CSC workloads ranged from 6.5% faster to 11.6% slower, so these measurements do not imply a universal speedup. Timings are small local samples, separate from CodSpeed CI.

### Review follow-up

The index-array list now has an explicit Numba element type at initialization, so an empty list can cross the JIT boundary. Its read-only, arbitrary-stride array type supports both contiguous and strided indices, including read-only inputs, without unsafe-cast warnings. Four empty-list and one mixed-layout failure were reproduced on the first PR commit before this change. Fourteen additional test cases cover these concerns and verify that empty column slices preserve read-only source buffers.

The selector's output buffer is named `out_indptr` and documented as newly allocated by `getitem`; it does not alias `x.indptr`. `ndarray.astype(dtype, copy=False)` is retained because it permits copying when the dtype differs, which the mixed-dtype tests verify. The descending `range` is also retained with a comment: `reversed(range(...))` raises a nopython `TypingError` in the tested Numba 0.67.0 environment.

AI assistance: the implementation, tests, benchmarks, and this description were prepared with Codex. Validation commands were executed locally.


sparse/numba_backend/_compressed/indexing.py
```diff
@@ -174,6 +197,74 @@ def getitem(x, key):
     return GCXS(arg, shape=shape, compressed_axes=compressed_axes, fill_value=x.fill_value)
 
 
+def _asindexarrays(key, dtype):
+    """Convert normalized indices to typed arrays for Cartesian flattening."""
+    arrays = []
```
hameerabbasi: This should be a `numba.typed.List`; reflected lists are deprecated as inputs/outputs from Numba functions. In addition, it should be typed on the same line.

sparse/numba_backend/_compressed/indexing.py
```diff
@@ -174,6 +197,74 @@ def getitem(x, key):
...
+    for ind in key:
+        if isinstance(ind, Integral):
+            ind = np.array([ind])
+        elif isinstance(ind, slice):
+            ind = np.arange(ind.start, ind.stop, ind.step)
+        elif isinstance(ind, np.ndarray) and ind.ndim > 1:
+            raise IndexError("Only one-dimensional iterable indices supported.")
+        arrays.append(ind.astype(dtype, copy=False))
+    return List(arrays)
```
hameerabbasi: See above, this will have an unknown type if `arrays` is empty; leading to errors compiling sometimes.

sparse/numba_backend/_compressed/indexing.py
```diff
@@ -174,6 +197,74 @@ def getitem(x, key):
...
+    """Convert normalized indices to typed arrays for Cartesian flattening."""
+    arrays = []
+    for ind in key:
+        if isinstance(ind, Integral):
+            ind = np.array([ind])
+        elif isinstance(ind, slice):
+            ind = np.arange(ind.start, ind.stop, ind.step)
+        elif isinstance(ind, np.ndarray) and ind.ndim > 1:
+            raise IndexError("Only one-dimensional iterable indices supported.")
+        arrays.append(ind.astype(dtype, copy=False))
```
hameerabbasi, line 211: `copy=False` might give issues with mixed-dtype indices, as it will raise an error.

sparse/numba_backend/_compressed/indexing.py
```diff
@@ -174,6 +197,74 @@ def getitem(x, key):
...
+    """
+    Filter stored columns for basic indexing without expanding the slice grid.
+
+    Decode each stored column into coordinates and map matching coordinates to
+    the sliced shape. Integer keys are represented by a length-one slice. The
+    temporary storage is bounded by the entries in the selected compressed rows,
+    including repetitions of a row, rather than the logical number of columns.
+    """
+    if np.any(lengths == 0):
+        indptr[:] = 0
```
hameerabbasi: Are we modifying the original array here? We really shouldn't do that.

sparse/numba_backend/_compressed/indexing.py
```diff
@@ -174,6 +197,74 @@ def getitem(x, key):
...
+        return arr_data[:0].copy(), arr_indices[:0].copy(), indptr
+
+    strides = np.empty(len(shape), dtype=np.intp)
+    out_strides = np.empty(len(shape), dtype=np.intp)
+    stride = 1
+    out_stride = 1
+    for axis in range(len(shape) - 1, -1, -1):
```
hameerabbasi, line 237:
```suggestion
    for axis in reversed(range(len(shape))):
```

hameerabbasi (changes requested):
Mostly LGTM, just a few small issues to address.

sparse/numba_backend/tests/test_compressed_slicing.py
```diff
@@ -5,11 +5,58 @@
...
+    def count_arrays(arrays):
+        return len(arrays)
+
+    arrays = indexing._asindexarrays([], np.dtype(idx_dtype))
+    assert count_arrays(arrays) == 0
+
+
+@pytest.mark.parametrize("idx_dtype", [np.int8, np.int16, np.int32, np.int64])
+@pytest.mark.parametrize("readonly", [False, True])
+@pytest.mark.filterwarnings("error::numba.core.errors.NumbaTypeSafetyWarning")
```
hameerabbasi, line 26: We shouldn't be filtering warnings, we should be fixing them if needed (e.g. via explicit type casts).

hameerabbasi (changes requested):
Just the one comment; we need to fix the warnings.

sparse/numba_backend/tests/test_compressed_slicing.py
hameerabbasi: Ahh; we're making sure there aren't any warnings here. Ignore me. 😅

hameerabbasi (approved):
I think a summary of why the unresolved comments are difficult to resolve would suffice.

Doribelove:
Thanks for the review. Here is the summary of the remaining points:

- `ndarray.astype(dtype, copy=False)` still permits the copy required for a dtype conversion; it reuses the input when the requested dtype/layout already match ([NumPy documentation](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html)). The mixed-dtype regressions in f700fc188115 exercise that conversion with contiguous, strided, and read-only indices, so I retained it.
- `reversed(range(...))` fails to compile in the tested Numba 0.67.0 nopython mode with `Untyped global name 'reversed'`. The equivalent descending `range` compiles, so it remains with an explanatory comment.

The other concerns are addressed in f700fc188115: the Numba list is explicitly typed before any append, including the empty case, and the helper parameter is named `out_indptr` to make its ownership clear. That buffer is newly allocated by `getitem` and does not alias `x.indptr`; regressions check empty-column slicing with read-only source buffers. The warning marker turns `NumbaTypeSafetyWarning` into an error, ensuring these tests cannot silently pass with that warning.

The validation already recorded for this revision includes 74 passing slicing regressions and 6,283 passing Numba-backend package tests/doctests, with the existing skips, expected failures, and two warnings noted in the PR description. The current main CI test jobs also pass.

hameerabbasi (approved):

Issue #860, opened 2025-04-09
Bug: Cannot use boolean GCXS array for indexing another GCXS array


ilia-kats:
### sparse version checks

- [x] I checked that this issue has not been reported before [list of issues](https://github.com/pydata/sparse/issues).

- [x] I have confirmed this bug exists on the latest version of sparse.

- [ ] I have confirmed this bug exists on the main branch of sparse.

### Describe the bug

Trying to use a boolean GCXS array to index another GCXS array with the same dimensions results in either `IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices` (when attempting to get values at the index) or `ValueError: <class 'int'> can be computed for one-element arrays only.` (when attempting to set values at the index). The setting operation is requred by the [Dask `nanvar` implementation](https://github.com/dask/dask/blob/0b526507ab7254f435039536d058c7c5b958ec4b/dask/array/reductions.py#L555). This is currently preventing me from using sparse with Dask.

### Steps or code to reproduce the bug

minimal reprex:
```python
import sparse
import numpy as np
test = sparse.GCXS(np.random.rand(5,5), compressed_axes=(0,))

test[test < 0] # IndexError

test[test < 0] = np.nan # ValueError
```

### Expected results

No error is thrown and the operation succeeds.

### Actual results

For getting values:
```python
File /data/ilia/envs/famo/lib/python3.11/site-packages/sparse/numba_backend/_compressed/indexing.py:29, in getitem(x, key)
     26         return result
     27     return GCXS.from_coo(result)
---> 29 key = list(normalize_index(key, x.shape))
     31 # zip_longest so things like x[..., None] are picked up.
     32 if len(key) != 0 and all(isinstance(k, slice) and k == slice(0, dim, 1) for k, dim in zip_longest(key, x.shape)):

File /data/ilia/envs/famo/lib/python3.11/site-packages/sparse/numba_backend/_slicing.py:58, in normalize_index(idx, shape)
     56 for i, d in zip(idx, none_shape, strict=True):
     57     if d is not None:
---> 58         check_index(i, d)
     59 idx = tuple(map(sanitize_index, idx))
     60 idx = tuple(map(replace_none, idx, none_shape))

File /data/ilia/envs/famo/lib/python3.11/site-packages/sparse/numba_backend/_slicing.py:122, in check_index(ind, dimension)
    120     return
    121 elif not isinstance(ind, Integral):
[...]

PR #962, opened 2026-09-22
feat: support sparse boolean masks in COO and GCXS

### LINKED ISSUE #304: Support boolean indexing with a COO sparse matrix
While investigating https://github.com/theislab/scanpy/issues/921 I ran into the following limitation when using a boolean array as an index.

This is the expected behaviour for regular NumPy:

```python
>>> import numpy as np
x = np.array([1, 2, 3, 4])
>>> x[x % 2 == 0]
array([2, 4])
```
But the same thing fails for sparse:

```python
>>> import sparse
>>> s = sparse.COO.from_numpy(x)
>>> s
<COO: shape=(4,), dtype=int64, nnz=4, fill_value=0>
>>> [s % 2 == 0]
[<COO: shape=(4,), dtype=bool, nnz=2, fill_value=True>]
>>> s[s % 2 == 0]
<class 'sparse._coo.core.COO'>
  File "<stdin>", line 1, in <module>
  File "/Users/tom/workspace/scanpy/venv/lib/python3.7/site-packages/sparse/_coo/indexing.py", line 60, in getitem
    index = normalize_index(index, x.shape)
  File "/Users/tom/workspace/scanpy/venv/lib/python3.7/site-packages/sparse/_slicing.py", line 58, in normalize_index
    check_index(i, d)
  File "/Users/tom/workspace/scanpy/venv/lib/python3.7/site-packages/sparse/_slicing.py", line 123, in check_index
    raise IndexError("only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and "
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```
It would be good to fix it for `GCXS` if that's a problem too. I couldn't reproduce it there since `from sparse import GCXS` fails.

Doribelove:

- [ ] 💾 Refactor
- [x] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [x] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

Closes #304. Addresses the retrieval part of #860; assignment remains unsupported.

## Checklist

- [x] Tests added
- [x] Documented the changes

`x[x > 0]` and `x[x == 0]` currently raise `IndexError` for COO and GCXS arrays. This adds a single sparse boolean mask over matching leading dimensions, including full-shape, scalar and zero-length masks. Python/NumPy boolean scalar keys insert the same leading axis as a zero-dimensional mask. The output retains the source fill value.

The COO path searches stored mask coordinates. With a True fill, it subtracts the number of preceding exclusions to compute output coordinates without enumerating implicit True entries. GCXS delegates through COO. An integer coordinate buffer avoids float promotion when source coordinates are uint64.

Both indexing paths use `SparseArray` for the mask check, as suggested in review. COO, GCXS, and DOK masks are covered by the regressions. The branch includes current main through `6f5f88f`, including the scalar conversion fix from #963 needed for zero-dimensional DOK masks.

The documentation explicitly limits this to standalone masks (`x[mask]` or `x[(mask,)]`); mixed indices, sparse integer indices, and assignment are outside this change. Source arrays are COO or GCXS. Other backends and the general Array API boolean-indexing capability flag are unchanged.

### Array API review follow-up

Removed the blanket skip for `test_getitem_masking`. The pinned upstream test is one Hypothesis property, so its existing ID cannot distinguish sparse-array operands from NumPy scalar operands. A small Numba-only pytest plugin now collects two parameters while keeping the upstream assertions:

- `[sparse-array]` keeps zero-dimensional sources as sparse arrays and retains the upstream shape, dtype, and mask strategies. It runs without a skip or xfail.
- `[numpy-scalar]` isolates `numpy_scalar[sparse_zero_dimensional_mask]`, which still fails in NumPy's indexing implementation. Only this parameter is in the xfails file. The case fails deterministically with CI's small example budget.

The plugin is loaded by the existing Numba Array API task, restores the generator after each test, and does not edit the external test checkout. This avoids treating the whole property as an expected failure or changing Sparse's public scalar-return behavior.

Validation on Python 3.12 / NumPy 2.5 / Numba 0.67:

- 243 parametrized regressions pass, including 52 additional DOK-mask cases, trillion-element shapes, True-fill masks without densification, empty masks and boolean scalar keys.
- The restored `[sparse-array]` property passes 1,000 generated examples. It fails against the original unmodified base, confirming that the adapter still detects the missing feature. With CI's example budget and parallel execution, masking reports 1 passed / 1 xfailed.
- Full suite and module doctests with the Numba backend: 6,602 passed, 15 skipped, 96 xfailed, 61 xpassed, 7 warnings. Finch/MLIR and other platforms were not tested locally.
- Pinned Array API suite: 1,180 passed, 84 skipped, 120 xfailed; one unrelated strict XPASS for complex `acosh` remains and was also reproduced on unmodified main in the same local environment. This is not an all-green Array API result.
- Repository pre-commit hooks, wheel build, and `git diff --check` pass. The earlier revision also passed 3,504 fixed-seed comparisons against NumPy.

GitHub’s [`xp-tests-numba` check](https://github.com/pydata/sparse/actions/runs/35705366708/job/106672865933) passed on revision `7e3cfa9`. Other CI jobs may still be running.

Prepared with assistance from OpenAI Codex; the reported validation was executed locally.


sparse/numba_backend/_coo/indexing.py
```diff
@@ -27,8 +27,13 @@ def getitem(x, index):
     index : {tuple, str}
         The index into the array.
     """
+    from .._compressed import GCXS
     from .core import COO
 
+    mask = index[0] if isinstance(index, tuple) and len(index) == 1 else index
+    if isinstance(mask, COO | GCXS) and mask.dtype == np.bool_:
```
hameerabbasi:
```suggestion
    if isinstance(mask, SparseArray) and mask.dtype == np.bool_:
```

hameerabbasi (changes requested):
Hello! Thanks for the contribution. It's broadly okay, you'd just need to run the Array API tests with the Numba backend; and remove the passing tests from the xfails file.

xp-tests/Numba-array-api-skips.txt
```diff
@@ -1,3 +1,7 @@
+# This test also indexes NumPy scalars with sparse masks, which is unsupported.
+# Sparse operands are covered by numba_backend/tests/test_boolean_indexing.py.
+array_api_tests/test_array_object.py::test_getitem_masking
```
hameerabbasi: We shouldn't skip/xfail the whole test, only the special cases that fail should be in the xfail file.

xp-tests/numba_masking.py
hameerabbasi: This whole file shouldn't be needed. What we'd need is to simply run the tests and put JUST the failing cases into the `xfails` file.

pixi.toml
hameerabbasi: This file shouldn't change either.

xp-tests/numba_masking.py
Doribelove: Removed the plugin and its README. The unparameterized upstream Hypothesis test passes with CI's two-example budget, so marking the whole test xfail gives `XPASS(strict)`. At 200 examples, it can draw a NumPy scalar source and fail in NumPy's scalar indexing. I removed the whole-test xfail, kept deterministic regressions for sparse source arrays, and documented this limitation in the PR. Since the current xfails file can only target the single pytest ID, I would welcome your preference for tracking that generated scalar case without a collection plugin.

pixi.toml
Doribelove: Reverted both changes to `pixi.toml`; it is no longer in the final PR diff.

Doribelove:
@hameerabbasi, the current head (`996182f`) addresses the review points: the mask check uses `SparseArray`, the test-only plugin and `pixi.toml` edits are gone, and `test_getitem_masking` is no longer marked as a whole-test xfail. The Numba Array API check and the Linux, macOS, and Windows test matrices pass on this head. The remaining NumPy-scalar case generated by the pinned property at a larger Hypothesis budget is documented in the PR description; I would appreciate your guidance on its treatment.

Two CI checks are still red. The benchmark job errors in Finch's `Reflector` setup, also seen on #964, before the new masking path is exercised. `codecov/patch` passes at 100% of the diff, while `codecov/project` reports 90.65% against a 100% project target; #964 has the same project-check failure. Could you take another look at the implementation and advise whether any further PR-side change is needed? Thank you.

PR #964, opened 2026-09-22
feat: support Fortran order in COO.flatten

### LINKED ISSUE #603: Support for Fortran order in COO.flatten()
**Is your feature request related to a problem? Please describe.**
I am currently doing a GSoC project for [cvxpy](https://github.com/cvxpy/cvxpy) and we need to use 3d sparse matrices. 
The way we process matrices in the backend (using variables on one axis and constraints on another), it is more intuitive to store the data in column-major format as it makes the problem readable. Hence, we made a design decision to use fortran order for all matrix operations. Currently, we are trying to move from scipy.sparse to using your sparse library and this missing API flag is the main obstacle in doing so. 

**Describe the solution you'd like**
I would like to have flatten in fortran (column major order) for increased compatibility with NumPy. 
Another API that is missing and could be nice to have is [expand_dims](https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html) or also [atleast_2d](https://numpy.org/doc/stable/reference/generated/numpy.atleast_2d.html). 

I would be happy to make a PR for this, if someone could guide me on how to get started.

Doribelove:
`COO.flatten(order="F")` currently raises `NotImplementedError`, preventing column-major flattening of multidimensional sparse arrays. This adds the transpose-then-flatten implementation suggested in #603; for example, `[[0, 1, 2], [3, 4, 5]]` now becomes `[0, 3, 1, 4, 2, 5]`.

- [ ] 💾 Refactor
- [x] 🪄 Feature
- [ ] 🐞 Bug Fix
- [ ] 🔧 Optimization
- [x] 📚 Documentation
- [x] 🧪 Test
- [ ] 🛠️ Other

## Related issues

Closes #603.

## Checklist

- [x] Tests added
- [x] Documented the changes

For `order="F"`, reverse the COO axes before using the existing C-order reshape. This operates on sparse coordinates and data, preserves the fill value, and keeps the existing behavior for the default order, `"C"`, and `None`. Document the supported orders and add a column-major doctest. The change is limited to `COO.flatten`; `reshape` and GCXS order support are unchanged.

Regression coverage includes multidimensional, singleton, scalar and empty shapes; zero, nonzero and NaN fill values; arrays with no stored entries; narrow coordinate dtype promotion; and alternating C/F calls with caching enabled. A three-entry array with a logical size of 10^12 checks the resulting coordinates and values while rejecting dense conversion.

Validation on Python 3.12.14, NumPy 2.5.3, Numba 0.67.0 and SciPy 1.18.1:

- New regression suite against the original implementation: **27 failed, 48 passed**; failures were unsupported F-order calls.
- Updated implementation: **75 flatten tests passed**.
- Full Numba unit tests and doctests: **6,431 passed, 15 skipped, 95 xfailed, 62 non-strict xpassed**, with 2 warnings and no failures.
- Repository pre-commit checks passed; both added executable lines were exercised by the coverage run.

AI assistance was used for implementation, tests, and drafting. The validation above was executed locally against this patch.


Doribelove:
CI triage for this PR: the Numba/Finch/MLIR test matrix, examples, notebooks, experimental tests, docs, and `codecov/patch` pass. The two red checks reproduce on other PRs that do not touch `COO.flatten`:

- `benchmarks` fails in nine existing Finch elementwise cases with `AttributeError: module 'finch' has no attribute 'Reflector'`. The same error appears on merged #963 and on #962.
- `codecov/project` reports 90.58% here, the same 90.58% reported on merged #963; the configured project target is 100% with a 5% threshold. The patch coverage check passes here.

This branch changes only `sparse/numba_backend/_coo/core.py` and its COO tests. These comparisons suggest the red checks are not introduced by the flatten change; I am not claiming the overall CI is green. If a reviewer sees a PR-specific failure, I will address it.

AI assistance was used to prepare this CI comparison; I checked the linked GitHub Actions and Codecov check results.

